PR #19, opened 2020-10-27
implemented member info form

morganbelous:
### Summary

This pull request is the first step towards implementing the member info form that allows a user to change their information on DTI's website. Log in to IDOL, navigate to Users > Edit Profile to access the form.

- [X] implemented form
- [X] all required fields must not contain only whitespace in order to be valid
- [ ] make sure a user has correct permissions in order to change their first and last name  

### Test Plan  

Currently no unit tests 

### Notes 
 I created a dummy user to fill in the form information,  but the next step is to connect with the backend in order to retrieve actual user information from the database.

### Breaking Changes

Optional fields, such as double major and minor, are currently set as empty strings. In the database, these are stored as null , so we will need to decide how to handle this situation.


JBoss925 (approved):
LGTM! Also, feel free to ping me or other devs if you have a PR waiting for review!

PR #20, opened 2020-11-10
Connected member form to backend

morganbelous:
### Summary

This pull request is the next step towards implementing the member info form that allows a user to change their information on DTI's website. Log in to IDOL, navigate to Users > Edit Profile to access the form.

- [x] when a user updates their member information, the changes are reflected in firebase
- [x] make sure a user has correct permissions in order to change their first and last name 

### Notes
Still having issues with null/undefined. The Member type had fields like minor, double major, etc. as optional, but they are stored as null in the database. I changed the types of these fields in Member on the frontend to be string | null so that the updated user object conforms to the Member type, but now the Member type for backend and frontend don’t match. In the future, we need to figure out if we want to store null or empty strings in firebase.

### DIFF
--- a/idol/src/API/MembersAPI.ts
+++ b/idol/src/API/MembersAPI.ts
@@ -4,11 +4,22 @@ import { APIWrapper } from "./APIWrapper";
 import { Emitters } from "../EventEmitter/constant-emitters";
 
 export type Member = {
-  email: string,
-  first_name: string,
-  last_name: string,
-  role: string
-}
+  email: string;
+  first_name: string;
+  last_name: string;
+  role: string;
+  graduation: string;
+  major: string;
+  double_major: string | null;
+  minor: string | null;
+  website: string | null;
+  linkedin_link: string | null;
+  github_link: string | null;
+  hometown: string;
+  about: string;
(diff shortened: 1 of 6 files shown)


henry-li-06 (commented):
Good work with the member form, it looks nice! I think it would be good it we added some kind of indication that a user has successfully saved their profile information. At the moment, it feels very unclear whether or not saving was successful.

morganbelous:
@henry-li-06 Thanks! Right now, I have an alert for when the information is submitted successfully, did it not work for you? And do you have any ideas on how maybe I can make it better? Thanks!
screenshot

henry-li-06:
@morganbelous So it still wasn't working for me but I think the problem is just a miscommunication with some backend changes. It seems there's a check before the `alert()` for `val.status === 'Success'` but the `status` field for the response was changed to the HTTP status code in the newest changes for the backend (to implement response types for the API). I think if the check is `val.status === 200` then it should be fine.

morganbelous:
@henry-li-06 just fixed it, thanks!

henry-li-06 (approved):
Looks good!

PR #22, opened 2020-12-11
Profile photo uploader

morganbelous:
### Summary

This pull request is the first step towards implementing the profile photo uploader that allows a user to change their profile photo on DTI’s website. Log in to IDOL, navigate to Users > Edit Profile Photo to access the uploader.

- [x] User can upload a jpg or png from their local computer to IDOL
- [ ] Profile images are being fetched from and sent to the AWS bucket
- [ ] Fixed quality of image upload

### Notes

The design is definitely not perfect, so if anyone has any suggestions about how to make the profile uploader prettier or thinks that there’s a better spot to place it within IDOL, please let me know! Also having trouble figuring out how to not sacrifice image quality when a photo is uploaded and cropped, so any suggestions would be appreciated!


JBoss925 (approved):
Looks good! I like the way you laid out the FC a lot.

PR #23, opened 2020-12-17
React testing

morganbelous:
### Summary

 Some testing using react testing library for dev assignment 3. Honestly have no idea if these even make sense…I’ve never used react testing library or done test cases like this. Spent a lot of time trying to understand it out so just pushing what I came up with


henry-li-06 (approved):
LGTM

PR #28, opened 2021-02-16
[easy] Remove unused lazy components

SamChou19815:
### Summary 

These unused components cluttered the codebase. They are probably generated by some CLI, but they are really unnecessary to be here.

### Test Plan 

Things still type check, which shows that they are truly unused.


morganbelous (approved):
LGTM

PR #30, opened 2021-02-28
[RFC] Setup monorepo for idol

SamChou19815:
### Summary 

This PR takes what's currently on idol-backend master, copy it into here and setup yarn workspace for both frontend and backend project. I renamed the `idol` folder to `frontend` and the `idol-backend` to `backend` for consistency.

I switched to yarn for the support of workspaces. You can install yarn [here](https://classic.yarnpkg.com/en/docs/install#mac-stable).

Most of the setup should be unchanged. You can run the script in frontend/backend folder at the root by `yarn workspace [frontend or backend] [script name]`. To run all the scripts that have the same name in both folder together, you can do `yarn workspaces run [script name]`.

### Test Plan 

CI

### Notes

To keep this PR focused, I only do the bare-minimum of setup the workspace. In the future, we should setup a repo-wide eslint config and prettier config to make things more consistent.


henry-li-06 (approved):
LGTM

PR #33, opened 2021-03-02
Enforce prettier format for the entire repo

SamChou19815:
### Summary 

Enforce prettier format for the entire repo, and reformat the code so that it can pass the CI check.

### Test Plan 

It passes CI.

### Notes 

I also setup `husky` and `lint-staged`. You just need to run `yarn husky install`, and then it will auto format the code before commit


henry-li-06 (approved):
LGTM

PR #34, opened 2021-03-03
Change memberAPI to remove auth requirement for nova integration

SamChou19815:
### Summary 

In our meeting, we decide to make this API public to help nova integration. This API is readonly, and the info it exposed will be on the public team website anyways, so it won't introduce any security vulnerability.

### Test Plan 

Check you can visit the public API url in deploy preview. (sad the instance still hasn't secret set, so the endpoint fails)


henry-li-06 (approved):
LGTM, the private env for the backend has been added to netlify and public URL in deploy preview seems to work

PR #38, opened 2021-03-08
Update IDOL Data Types to Match Nova

riyajaggi:
### Summary 

This pull request updates the `member` and `role` data types for IDOL to match Nova's member jsons. The existing documents in the firebase are also updated to reflect the same changes.

- [x] change the case of member fields to camelCase
- [x] add missing fields (`roleDescription` and `netid`)
- [x] add missing roles (`'business'`)
- [x] update documents in firebase 

### Notes 

- Consider sharing .d.ts files with frontend for data type consistency

### Breaking Changes 

- Updated `Member` type: Check and update any frontend functionality for the same.

### DIFF
--- a/backend/data/run.ts
+++ b/backend/data/run.ts
@@ -43,7 +43,7 @@ db.collection('members')
         subteam: jsonData.subteam,
         otherSubteams: jsonData.otherSubteams,
         role: jsonData.roleId,
-        roleDescription: jsonData.roleDescription,  
+        roleDescription: jsonData.roleDescription
       };
 
       removeEmptyOrNull(data);
--- a/backend/data/run.ts
+++ b/backend/data/run.ts
@@ -27,20 +27,20 @@ db.collection('members')
       const email: string = jsonData.netid + emailDomain;
 
       const data = {
-        first_name: jsonData.firstName,
-        last_name: jsonData.lastName,
+        firstName: jsonData.firstName,
+        lastName: jsonData.lastName,
         email,
         role: jsonData.roleId,
         about: jsonData.about,
-        github_link: jsonData.github,
-        linkedin_link: jsonData.linkedin,
+        github: jsonData.github,
+        linkedin: jsonData.linkedin,
         major: jsonData.major,
         minor: jsonData.minor,
-        double_major: jsonData.doubleMajor,
+        doubleMajor: jsonData.doubleMajor,
         hometown: jsonData.hometown,
         graduation: jsonData.graduation,
         subteam: jsonData.subteam,
-        other_subteams: jsonData.otherSubteams,
+        otherSubteams: jsonData.otherSubteams,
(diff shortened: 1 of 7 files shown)


backend/src/DataTypes.d.ts
```diff
@@ -1,23 +1,41 @@
...
+export type Role =
+  | 'lead'
+  | 'admin'
+  | 'tpm'
+  | 'pm'
+  | 'developer'
+  | 'designer'
+  | 'business';
+
+export type roleDescription =
```
SamChou19815, line 12: Can we standardize to use `PascalCase` here?

backend/data/run.ts
```diff
@@ -27,21 +27,23 @@ db.collection('members')
...
+        about: jsonData.about,
         subteam: jsonData.subteam,
-        other_subteams: jsonData.otherSubteams,
-        website: jsonData.website
+        otherSubteams: jsonData.otherSubteams,
+        role: jsonData.roleId,
+        roleDescription: jsonData.roleDescription
       };
 
       removeEmptyOrNull(data);
```
SamChou19815, line 49: while you are here, I will point out that `removeEmptyOrNull` only removed undefined value, but not null values

SamChou19815 (changes requested):
Changes LGTM with some small nits, but I will put a hold here since this will break our frontend. Now that we are in a monorepo, we can fix the frontend together as well. @riyajaggi do you want to make related frontend changes (I think it's mostly change the type def and fix casing) or I can do it?

backend/src/DataTypes.d.ts
riyajaggi: done!

riyajaggi:
> Changes LGTM with some small nits, but I will put a hold here since this will break our frontend. Now that we are in a monorepo, we can fix the frontend together as well. @riyajaggi do you want to make related frontend changes (I think it's mostly change the type def and fix casing) or I can do it?

Thanks for reviewing this PR! Yes, I added the changes to the `MembersAPI.ts` for frontend. If there's anything I missed for frontend, feel free to add.

backend/data/run.ts
riyajaggi: Thank you for pointing this out! I believe this script was made by @henry-li-06 for initially adding/modifying jsons to the firebase. For this PR, I actually edited the script a little when I updated the documents in the firebase to take care of the undefined values and update fields. Should I push the updated script to git?

SamChou19815: Feel free to add it.

SamChou19815:
@riyajaggi you need to fix the additional type errors in the frontend

SamChou19815:
I fixed typing for now. But the frontend doesn't have support for role field yet. We should fix this in another PR.

SamChou19815 (approved):

PR #40, opened 2021-03-08
Derive NetID and role description from email and role

SamChou19815:
### Summary 

Unbreak saving user on the frontend by deriving netid and role description from existing fields. In the future, we probably shouldn't store them at all, since we can always easily derive them.

### Test Plan 

A screenshot showing that a setMember request sent by frontend. Here, the role description and netid is correct.
Screen Shot 2021-03-08 at 13 41 05

### DIFF
--- a/frontend/src/API/MembersAPI.ts
+++ b/frontend/src/API/MembersAPI.ts
@@ -9,6 +9,24 @@ type MemberResponseObj = {
   error?: string;
 };
 
+export type Role =
+  | 'lead'
+  | 'admin'
+  | 'tpm'
+  | 'pm'
+  | 'developer'
+  | 'designer'
+  | 'business';
+
+export type RoleDescription =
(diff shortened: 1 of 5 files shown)


henry-li-06 (commented):
What are your thoughts on having the backend derive the NetID and role instead of the frontend?

SamChou19815:
> What are your thoughts on having the backend derive the NetID and role instead of the frontend?

I think the ideal case is that these derivable fields don't exist in both frontend and backend at all, and only produce them in the script. However, it requires some type refactoring and data cleanup. This PR just serves as a temporary step to bring back data consistency.

Having backend derive the netid and role and send it to frontend doesn't quite help, since the frontend still need some code to derive them anyways.

henry-li-06 (approved):
Ah makes sense, LGTM

PR #43, opened 2021-03-15
only edit name if user has correct permissions

morganbelous:
### Summary

The user can now only edit their first and last name if they have the correct permissions ('admin' or 'lead').

- [x] fixed editing on user permissions
- [ ] fix error modal for when member info can't be changed

### Test Plan

admin_permissions
 Form for a user with 'admin' permissions.

developer_permissions
Form for a user with 'developer' permissions.

### Notes 

- This has a bit of duplicated code, so if anyone has any suggestions on how to reduce that I would appreciate it!
- I also noticed that the error modal isn't working when a user with the wrong permissions tries to edit their information, so that needs to be addressed

### DIFF
--- a/frontend/src/User/UserProfile/UserProfile.test.tsx
+++ b/frontend/src/User/UserProfile/UserProfile.test.tsx
@@ -6,8 +6,6 @@ import UserProfile from './UserProfile';
 describe('<UserProfile />', () => {
   test('labels', () => {
     render(<UserProfile />);
-    expect(screen.getByText('First name')).toBeInTheDocument();
-    expect(screen.getByText('Last name')).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


SamChou19815:
@morganbelous looking at the CI log looks like this change breaks a test...

SamChou19815:
@morganbelous You can use the trick [here](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) to avoid these duplicate code.

morganbelous:
> @morganbelous You can use the trick [here](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) to avoid these duplicate code.

thanks!!

SamChou19815 (approved):

PR #45, opened 2021-03-21
Setup common types between frontend, backend, and nova

SamChou19815:
### Summary 

Currently the idol type is still inconsistent with the nova type. However, since we are now in a monorepo, we can start to tweak the nova type to be closer to idol type.

### Test Plan 

This is a type only change, so check CI to ensure no errors.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -5,11 +5,11 @@
   "main": "index.js",
   "scripts": {
     "start": "./node_modules/.bin/netlify-lambda serve build",
-    "dev": "ts-node src/api.ts",
+    "dev": "ts-node --transpile-only src/api.ts",
     "tsc": "tsc",
     "build": "tsc"
   },
-  "license": "ISC",
+  "license": "AGPL-3.0",
   "dependencies": {
     "@types/body-parser": "^1.19.0",
     "@types/cors": "^2.8.7",
@@ -17,6 +17,7 @@
     "@types/express-session": "1.17.0",
     "@types/uuid": "^8.3.0",
     "body-parser": "^1.19.0",
+    "common-types": "1.0.0",
     "cors": "^2.8.5",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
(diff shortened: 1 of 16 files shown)


morganbelous:
On the frontend, the reason that the types for non-required fields like GitHub and website were string | null (instead of string?) was because Firebase doesn't store undefined values. When I tested this by removing my website link on my member form, the frontend said my information was updated but the field in Firebase didn't change. I guess a way to solve this would be to make sure that the backend converts undefined fields to null?

SamChou19815:
@morganbelous I changed the code to filter away the undefined values. I think it should work now.

morganbelous:
> @morganbelous I changed the code to filter away the undefined values. I think it should work now.

It seems like it's still not working? If I delete my website or GitHub link from the form and then save it I still get the confirmation that the info was updated, but the deletion isn't reflected in Firebase. It seems like if fields are empty they just aren't sent to the backend, but then the deletion isn't updated in Firebase.

riyajaggi:
> > @morganbelous I changed the code to filter away the undefined values. I think it should work now.
> 
> It seems like it's still not working? If I delete my website or GitHub link from the form and then save it I still get the confirmation that the info was updated, but the deletion isn't reflected in Firebase. It seems like if fields are empty they just aren't sent to the backend, but then the deletion isn't updated in Firebase.

Our backend endpoint is currently using the `update` firebase operation, which will only add/change those fields that are sent to the backend. So like you said, if the empty fields are not sent, they are not removed from the Firebase documents. 

We can discuss if you would prefer the backend to fix this issue instead. One way that I can think of  is to use the `set` operation, which would replace the entire document with the new body sent by the frontend (so including fields that were not changed).

SamChou19815:
I just updated the types. I changed the idol type to include null, which is actually more accuurate.

morganbelous:
> I just updated the types. I changed the idol type to include null, which is actually more accuurate.

LGTM!

morganbelous (commented):
LGTM

morganbelous (approved):
LGTM

PR #46, opened 2021-03-22
Add backend support for profile images

henry-li-06:
### Summary 

This PR adds backend support for uploading images. There are three additional ends for:
- Getting a signed URL for the current user's profile image
- Getting a signed URL for uploading a new profile image for the current user 
- Getting a list of signed URLs for all profile images 

### Test Plan 
If `images/<netid.jpg>` exists in firebase storage, then the current member's image should be displayed when the profile image upload page is loaded. 
Screen Shot 2021-03-21 at 2 51 02 AM

If the file does not exist, then a placeholder headshot should be displayed. 
image

The `/allMemberImages` endpoint should give a list of signed URLs in the form of:
```
[
   {
        fileName: <netid>.jpg,
        url: <signed URL>
    },
    ...
]
```


SamChou19815 (approved):
LGTM

PR #48, opened 2021-03-23
Cleanup nova project to prepare integration

SamChou19815:
### Summary 

Cleanup the nova workspace to prepare for using script to fetch from idol backend:

- Use the types from common-types
- Tweak json field order to be consistent

### Test Plan 

It's a type only change plus some json field order change. It should have no runtime effect. Check CI logs to ensure everything still passes.


henry-li-06 (approved):
LGTM

PR #49, opened 2021-03-23
[nova] Remove firstName and lastName from member json

SamChou19815:
### Summary 

They are never meaningfully used, since we always have a name field.

### Test Plan 

Things still type check.

### DIFF
--- a/dti-website/src/components/MemberProfile.tsx
+++ b/dti-website/src/components/MemberProfile.tsx
@@ -54,20 +54,14 @@ export default function MemberProfile({
           <img
             className="profile-image rounded-circle mx-auto d-block"
             src={`/static/members/${profile.info.netid}.jpg`}
-            alt={`${profile.info.firstName} ${profile.info.lastName}`}
+            alt={profile.info.name}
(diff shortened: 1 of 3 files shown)


dti-website/src/components/MemberProfile.tsx
```diff
@@ -54,20 +54,14 @@ export default function MemberProfile({
...
           <Col className="my-auto">
             <div className="profile-name-header">
-              {profile.info.name === undefined ? (
-                <div>{profile.info.name}</div>
-              ) : (
-                <div>
-                  {profile.info.firstName} {profile.info.lastName}
-                </div>
-              )}
+              <div>{profile.info.name}</div>
```
SamChou19815, line 64: lol this existing code is actually wrong when I migrated it (I flipped the condition). See the original code  https://github.com/cornell-dti/nova.cornelldti.org/blob/baa0f99a9b59d0c987ba39d83cfdedb89e88b74f/packages/website/src/components/MemberProfile.vue#L18-L21

henry-li-06 (approved):
LGTM

PR #50, opened 2021-03-23
Add script to generate members json from idol backend

SamChou19815:
### Summary 

This PR adds a script to generate the members json for nova site from the idol backend. This script now completely replaces the old script in `backend/scripts` so I removed them.

The script itself is now ready, but our backend is not. The script generates these changes that I didn't commit:

Screen Shot 2021-03-23 at 16 30 45

We need another admin collection instead of storing admin role in the member json, which doesn't work for the website.

I also added data for Ying since it looks like her data is missing by mistake.

### Test Plan 

Run `yarn workspace dti-website pull-from-idol` to play with the script by yourself.


henry-li-06 (approved):
LGTM, I'll update the backend for the issue with admins

PR #51, opened 2021-03-23
Update backend for admins collection 

henry-li-06:
### Summary 
This PR updates the backend for the admin collections to resolve the `role` field issue for admins that Sam mentioned in #50. The permissions manager now checks if the user is in the admins collection instead of checking the user's role in the members collection. 

### Test Plan 
Check that admins still have the correct permissions:
- Being able to edit users on `Edit Users` page
- Being able to change name in `Edit Profile` page
- Being able to edit teams 

### Notes 
The frontend needs to be updated to check if a user is an admin on the `Edit Profile` page in an alternative way since admins should not have `admin` in their `role` field anymore. 

All the admins should also have their role changed to their appropriate role in DTI to resolve the issue with the `role` field in the DTI website.


dti-website/server/server.js
```diff
@@ -0,0 +1,8 @@
+const express = require('express');
+const path = require('path');
+
+const app = express();
+
+app.use(express.static(path.join(__dirname, './../out')));
+
+app.listen(3000, () => console.log('DTI Website is listening on PORT 3000'));
```
SamChou19815, line 8: what's this?

SamChou19815 (approved):
LGTM. Question about a file

dti-website/server/server.js
henry-li-06: oh oops didn't mean to add that

PR #52, opened 2021-03-23
Setup login-checked endpoints

SamChou19815:
### Summary 

Setup the logged in check in one place, so we don't have to check at again at every handler. It can also help remove undefined returns in those endpoints, so we have less linter warnings.

### Test Plan 

### Notes

The PR will be very to review if you turn on hide whitespace change in github UI:

Screen Shot 2021-03-23 at 18 05 58


henry-li-06 (approved):
LGTM

PR #53, opened 2021-03-23
Remove all admin role from codebase

SamChou19815:
### Summary 

Now we have an admin collection, we no longer needs an admin role. This PR removes all of them.

### Test Plan 

👀

### DIFF
--- a/frontend/src/User/UserProfile/UserProfile.tsx
+++ b/frontend/src/User/UserProfile/UserProfile.tsx
@@ -112,7 +112,7 @@ const UserProfile: React.FC = () => {
   };
 
   let name;
-  if (role === 'admin' || role === 'lead') {
+  if (role === 'lead') {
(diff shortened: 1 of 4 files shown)


frontend/src/User/UserProfile/UserProfile.tsx
```diff
@@ -112,7 +112,7 @@ const UserProfile: React.FC = () => {
   };
 
   let name;
-  if (role === 'admin' || role === 'lead') {
+  if (role === 'lead') {
     name = (
       <Form.Group widths="equal">
         <Form.Input
```
SamChou19815, line 118: Technically this is a behavior change. But it doesn't really better, since if you really have admin permission, you can edit it in the user list any ways.

henry-li-06 (approved):
LGTM

PR #54, opened 2021-03-24
Fix linter warnings

SamChou19815:
### Summary 

Fix a lot of linter warnings. 99 warnings -> 12 warnings.

### Test Plan 

Type only changes, so it should have no runtime effect.


henry-li-06 (approved):
LGTM

PR #58, opened 2021-03-26
Handle all errors in handlers on top-level

SamChou19815:
### Summary 

This PR changes the backend implementation so that all errors are handled at the top-level in `src/api.ts`. I added error handlers to `loginCheckedHandler` so that it can also handle specific errors and general errors alike, and output correct status code.

As a result, we no longer need to pass along the error all the way from database DAO to xxxAPI to api.ts. Instead, we just need to let the error throw or throw specific errors (e.g. `PermissionError`, `NotFoundError`) for the top-level handler to handle.

### Test Plan 


henry-li-06 (approved):
LGTM

PR #59, opened 2021-03-26
Use firestore data converter for statically typed DB operations

SamChou19815:
### Summary 

Using typescript with [`FirestoreDataConverter`](https://firebase.google.com/docs/reference/js/firebase.firestore.FirestoreDataConverter) gives all the database operations a more specific type instead of any type. This means that we won't accidentally write data that doesn't belong to a collection.

Here is a little demo showing how it can help to type check database operations:

Screen Shot 2021-03-26 at 11 09 50

I replaced all the existing untyped calls to the database collection methods with the typed ones. This change reveals the hidden bug where we don't check member is `null` in team API and using `first_name` instead of `firstName` in member API.

### Test Plan 

### DIFF
--- a/backend/src/dao/MembersDao.ts
+++ b/backend/src/dao/MembersDao.ts
@@ -1,28 +1,25 @@
-import { db } from '../firebase';
+import { memberCollection } from '../firebase';
 
 export default class MembersDao {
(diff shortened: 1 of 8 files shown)


JBoss925 (approved):
this is sick, I did not know this existed.

PR #61, opened 2021-03-27
Added netlify build cron job and dti site netlify functions server

JBoss925:
### Summary

NOTE: Production DTI site is unaffected by these changes for now, they are deploying under "new" subdomain. Moving the site over will be very easy, as we just change the DNS cname.

This PR does three things. 

- Added a github workflow cron job to trigger the DTI netlify site to rebuild every 2 hours (and update its data in doing so, allowing the website to update if nobody merges the PR to update the members json)
- Added a netlify functions server to the dti site so that it can potentially become interactive and/or we can collect analytics. This can be easily removed, but I want to play around with it for the time being.
- Added build steps for the DTI website and website server to allow live redeployment.

Currently this is all running on https://new.cornelldti.org/ to test, but we can move it over to prod if it proves to work well on the test site. Please ignore this PR otherwise, as I just need to get the workflow running.

### Test Plan

I already tested it by building on netlify!

### Breaking Changes

Nah.


dti-website-server/src/server.ts
```diff
@@ -0,0 +1,15 @@
...
+import serverless from 'serverless-http';
+
+// Constants and configurations
+const app = express();
+const router = express.Router();
+
+router.get('/', (req: Request, res: Response) => {
+  console.log('Ping!');
+  res.send('Pong!');
+});
```
SamChou19815, line 11: I don't understand why we need this server file to collect analytics. If we need analytics, we can always do it on client side using google analytics.

.github/workflows/rebuild-dti-site.yml
```diff
@@ -0,0 +1,12 @@
+name: Rebuild DTI from Master Every 2 Hours
+on:
+  schedule:
+  - cron: '0 */2 * * *'
```
SamChou19815, line 4: This is potentially very wasteful, since netlify build minutes are quite expensive (compare to free github action minutes for open source)... The website is not going to change a lot, and I think one day interval (like the current idol backend syncing bot) is probably good enough.

dti-website/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "start": "next",
-    "build": "next build && next export",
+    "build": "yarn run pull-from-idol && next build && next export",
```
SamChou19815, line 6: Not sure whether we want this kind of automatic deployment without approval process. Imaging someone added some bad self description or we messed up the json, then it will appear on the website live without any human intervention.

netlify.toml
```diff
@@ -1,4 +0,0 @@
-[build]
-  publish = "frontend/build/"
-  functions = "backend/build"
-  command = "npm run build"
```
SamChou19815, line 4: Is there a reason why it's removed?

.github/workflows/rebuild-dti-site.yml
```diff
@@ -0,0 +1,12 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Trigger the build webhook on Netlify
+      run: curl -s -X POST "https://api.netlify.com/build_hooks/${TOKEN}"
+      env:
+        TOKEN: ${{ secrets.NETLIFY_CRON_BUILD_HOOK }}
```
SamChou19815, line 12: Doing this trigger on every merge to master is probably much cheaper...

dti-website-server/src/server.ts
JBoss925: This is just a placeholder for something I wanted to try out, so it's not really for analytics, but I've just realized I can manually add the workflow to main, which is why I opened the PR in the first place. but I see the PR review is how you're reviewing the member JSON changes, so I'm going to switch up the workflow to only be on merges to master. Then I'll add a button in IDOL that let's admins redeploy the website and pull the data from IDOL at will, which would allow us to update it immediately if we so choose. I'm closing this PR for now, but I'll make those changes now that I understand the structure a bit better.

netlify.toml
JBoss925: this is removed because the dti site needed a different publish directory and it was overriding the one I put in the netlify website lol

PR #62, opened 2021-03-27
Ensure type safety and basic member authorization in loginCheckedHandler

SamChou19815:
### Summary 

Right now, every xxxApiHandler checks whether the member actually exists in database, which is a check that can be easily forgotten by programmers. In addition, every handler parsed information from the request object on its own, which can be error prone.

This PR moves all the logic to `loginCheckedHandler`, which checks the existence of user in db and returns a user object to the handler. Each handler is also provided information relevant to them in typed form, so that all the request object parsing happens in one place: `api.ts`

### Test Plan 

### DIFF
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -22,6 +22,7 @@ import {
 } from './site-integration';
 import { allRoles, PermissionsManager } from './permissions';
 import { HandlerError } from './errors';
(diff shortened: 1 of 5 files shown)


JBoss925 (approved):
LGTM. We do end up requesting the member data in like 99% of the endpoints so I like this pattern.

PR #63, opened 2021-03-27
Replace untyped materialize with manual typed construction for team object

SamChou19815:
### Summary 

The `materialize` function is untyped and doesn't have a chance to be typed because it's very dynamic. However, we don't really need its superpower since our data format is very simple. The only place that uses it is in the team object construction.

I replaced `materialize` with manual construction, which removes 6 additional linter warnings.

### Test Plan 

You can still load and update teams, so the changes don't break anything.


henry-li-06 (approved):
LGTM

PR #64, opened 2021-03-27
[easy] Fix simple accessibility warning on website button

SamChou19815:
### Summary 

According to accessibility standards, `a` tag must have `href`. In the existing code, we are mostly using `a` tag to achieve some styling effect, which is bad. This PR replaces it with `div` and everything still looks the same.

### Test Plan 

No change in appearance

Screen Shot 2021-03-26 at 23 42 20


henry-li-06 (approved):
LGTM

PR #78, opened 2021-03-29
Refactor frontend API calls to reduce boilerplate

### LINKED ISSUE #27: [code-quality] A lot of frontend code uses any
A search shows there are a lot of `any` types:

https://github.com/cornell-dti/idol/search?q=any

Use of `any` type basically turns off the ts compiler and can hide type errors. They should be replaced with accurate types.

SamChou19815:
### Summary 

- All the APIWrapper's config parameters are all about `withCredentials: true`, which is already set in APIWrapper. Therefore, there is no need to pass it again, so I removed the parameters.
- Strictified the types of APIWrapper return a little bit

As a result of this refactoring all the linter warnings are gone!

Close #27.

### Test Plan 

Go into the app to ensure that things can load, which shows that the credentials are still being correctly passed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,13 +8,11 @@
   },
   "rules": {
     "camelcase": "off",
+    "no-alert": "off",
     "no-console": "warn",
     "no-shadow": "off",
-    "no-restricted-syntax": "warn",
     "no-use-before-define": "off",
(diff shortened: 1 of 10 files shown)


henry-li-06 (approved):
LGTM

PR #79, opened 2021-03-29
Script to download all member images from Firebase Storage

riyajaggi:
### Summary 

This pull request implements a script `download_images.ts` to download all member images from the `images` bucket in IDOL's Firebase Storage and saves them with the same name (i.e. _netid.jpg_). It does not affect any existing functionality.

### Test Plan 

- [x] Run the script to ensure all images are downloaded.
image

### Breaking Changes 

None

### DIFF
--- a/backend/data/download-images.ts
+++ b/backend/data/download-images.ts
@@ -1,25 +1,31 @@
-import { bucket } from './firebase';
+#!/usr/bin/env node
+
+import fetch from 'node-fetch';
+import { ProfileImage } from '../src/DataTypes';
 
 const fs = require('fs-extra');
 
-const dirPath = 'backend/data/members/images/';
-if (!fs.existsSync(dirPath)) {
-  fs.mkdirSync(dirPath);
-} else {
-  fs.emptyDirSync(dirPath);
+const dirPath = './data/members/images/'; 
+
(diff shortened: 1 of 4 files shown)


henry-li-06 (commented):
So our backend already has an endpoint `/allMemberImages` that does not require authentication and gives a list of download URLs for each image in the Firebase storage. I think this script can just fetch that endpoint and use the download URLs to download the images, somewhat similar to what was done for the member JSONs in #50. That way this script doesn't need to connect to our firebase instance.

backend/data/download-images.ts
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env node
+
+import fetch from 'node-fetch';
+import { ProfileImage } from '../src/DataTypes';
+
+const fs = require('fs-extra');
```
SamChou19815, line 6: Change this to use import and `yarn workspace backend add fs-extra`

SamChou19815 (changes requested):
Almost good, just one change.

backend/data/download-images.ts
riyajaggi: done!

henry-li-06: image
There's still another linting error that needs to be fixed

SamChou19815 (approved):

PR #88, opened 2021-03-31
Setup dev Firebase instance 

henry-li-06:
### Summary 

I just set up another Firebase instance for development so that testing doesn't affect the data in the production Firebase instance. This way changes in our data due to testing will not affect the production data and potentially cause the DTI website to unnecessarily rebuild. 

### Test Plan 

N/A

### Notes 
- All of you should've received an invitation to the development Firebase instance
- The data in the development Firebase instance isn't complete

### DIFF
--- a/backend/resources/cornelldti-idol-firebase-adminsdk-ifi28-9aaca97159.json
+++ b/backend/resources/cornelldti-idol-firebase-adminsdk-ifi28-9aaca97159.json
@@ -1,0 +1,12 @@
+{
+  "type": "service_account",
+  "project_id": "cornelldti-idol",
+  "private_key_id": "",
+  "private_key": "",
+  "client_email": "firebase-adminsdk-ifi28@cornelldti-idol.iam.gserviceaccount.com",
+  "client_id": "105964523598957732075",
+  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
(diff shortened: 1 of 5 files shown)


SamChou19815:
@henry-li-06 can you document how you can run the server using prod vs dev dababase in some readme file? Also how does it affect deployed server in preview and prod?

henry-li-06:
> @henry-li-06 can you document how you can run the server using prod vs dev dababase in some readme file? Also how does it affect deployed server in preview and prod?

I don't think it should have any effect on production. It should just change the Firebase instance that the backend/frontend uses by default when testing things locally. That way our production data doesn't get messed with and cause unnecessary redeploys.

henry-li-06:
Seems to actually cause an issue w/ the preview server.... I'll take a look at that

SamChou19815 (approved):
Looks like the preview is now correctly using the dev db. Not sure what's happening in production, but if the production uses prod db then everything looks good to me.

henry-li-06:
I'm still having an issue with logging into the IDOL preview so I'm gonna try to figure out what's going on there first.

henry-li-06:
@SamChou19815 I think the issue w/ logging in should be fixed now. The preview should now be using the production database. Could you take another look just to make sure everything looks ok?

SamChou19815 (approved):
yes it seems to be working now

PR #94, opened 2021-04-08
Add script to pull and compress headshots from idol-backend

SamChou19815:
### Summary 

Mostly based on @riyajaggi's script in backend, but connected with image-cli compression script and added cleanup step. This script will now pull images from backend, compress it, and put them into the right place of the website directory.

### Test Plan 

After running `yarn workspace dti-website pull-compress-headshots` we get:
Screen Shot 2021-04-08 at 13 44 36

### Notes 

Unfortunately, this is not production ready beyond the script. The dti-website expects headshots of size 500x667, while the idol site only crops the image to be a square...


henry-li-06:
How much of an issue is cropping the images to be square? It seems that Emily's current headshot, for example, is 500x500 and looks fine on the website.
image
image

henry-li-06:
The image-cli also seems to resizing the images to have a height of 500 instead of 667, also not sure how much of an issue this is.

SamChou19815:
@henry-li-06 looks like it should be good then. Although I would still prefer to not connect this to the main pull from idol script yet, before we load in all the high-resolution old headshots into the file storage.

henry-li-06:
> @henry-li-06 looks like it should be good then. Although I would still prefer to not connect this to the main pull from idol script yet, before we load in all the high-resolution old headshots into the file storage.

You mentioned that the high-resolution images should be stored in some AWS storage bucket right? If that's the case I'll try to get those and store in our Firebase storage.

henry-li-06 (approved):

SamChou19815:
> > @henry-li-06 looks like it should be good then. Although I would still prefer to not connect this to the main pull from idol script yet, before we load in all the high-resolution old headshots into the file storage.
> 
> You mentioned that the high-resolution images should be stored in some AWS storage bucket right? If that's the case I'll try to get those and store in our Firebase storage.

@henry-li-06 yes that's right

PR #99, opened 2021-04-16
Allow approving a subset of member information change

SamChou19815:
### Summary 

The current flow of member information change approval has a big issue: if one member's information change has something we don't want, we cannot approve others' information without leaving the bad one out.

This PR aims to fix the issue by building a per-person information change approval flow. I build an IDOL page that loads in diffs **on each member**, and allows the admin to selectively approve a subset of them. This allows the good ones to go through, without making the bad change taking hostage of the entire flow.

This PR only implements the review part. It does not alter the rest of the system yet. It currently only stores the approved members' information into a separate collection. I exposed an endpoint that fetches from that collection instead of the old collection. In the next PR, we can make the `pull-from-idol` script pull from that collection instead.

#### Future Work

- Integrate the new endpoint into `pull-from-idol` script.
- Implement similar feature for profile pictures
- Have some mechanism to notify the admin that there is something to review (could be a slackbot).

### Test Plan 

Example:
Screen Shot 2021-04-15 at 22 01 40

### Notes 

@meganyin13 what you want has now been built.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -24,6 +24,7 @@
     "body-parser": "^1.19.0",
     "common-types": "1.0.0",
     "cors": "^2.8.5",
+    "diff": "^5.0.0",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
     "express-session": "^1.17.1",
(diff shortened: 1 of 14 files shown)


JBoss925 (approved):
LGTM!

PR #100, opened 2021-04-17
Idol backend/update subteam

gracshan:
### Summary 

This pull request updates the subteam field for members when a team is edited on IDOL. It also allows for creating new teams and deleting existing teams.

### Test Plan 
Initially, subteam is idol: 
Screen Shot 2021-04-16 at 10 57 17 PM
Adding new team:
Screen Shot 2021-04-16 at 10 54 52 PM
Updates in Firestore:
Screen Shot 2021-04-16 at 10 56 28 PM
Deleting the team replaces subteam field with empty string:
Screen Shot 2021-04-17 at 9 50 27 AM


backend/src/dao/TeamsDao.ts
```diff
@@ -1,4 +1,4 @@
-import { v4 as uuidv4 } from 'uuid';
+import { EmitFlags } from 'typescript';
```
SamChou19815, line 1: Remove this

backend/src/teamAPI.ts
```diff
@@ -1,11 +1,47 @@
...
+    deletedMembers = oldTeamMembers.filter(
+      (member) => !newTeamMembers.includes(member)
+    );
+  } else {
+    newMembers = [...team.leaders, ...team.members];
+  }
+
+  newMembers.forEach((member) => {
+    const updatedMember = { ...member };
+    updatedMember.subteam = team.name;
```
SamChou19815, line 34: You can do `const updatedMember = { ...member, subteam: team.name };`

dti-website/src/data/all-members.json
```diff
@@ -6,7 +6,7 @@
     "major": "CS",
     "hometown": "Manhattan, KS",
     "about": "Hi! I'm a CS and Math double major in the college of A&S. My main interests are in web dev, data science, and AI/ML. I love working on creative products that positively impact people's lives. In my free time I like playing league of legends, watching anime, and hanging out with friends.",
-    "subteam": "reviews",
+    "subteam": "",
```
SamChou19815, line 9: Revert changes in this file unless the data is good.

backend/src/teamAPI.ts
```diff
@@ -1,11 +1,47 @@
...
+    MembersDao.setMember(updatedMember.email, updatedMember);
+  });
```
SamChou19815, line 36: It should be done using promise, like: 

```typescript
await Promise.all(newMembers.map(async member => {
  const updatedMember = { ...member, subteam: team.name };
  await MembersDao.setMember(updatedMember.email, updatedMember);
});
```

SamChou19815 (changes requested):

dti-website/src/data/all-members.json
gracshan: Can I ask how I can check whether the data is good? I didn't make these changes, so I'm not quite sure where to look.

SamChou19815: @gracshan The current data about subteam should be mostly right, so diffs that change subteam field to empty string is definitely wrong. We cannot merge this PR because once we merge it, the team website with the wrong data will be auto-deployed.

gracshan: @SamChou19815 I see. I'll revert it based on what I see displayed in the changes here. Thank you for letting me know!

backend/src/dao/TeamsDao.ts
```diff
@@ -27,9 +26,29 @@ export default class TeamsDao {
...
+          ref.get().then((doc) => doc.data() as IdolMember)
+        )
+      )
+    };
+  }
+
   static async setTeam(team: Team): Promise<Team> {
     const teamRef: DBTeam = {
-      uuid: team.uuid ? team.uuid : uuidv4(),
+      uuid: team.uuid,
```
SamChou19815, line 51: why this change?

henry-li-06: I think originally in the `updateTeamMembers` function, it checked if `team.uuid` was empty and generated a new one if it was. That way when `setTeam` was called, `team.uuid` would always be set already. I think that caused some linter error though since doing so updated the parameter to the function so the function now makes a copy of the team. So this line should actually probably. be changed back.

SamChou19815 (approved):

PR #101, opened 2021-04-19
Basic script for populating db

riyajaggi:
### Summary 

This pull request is the first step towards populating the teams db in idol. This makes use of existing member documents and adds them to the team collections based on their role and subteam.

### Test Plan 

Ensure all team endpoints are working correctly after populating the db

### Notes 

This still needs some work.

### Breaking Changes 

- no change in schema
- need to ensure all teams are updated correctly

### DIFF
--- a/backend/data/firebase.ts
+++ b/backend/data/firebase.ts
@@ -1,0 +1,28 @@
+require('dotenv').config({path:'/backend/data/.env'})
+
+import admin = require('firebase-admin');
+
+const serviceAccount = require('backend/resources/cornelldti-idol-firebase-adminsdk-ifi28-9aaca97159.json');
+console.log("here");
+const configureAccount = (sa) => {
+  const configAcc = sa;
+  let parsedPK;
+  try {
+    console.log(process.env.FIREBASE_DEV_PRIVATE_KEY);
+    parsedPK = JSON.parse(process.env.FIREBASE_DEV_PRIVATE_KEY as string);
+    console.log("here");
+  } catch (err) {
+    console.log(err);
+    parsedPK = process.env.FIREBASE_DEV_PRIVATE_KEY as string;
+  }
+  configAcc.private_key = parsedPK;
+  configAcc.private_key_id = process.env.FIREBASE_DEV_PRIVATE_KEY_ID;
+  return configAcc;
+};
+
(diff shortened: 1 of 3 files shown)


backend/data/firebase.ts
```diff
@@ -0,0 +1,28 @@
+require('dotenv').config({path:'/backend/data/.env'})
+
+import admin = require('firebase-admin');
+
+const serviceAccount = require('backend/resources/cornelldti-idol-firebase-adminsdk-ifi28-9aaca97159.json');
```
SamChou19815, line 5: could you fix these linter errors and warnings first

backend/data/populate-teams.ts
```diff
@@ -0,0 +1,35 @@
...
+      } else {
+        teamData = {
+            uuid: teamData.uuid,
+            name: teamData.name, 
+            leaders: teamData.leaders,
+            members: teamData.members.add(/members/+email)
+        };
+      }
+    //   db.doc(`teams/${teamData.uuid}`).update(teamData);
+    console.log(teamData);
```
SamChou19815, line 34: this function doesn't seem to do anything at the moment, due to this db code being commented out

SamChou19815 (changes requested):

henry-li-06:
I'm not sure if we really need to keep the script to populate `teams` collection. We theoretically won't ever need to mass populate the collection again since teams can be added/deleted/edited from the _Edit Teams_ page. Thoughts @SamChou19815?

SamChou19815:
@henry-li-06 good point. Also why are we keeping the teams collection if the team information can already be inferred from the subteam field?

henry-li-06:
@SamChou19815 I guess we don't necessarily need the teams collection since the entirety of the collection can be derived from the members collection and the relationship between members and teams adds some complexity to our data. But at the same time I think operations on teams would be less computationally wasteful if we leave the collection (ex. don't need to go through all members to delete a team).

SamChou19815:
@henry-li-06 Given that we are only deal with data set with size around 100, I think that kind of performance benefits are negligible, but the cost of syncing the data in two collections are real. That being said, I'm ok with #103 merged if breaks are fixed, since we need to fix the subteam issue first.

PR #103, opened 2021-04-21
Change subteams to an array and former members

henry-li-06:
### Summary 

This PR updates the `subteam` field for the `Member` type to `subteams` which is now an array since a member can be part of multiple subteams and also renames the `otherSubteams` field to `formerSubteams`. A `formerMembers` field has also been added to the `Team` type and adds a `Former Members` section on the _Edit Teams_ page where the former members of a team can be edited. 

#### DTI Website 
No change in functionality for the DTI website, just run locally to make sure that subteams and former subteams are properly displayed. 

#### IDOL Website 
Check that operations on teams work as expected (creating, deleting, editing teams). Also check that adding/removing a member or former member correctly updates the member's own data. 
**(Test the IDOL website locally using dev database since the prod db does not have the updated fields, also the preview most likely will not work)**

### Breaking Changes 
**This PR requires database schema changes so DO NOT merge before updating the prod db**


common-types/index.d.ts
```diff
@@ -27,8 +27,8 @@ interface IdolMember {
   readonly github?: string | null;
   readonly hometown: string;
   readonly about: string;
-  readonly subteam: string;
-  readonly otherSubteams?: readonly string[] | null;
+  readonly subteams: string[];
+  readonly formerSubteams?: readonly string[] | null;
```
SamChou19815, line 31: Make all the string array type `readonly` in this file for consistency?

SamChou19815 (commented):
Changes mostly look good. Just need to fix type errors.

henry-li-06:
@SamChou19815 I think the type errors should be all good now, mostly just had to update `all-members.json` with the new fields.

dti-website/src/pages/team.tsx
```diff
@@ -24,13 +24,13 @@ export default function TeamPage(): JSX.Element {
...
   ): readonly { info: NovaMember; id: string }[] =>
-    members
+    (members as NovaMember[])
       .filter(
         (member) =>
           ((typeof member.roleId === 'string' &&
             member.roleId.endsWith(role)) ||
             role === '') &&
-          // @ts-expect-error: missing
+          // // @ts-expect-error: missing
```
SamChou19815, line 33: solved?

dti-website/src/pages/team.tsx
```diff
@@ -24,13 +24,13 @@ export default function TeamPage(): JSX.Element {
     role = '',
     isLead = false
   ): readonly { info: NovaMember; id: string }[] =>
-    members
+    (members as NovaMember[])
```
SamChou19815, line 27: why this cast?

dti-website/src/components/DtiProject.tsx
```diff
@@ -20,11 +20,11 @@ export default function DtiProject({
...
   if (project == null) throw new Error();
-  const pastMembers = allMembers.filter((member) =>
-    (member.otherSubteams || []).includes(teamId)
+  const pastMembers = (allMembers as NovaMember[]).filter((member) =>
+    ((member.formerSubteams as string[]) || []).includes(teamId)
   );
-  const currentMembers = allMembers.filter(
-    (member) => member.subteam === teamId
+  const currentMembers = (allMembers as NovaMember[]).filter((member) =>
+    member.subteams ? (member.subteams as string[]).includes(teamId) : false
```
SamChou19815, line 27: why do we need cast here

dti-website/src/pages/team.tsx
henry-li-06: i think this and the other casts are unnecessary actually

henry-li-06: oh just kidding, without these casts i was getting some type errors but not sure if there is a better way to resolve that

dti-website/src/pages/team.tsx
henry-li-06: I was getting `Unused '@ts-expect-error' directive.`. Should this just be removed?

dti-website/src/pages/team.tsx
SamChou19815: I think it's caused by the `hello DTI` in the member. Deleting that and things should work without casts.

dti-website/src/components/DtiProject.tsx
```diff
@@ -21,10 +21,10 @@ export default function DtiProject({
   const project = allProjects.find((it) => it.teamId === teamId);
   if (project == null) throw new Error();
   const pastMembers = allMembers.filter((member) =>
-    (member.otherSubteams || []).includes(teamId)
+    ((member.formerSubteams as string[]) || []).includes(teamId)
   );
-  const currentMembers = allMembers.filter(
-    (member) => member.subteam === teamId
+  const currentMembers = allMembers.filter((member) =>
+    member.subteams ? (member.subteams as string[]).includes(teamId) : false
```
SamChou19815, line 27: probably don't need cast here

SamChou19815 (approved):
Final nit. Otherwise looks good.

dti-website/src/components/DtiProject.tsx
henry-li-06: Getting `Argument of type 'string' is not assignable to parameter of type 'never'` without the cast

SamChou19815 (approved):

PR #105, opened 2021-04-30
shoutouts backend

morganbelous:
### Summary

This pull request is the next step towards implementing the shoutouts page on IDOL. The user can now send shoutouts to other DTI members and view their given and received shoutouts from Firebase.

- [x] implemented backend routes
- [ ] implemented option to give an anonymous shoutout
- [ ] implemented option to give multiple people a shoutout at once

### Test Plan

- Ensured that nothing breaks when users have either or both lists of shoutouts empty
- Ensure that users can’t post shoutouts for themselves

### Notes

- Right now users can view their shoutouts right as they are posted, not sure if we would want to give the leads the ability to look over each shoutout before confirming that they are ok?

### DIFF
--- a/backend/src/dao/ShoutoutsDao.ts
+++ b/backend/src/dao/ShoutoutsDao.ts
@@ -2,7 +2,10 @@ import { memberCollection, shoutoutCollection } from '../firebase';
 import { Shoutout, DBShoutout } from '../DataTypes';
 
 export default class ShoutoutsDao {
-  static async getShoutouts(email: string, type: string): Promise<Shoutout[]> {
+  static async getShoutouts(
+    email: string,
+    type: 'given' | 'received'
+  ): Promise<Shoutout[]> {
     const givenOrReceived = type === 'given' ? 'giver' : 'receiver';
     const shoutoutRefs = await shoutoutCollection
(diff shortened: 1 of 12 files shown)


backend/src/firebase.ts
```diff
@@ -1,5 +1,5 @@
 import admin from 'firebase-admin';
-import { DBTeam } from './DataTypes';
+import { DBShoutout, DBTeam, Shoutout } from './DataTypes';
```
SamChou19815, line 2: Fix the linter warning on this line

backend/src/dao/ShoutoutsDao.ts
```diff
@@ -0,0 +1,33 @@
+import { memberCollection, shoutoutCollection } from '../firebase';
+import { Shoutout, DBShoutout } from '../DataTypes';
+
+export default class ShoutoutsDao {
+  static async getShoutouts(email: string, type: string): Promise<Shoutout[]> {
```
SamChou19815, line 5: Use literal type for `type`: `type: 'given' | 'received'` Same with the frontend one.

SamChou19815 (approved):
LGTM

PR #106, opened 2021-05-01
Allow both accept and revert in member reviews

SamChou19815:
### Summary 

This PR adds support for reverting a change in member information. This is helpful in several scanarios:

- Some member just puts some bad stuff that can never be accepted, so it's better to reject it.
- We put in some changes to test the system. It's handy to build some UI to revert some of the testing changes.

### Test Plan 

Screen Shot 2021-05-01 at 19 25 53


henry-li-06 (approved):
LGTM

PR #117, opened 2021-05-08
Try fixing race condition in session

SamChou19815:
### Summary 

According to the [docs](https://www.npmjs.com/package/express-session#resave), setting them to `true` can potentially cause race conditions. In our shoutouts page, we make 3 concurrent requests to the backend. Due to much higher latency from client to prod server than local client to server, it becomes increasing likely that the server is handling concurrent requests, which can mess up.

This PR sets it them false to see whether the issue can be fixed.

### Test Plan 

Test in staging


SamChou19815:
Still doesn't fix the problem

PR #118, opened 2021-05-08
Sign in link

JBoss925:
### Summary 
Added a sign in link functionality

This replaces bit.ly, and allows us to automatically track sign in forms.

We can provide a deeplink like:
`https://idol.cornelldti.org/forms/signin/<id>`
to automatically sign the user in when they log in with google (or automatically if already logged in).

It also rearranges the menu system quite substantially. But it definitely makes more sense to me, and it allows us to demo the forms section and let people play with it at all hands.

### Test Plan
Tested locally on dev db instance, and it works! It's really just basic document CRUD stuff lol.

- Added a new Firebase collection called sign-in-forms to hold, you guessed it, sign-in forms


frontend/src/environment.ts
```diff
@@ -4,7 +4,7 @@ export const isProduction = process.env.NODE_ENV === 'production';
 const useProdBackendForDev = false;
 
 /** Switch to false to use development Firebase instance. Change back to true before committing. */
-export const useProdDb = true;
+export const useProdDb = false;
```
henry-li-06, line 7: this should be changed back to `true`

JBoss925: good catch lol

SamChou19815 (changes requested):
Put a break on this. Let's do the routes refactoring first before introducing the signin links features.

SamChou19815 (approved):

PR #122, opened 2021-05-08
[easy] Fix linter warnings and strictify config

SamChou19815:
### Summary 

Make `unused-var` rule error level so that it will always be respected.

### Test Plan 

👀


henry-li-06 (approved):
LGTM

PR #124, opened 2021-05-08
Fix session bug

henry-li-06:
### Summary 

This should fix (at least mostly) the bug in prod that's causing sessions to randomly time out. Essentially, there were some requests in the shoutouts page that were happening concurrently that I guess were causing the backend to bug. I chained together these requests so that they were no longer happening at the same time. 
### Test Plan 

On the preview, you should not be randomly logged out as you click around, particularly on the `Shoutouts` and `Edit Profile` pages.

### DIFF
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -246,7 +246,7 @@ app.use('/.netlify/functions/api', router);
 if (!isProd) {
   app.listen(PORT, () => {
     // eslint-disable-next-line no-console
-    console.log(`IDOL backend listening on port ${PORT}`);
+    console.log(`IDOL backend listening on port: ${PORT}`);
   });
 }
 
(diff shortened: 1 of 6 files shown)


SamChou19815 (approved):
Accepting this with some reservation. I think this is only a temporary solution, since the backend shouldn't really depend on nice client side behavior. In principle, the backend should be able to handle arbitrarily many concurrent requests.

I think a better fix in the future is to get rid of the session and use the `verifyIDToken` from firebase in every request. According to this [issue](https://github.com/firebase/firebase-admin-java/issues/430), the function should be very efficient since it does the caching on its own.

henry-li-06:
> Accepting this with some reservation. I think this is only a temporary solution, since the backend shouldn't really depend on nice client side behavior. In principle, the backend should be able to handle arbitrarily many concurrent requests.
> 
> I think a better fix in the future is to get rid of the session and use the `verifyIDToken` from firebase in every request. According to this [issue](https://github.com/firebase/firebase-admin-java/issues/430), the function should be very efficient since it does the caching on its own.

Sounds good

PR #128, opened 2021-05-10
Change prettier line width to 100

SamChou19815:
### Summary 

Finally no open PRs now. Let's change the line width to 100 so that we can more things in a line. Reviewers can take a look at the diff so see the code with fewer wraps.

### Test Plan 

CI


morganbelous (approved):
LGTM

PR #131, opened 2021-05-12
Anonymous Shoutouts

morganbelous:
### Summary

This pull request is the next step towards implementing shoutouts on IDOL. The user can now give anonymous shoutouts.
- [x] implemented option to give an anonymous shoutout

Before:
shoutouts_before

After:
shoutouts_after

anon_form

### Test Plan

Checked that giving and displaying shoutouts still works as intended.


backend/src/dao/ShoutoutsDao.ts
```diff
@@ -28,11 +28,12 @@ export default class ShoutoutsDao {
...
     return Promise.all(
       shoutoutRefs.docs.map(async (shoutoutRef) => {
-        const { giver, receiver, message } = shoutoutRef.data();
+        const { giver, receiver, message, isAnon } = shoutoutRef.data();
         return {
           giver: (await giver.get().then((doc) => doc.data())) as IdolMember,
           receiver: (await receiver.get().then((doc) => doc.data())) as IdolMember,
-          message
+          message,
+          isAnon
```
SamChou19815, line 36: When it's anonymous, you should strip away the giver from the returned result from the backend, otherwise people can still see by checking the chrome dev tools networks tab. To support that, you can do a union type on shoutouts:

```ts
type Shoutout = {
  receiver: IdolMember;
  message: string;
  isAnon: true;
} | {
  giver: IdolMember;
  receiver: IdolMember;
  message: string;
  isAnon: false;
};;
```

SamChou19815 (changes requested):

SamChou19815 (approved):

PR #138, opened 2021-08-30
[dti48] Allow user to choose a different semester of data

SamChou19815:
### Summary 

Fetch all semesters' data and let user choose. Step 2 of making DTI 48 playable for old members. Next step would be allow old members to log in.

### Test Plan 

Screen Shot 2021-08-30 at 15 34 02

### DIFF
--- a/frontend/src/Games/DTI48/DTI48.tsx
+++ b/frontend/src/Games/DTI48/DTI48.tsx
@@ -1,22 +1,38 @@
 import { useState, useEffect } from 'react';
 import { Loader } from 'semantic-ui-react';
 import { MembersAPI } from '../../API/MembersAPI';
-import DTI48MainGame from './DTI48MainGame';
+import DTI48MainGame, { MembersFromAllSemesters } from './DTI48MainGame';
 import styles from './DTI48.module.css';
 
+async function getMembers(semester?: string): Promise<readonly IdolMember[]> {
+  if (semester == null) return MembersAPI.getAllMembers(true);
+  return fetch(`/members-archive/members-${semester}.json`)
+    .then((response) => response.json())
+    .then((json) => json.members);
+}
+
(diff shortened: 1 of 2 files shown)


frontend/src/Games/DTI48/DTI48.tsx
```diff
@@ -1,22 +1,38 @@
...
+}
+
+async function getMembersFromAllSemesters(): Promise<MembersFromAllSemesters> {
+  const [currentSemesterMembers, sp21Members] = await Promise.all([
+    getMembers(),
+    getMembers('sp21')
+  ]);
+  return {
+    'Current Semester': currentSemesterMembers,
+    'Spring 2021': sp21Members
```
henry-li-06, line 21: Is there a way to not hardcode `sp21Members`, `'sp21'`, and `'Spring 2021'`

henry-li-06 (commented):
In `computeDTI49UpgradeChain` what happens to leads from previous semesters when the list of ids for each of the role leads changes. E.g. Jagger is former dev lead so his role should be a `lead` but `DEV_LEAD_NETIDS` for this semester should be Scott and Enoch.

SamChou19815:
> In `computeDTI49UpgradeChain` what happens to leads from previous semesters when the list of ids for each of the role leads changes. E.g. Jagger is former dev lead so his role should be a `lead` but `DEV_LEAD_NETIDS` for this semester should be Scott and Enoch.

Yes, I do plan to address that in another PR, potentially by making each lead have a subteam that can identify their types (e.g. devleads), then we can get rid of this hardcoding and make the tree generation fully automated.

frontend/src/Games/DTI48/DTI48.tsx
SamChou19815: There is probably a way to make the backend return an API like this, but I plan to address it in another PR where I add the login support for achieved members.

henry-li-06 (approved):

PR #139, opened 2021-09-01
Fix team page

Enochen:
### Summary 

Previously a header that said "Applications 👇" was added to the RoleSelector component, which was used in multiple places in the website. This resulted in the header appearing in unintended places.

The header has been factored out of the RoleSelector component and into the `apply` page ONLY, which is its intended behavior.

### Test Plan 

image
image


ngwattcos (approved):
lVerygtm

PR #141, opened 2021-09-02
[dti48] Remove hardcoding of leads netid

SamChou19815:
### Summary 

This PR stops using hardcoded leads netID. Instead, we use the newly added tag to `subteams` that denotes the type of leads. This ensures that we can auto-generate the old tree of org chart from the old members json without any versioning of hardcoded netids.

### Test Plan 

- Modified tests to not using exported hardcoded ID
- Play as dev, design, business to ensure things don't crash.


frontend/src/Games/DTI48/dti48-upgrade-chain.ts
```diff
@@ -43,13 +36,13 @@ function getDirectChildrenForDTI48OrganizationTree<M extends SimplifiedMemberFor
       }
       return [allPMs[allPMs.findIndex((pm) => pm.netid === member.netid) - 1]];
     }
-    case 'lead':
+    case 'lead': {
       // Design Leads have no children :(
-      if (DESIGN_LEAD_NETIDS.includes(member.netid)) return [];
-      if (BUSINESS_LEAD_NETID === member.netid) {
+      if (member.subteams.includes('design-leads')) return [];
```
henry-li-06, line 41: Should we create constants for `design-leads`, `business-leads`, etc

SamChou19815: added

henry-li-06 (approved):
LGTM

PR #142, opened 2021-09-02
[dti48] Load members from all semesters from backend

SamChou19815:
### Summary 

Move the data to the backend and fetch from backend. This allows the backend to later use the archived members json to decide who can login once the old members are removed.

### Test Plan 

Loads without issues

Screen Shot 2021-09-02 at 16 50 13


henry-li-06 (commented):
Seems like DTI48 isn't loading for me on the Netlify preview

SamChou19815:
@henry-li-06 this looks like a configuration for me. Looking at the logs [here](https://app.netlify.com/sites/cornelldti-idol/deploys/61313987e5971f000790656a), it appears to me that the server build only includes the build folder, which causes the function to fail due to missing `members-archive` folder. Unfortunately I don't have permission to fix that...

To solve this problem, we could either try to configure it or just import the json in the src folder. If we import, then we do have to hardcode things like `Spring 2021`, but I think it's probably much simpler.

henry-li-06:
I'll take a look at the configurations and see if I can fix that

henry-li-06 (approved):
LGTM

PR #144, opened 2021-09-03
Allow past members to signin

SamChou19815:
### Summary 

This PR implements signin support for past members that have been removed from `members` collection, so that they can continue access the site, and most importantly, play DTI48. After this PR, the project to allow past members to play DTI48 is done.

It's achieved by making the members stored inside the archive as a fallback.

### Test Plan 

1. Temporarily remove yourself in the members collection.
2. Try to sign in. You should be able to successfully signin.
3. Try to play DTI48 as yourself in spring 2021. It should still be the default choice.

### Notes

It's important to note that this does open up some security issues, although not that critical. For example, past leads will be still be able to have admin permissions. In the future,  I plan to attach an optional flag `isPastMember` so that we can distinguish current and past members by looking at the returned `IdolMember` object.

### DIFF
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -72,7 +72,7 @@ const loginCheckedHandler =
       res.status(440).json({ error: 'Not logged in!' });
       return;
     }
-    const user = await MembersDao.getMember(userEmail);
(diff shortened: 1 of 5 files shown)


henry-li-06 (approved):
LGTM

PR #147, opened 2021-09-16
Factor out page templates to prepare for Next.js migration

SamChou19815:
### Summary 

Next.JS routing is [based on filesystem](https://nextjs.org/docs/routing/introduction). For common components of all pages, it [uses the template defined in `src/pages/_app`](https://nextjs.org/docs/advanced-features/custom-app).

To prepare the eventual switch to this model, I refactored `src/App.tsx` to have a `AppContent` component with signature `({children}) => JSX.Element`. This will become the component in `src/pages/_app` when we migrate to Next.JS

### Test Plan 

Site still loads. You can login and move around pages.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { ReactNode, useEffect } from 'react';
 import './App.css';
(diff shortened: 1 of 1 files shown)


henry-li-06 (approved):
LGTM

PR #153, opened 2021-09-23
Flatten directories

SamChou19815:
### Summary 

Flattern directories so that it's easier to find where files are without expanding a lot of folders.

### Test Plan 

Everything still compiles.


henry-li-06 (approved):
LGTM

PR #161, opened 2021-09-28
Add zing project (for realsies)

Enochen:
### Summary 

Adds Zing project with info+assets provided by @bahartehrani 

Adds conditional rendering of header imgs (IDOL and Zing both are lacking header imgs right now)

### Test Plan 
image

image

### Breaking Changes 


SamChou19815 (changes requested):
Can be get graphics that have the same size as other ones?

Enochen:
> Can we get graphics that have the same size as other ones?

Yep, I've updated with a card for Zing that is the same size as the others. Screenshot has been updated to reflect this

PR #163, opened 2021-09-28
Move react components into `components` folder

SamChou19815:
### Summary 

Another mass moving refactor PR for Next.js migration.
Next.js expects all the pages routes to be in `src/pages`. While it doesn't mandate that all components be in `src/components`, it doesn't feel logically correct when the entry points are in a more nested folder than say `Forms`. Therefore, this PR migrates so that the directory structure is like:

```
src
 - API/
 - components/
 - static/
 - Other toplevel simple stuff like utils, firebase def etc.
```

### Test Plan 

CI. Everything should still compile.


frontend/src/utils.ts
```diff
@@ -18,3 +20,78 @@ export const getRoleDescriptionFromRoleID = (role: Role): RoleDescription => {
...
+
+  // Teams
+  static teamEditError: EventEmitter<{
+    headerMsg: string;
+    contentMsg: string;
+  }> = new EventEmitter();
+
+  // Sign-in code
+  static signInCodeCreated: EventEmitter<void> = new EventEmitter();
+}
```
SamChou19815, line 97: Also moved those scattered util classes into `utils.ts`

henry-li-06 (approved):
LGTM

PR #167, opened 2021-09-29
Migrate IDOL frontend to next.js

SamChou19815:
### Summary 

This PR bases on previous work and finally migrates the IDOL frontend to next.js. This unblocks future work to merge together  IDOL frontend and team website into a single stack, which allows us to do stuff like like preview using the same components.

### Test Plan 

CI preview. Every page should work fine.

### Notes

If you run it locally page transition might be slow. This is expected since next.js only compiles a page when you first visited it. It shouldn't be a problem in prod.


frontend/src/pages/admin/edit.tsx
```diff
@@ -0,0 +1,3 @@
+import AddUser from '../../components/Admin/AddUser';
+
+export default AddUser;
```
henry-li-06, line 3: Should we just move the `AddUser` component to this file?

henry-li-06: Nevermind, makes sense for consistency

henry-li-06 (approved):
LGTM

PR #168, opened 2021-09-29
Jest backend testing

henry-li-06:
### Summary 

Setup testing for IDOL backend and implemented testing for `MembersDao`.

### Test Plan 

`yarn test` in CI


backend/package-lock.json
```diff
@@ -0,0 +1,2882 @@
+{
```
SamChou19815, line 1: uhh remove this since we use yarn

frontend/package-lock.json
```diff
@@ -0,0 +1,1594 @@
+{
```
SamChou19815, line 1: also this

backend/tests/MembersDao.test.ts
```diff
@@ -0,0 +1,55 @@
+const MembersDao = require('./../src/dao/MembersDao.ts').default;
+const jaggerData = require('./data/jagger-profile.json');
```
SamChou19815, line 2: use import instead of require

SamChou19815 (approved):

PR #171, opened 2021-10-01
Fix backend testing config 

henry-li-06:
### Summary 
Fixed the config for jest backend testing, the original config for testing the backend wasn't finding any tests.

Changes:
- `testMatch`: changed from `<rootDir>/backend/**/*.test.ts` to `<rootDir>/**/*.test.ts'`
- `backend/tests/setup.ts`: created setup file to configure `.env` when running tests locally 

Also added Firebase private keys to repo secrets so that Firebase can be configured during test workflow 

### Test Plan 
Test step now runs `IDOL Backend Tests` in CI checks and pass

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -26,6 +26,10 @@ jobs:
       - name: Format
         run: yarn format:check
       - name: Test
+        env:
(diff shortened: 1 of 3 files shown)


SamChou19815 (approved):
LGTM

PR #172, opened 2021-10-04
Bump dependencies

SamChou19815:
### Summary 

Bump dependencies and regenerate `yarn.lock`. Hopefully it can help solve some yarn issues.

### Test Plan 

CI. Every preview site should still be fine.


SamChou19815:
#173 is fine, which probably means that some dependency really mess up in this one...

backend/package.json
```diff
@@ -4,38 +4,35 @@
   "description": "The backend for the DTI Internal DTI Organization Logic",
   "main": "index.js",
   "scripts": {
-    "start": "./node_modules/.bin/netlify-lambda serve build",
-    "dev": "ts-node --transpile-only src/api.ts",
+    "start": "netlify-lambda serve build",
+    "dev": "esr src/api.ts",
```
SamChou19815, line 8: Use esbuild runner to run it much faster than `ts-node`

netlify.toml
```diff
@@ -0,0 +1,2 @@
+[functions]
+  node_bundler = "esbuild"
```
SamChou19815, line 2: Switch the function bundler to esbuild. See https://www.netlify.com/blog/2021/04/02/modern-faster-netlify-functions/. The default one seems to have some bugs on dynamic requires.

henry-li-06: Is this what was causing the build issues for the netlify functions?

SamChou19815: Yeah I believe. I look into the files that failed and it contains a dynamic require.

henry-li-06 (approved):
LGTM

PR #177, opened 2021-10-06
Team credits form

morganbelous:
### Summary

This pull request is the first step towards implementing tracking team event credits on IDOL. This form replaces the current Google Form that tracks team event attendance (https://docs.google.com/forms/d/e/1FAIpQLSeZJ9xABYjFqgD7Ethwkr5MnbqWVm20tM3B06YqtK_Px7HpgQ/viewform). It allows DTI members to submit a photo of their attendance at a team event and request a team event credit from the leads.

- [x] created team event credit form
- [ ] connect frontend to backend

### Test Plan

no_event_selected

event_selected

### Notes 

- Hard coded 1 team event for show
- Currently only allows 1 image upload at a time
- I’m unsure how to handle the number of event credits. Best case scenario is IDOL keeps track of the credit amount for each event and does all computations behind the scenes. However, there are some things (like interviews, coffee chats, attending club fest, etc.) that can have different credit amounts. For example, if someone attends club fest, then 1 hour = 0.5 credits. It doesn’t seem right to have users submit a form for each hour attended if they don’t have the ability to enter the credit value. Any ideas


morganbelous:
> image
> I don't think we need this role selector again since that info is already associated with a member

its read-only based on what IDOL has as their role but yeah I agree I can just make it a label or just remove it all together

henry-li-06:
I think the credits for the team event can be fixed (e.g. 0.5 credits associated w/ coffee chats). Members should be able to make multiple requests for receiving credit for an event (e.g. someone attending multiple coffee chats).

morganbelous:
> I think the credits for the team event can be fixed (e.g. 0.5 credits associated w/ coffee chats). Members should be able to make multiple requests for receiving credit for an event (e.g. someone attending multiple coffee chats).

but what if they attend two hours of club fest? that would be 2 requests for the same event?

henry-li-06:
> > I think the credits for the team event can be fixed (e.g. 0.5 credits associated w/ coffee chats). Members should be able to make multiple requests for receiving credit for an event (e.g. someone attending multiple coffee chats).
> 
> but what if they attend two hours of club fest? that would be 2 requests for the same event?

hmm good point... i don't think that would be the worst solution... on the other hand another idea might be to make the number of credits for each event be an array, so like the possible credits you could get for the team event

riyajaggi:
> > > I think the credits for the team event can be fixed (e.g. 0.5 credits associated w/ coffee chats). Members should be able to make multiple requests for receiving credit for an event (e.g. someone attending multiple coffee chats).
> > 
> > 
> > but what if they attend two hours of club fest? that would be 2 requests for the same event?
> 
> hmm good point... i don't think that would be the worst solution... on the other hand another idea might be to make the number of credits for each event be an array, so like the possible credits you could get for the team event

I think this is a good idea! We could perhaps save a range for each event with the min credit (default) and the max credit (which could be optional) you can earn, since events like club fest can be capped, but on the other hand you can do has many coffee chats as possible.

airaagrawal1:
Would it be useful to have an optional 'Description' field, so that members can put who they coffee chatted with/more details of the event they went to? This would also give more information to the leads approving credit request.

henry-li-06:
image
This is kinda nitpicking but do you think we should also display the number of credits for the team event after it's been selected (maybe somewhere underneath or next to the name)?

PR #178, opened 2021-10-06
Add release branch for CI build workflow 

henry-li-06:
### Summary 
Change config so that CI build workflow runs on PR to `release` branch
### Test Plan 
N/A

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
+      - release
   pull_request:
 
 jobs:


SamChou19815 (approved):
LGTM but not sure what's the purpose.

henry-li-06:
> LGTM but not sure what's the purpose.

Making new `release` branch for prod branch instead of `master`. #177 was the motivation for this. If more people start using IDOL, my thought was that we wouldn't want people seeing half-implemented features in prod.

SamChou19815:
> > LGTM but not sure what's the purpose.
> 
> Making new `release` branch for prod branch instead of `master`. #177 was the motivation for this. If more people start using IDOL, my thought was that we wouldn't want people seeing half-implemented features in prod.
> 
> I've already changed to build/deploy config on Netlify
> 
> @SamChou19815 any reason not to do this?

We could instead putting half finished features behind a flag, like what i did in https://github.com/cornell-dti/course-plan/pull/521 for course plan. I feel like maintaining another deployment target can introduce a lot of dev overhead, especially when we need to change the format of the database records.
On public facing products like courseplan, such overhead is inevitable and we have to live with it. However, IDOL is completely internal facing. We could afford to ship some broken stuff as long as we can quickly fix it.

PR #187, opened 2021-10-09
Add Expiry Date for Sign-in Code

riyajaggi:
### Summary 

This PR allows admins to set an expiry date for attendance sign-in form. Members can only sign-in before the expiry date.

**Functionality**
- Admins can create sign-in form and set expiry date
- Members can sign-in before expiry date and their attendance is recorded
- Members are not able to sign-in after the code is expired
- Admins can view list of sign-in code with the expiry dates.

**Code**
- Used [react-datepicker](https://github.com/Hacker0x01/react-datepicker/) for expiry date input
- New _required_ field in database to store expire date as `expireAt: number`

### Test Plan 

Tested locally.

### Notes 

- The react-datepicker by itself doesn't disable past times (it does for dates). Example: if it is Oct 8, 5pm then admins can still set Oct 8, 4pm as expiry date. so need to fix this
- The design may not be consistent.

### Breaking Changes 

- DB changes:`expireAt: number` added to `DBSignInForm` and `SignInForm`
- New package: react-datepicker

### Screenshots
**Admin SignIn Form Creator**
image

**Adding Expiry Date**
image

image

**Member sign-ins after code has expired**
image


backend/src/dao/SignInFormDao.ts
```diff
@@ -10,22 +10,24 @@ export default class SignInFormDao {
     if (!formRef.exists) throw new NotFoundError(`No form with id '${id}' found.`);
     const form = formRef.data();
     if (form == null) throw new NotFoundError(`No form content in form with id '${id}' found.`);
+    const signedInAtVal = Date.now();
+    if (form.expireAt <= signedInAtVal)
+      throw new PermissionError(`User is not allowed to sign into expired form with id '${id}.`);
```
henry-li-06, line 15: I think it might be better to have this as `BadRequestError`, doesn't seem to quite fit a permissions error.

backend/src/signinformAPI.ts
```diff
@@ -6,17 +6,26 @@ import { PermissionsManager } from './permissions';
...
 
 export const createSignInForm = async (
   id: string,
+  expireAt: number,
   user: IdolMember
-): Promise<{ id: string; createdAt: number }> => {
+): Promise<{ id: string; createdAt: number; expireAt: number }> => {
   if (!PermissionsManager.canEditSignIn(user)) {
     throw new PermissionError("You don't have permission to create a sign-in form!");
   }
```
henry-li-06, line 26: I think it might be good to check that expiry time for the form isn't in the past here

backend/src/dao/SignInFormDao.ts
riyajaggi: Done!

backend/src/signinformAPI.ts
riyajaggi: Done! also displaying it on frontend
image

PR #191, opened 2021-10-16
Fix issue with sign in form creator page forever loading 

henry-li-06:
### Summary 
On page reload, the page would be stuck on the loading symbol.

image

Seemed like it was an issue with async commands executing in the wrong order in `UserProvider.tsx`
```
setUser({ email: getUserEmail(userAuth) });
cachedUser = userAuth;
```
where `setUser` was completing before `cacheduser = userAuth` causing `cachedUser` to be `null`. 

The solution in this PR is just to chain the commands with `.then()` so the assignment always finishes first. 

### Test Plan 
Check on netlify preview that sign in forms load after refresh

### DIFF
--- a/frontend/src/components/Common/UserProvider.tsx
+++ b/frontend/src/components/Common/UserProvider.tsx
@@ -22,6 +22,8 @@ export const useUserEmail = (): string => {
 
 let cachedUser: User | null = null;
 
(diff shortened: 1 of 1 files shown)


riyajaggi (approved):
LGTM!

PR #197, opened 2021-10-30
[easy] A small framework to hide entry point if user is not admin

SamChou19815:
### Summary 

Add `unstable: true` to a navigation card data and they will be hidden if you are not admin.

### Test Plan 

Added an example hidden page for admin. It will not be visible if you are not admin.

Screen Shot 2021-10-30 at 17 14 59


JacksonStaniec (commented):
I didn't even know this was an option. It's definitely a good idea.

henry-li-06 (approved):
LGTM, this is definitely helpful for hiding incomplete subpages.

PR #199, opened 2021-11-01
Make teams APIs no longer depend on team collection

SamChou19815:
### Summary 

There are data sync issues between team collection and members collection. We should only have one source of truth, which should be members data since that's directly shipped to the team website.
This diff changes the backend API to use members' data only. Teams are derived from members' data. Right now they are inefficient, but eventually we should move away from editing a team, but instead edit the team data of each individual member.

### Test Plan 

https://user-images.githubusercontent.com/4290500/139607097-f507affb-4817-49bc-9273-dfe7d94e0aa4.mov


riyajaggi (approved):
This is a great way to think about our data collections! It can definitely help keep our backend simpler and avoid issues in the future as our features and collections grow. 

Looks good to me!

PR #200, opened 2021-11-01
Edit team events

morganbelous:
### Summary

This pull request is the next step towards implementing tracking team event credits on IDOL, and the first part of the admin team events page. The form allows admin to create a team event. Below the form, admin can view all team events, view their details, or delete an event.

- [x] implemented create event form
- [x] view all events
- [x] view event details
- [x] delete event
- [ ] view members for each event
- [ ] edit events
- [ ] connect frontend to backend

### Test Plan

team-events-admin

team-events-details

delete-event

### Notes

- the design of the event details view is kinda ugly if anyone has any suggestions
- need to add uuid to type to match backend


henry-li-06:
Should "Event Date" be an optional field? For coffee chat kinda events there probably doesn't need to be a date but not sure if there's any similar cases...

henry-li-06:
For the event details view, what are your thoughts on expanding the cards to show the details instead of a modal?

JacksonStaniec:
> For the event details view, what are your thoughts on expanding the cards to show the details instead of a modal?

I second this idea. The modal fits well when deleting a team, but when viewing details I think it would seem less interrupting to expand instead, especially since there are only a couple of fields.

riyajaggi:
I noticed now the alignment of the create/save buttons is different in different forms (both admin/users). I believe the older forms have it on the left side and the newer ones on the right. What do you think about having it consistent on all forms (any one side)?

frontend/src/components/Admin/EditTeamEvents.tsx
```diff
@@ -0,0 +1,170 @@
...
+  return (
+    <div>
+      <Form
+        style={{
+          width: '60%',
+          alignSelf: 'center',
+          margin: 'auto',
+          padding: '3rem 0 5rem 0'
+        }}
+      >
```
riyajaggi, line 92: If it is easier, we can also keep styles in a separate css file. What do you think?

morganbelous:
> Should "Event Date" be an optional field? For coffee chat kinda events there probably doesn't need to be a date but not sure if there's any similar cases...

hmmm I guess you're right. do you think event date is worth having at all? not sure if saving that information even adds much.

morganbelous:
> For the event details view, what are your thoughts on expanding the cards to show the details instead of a modal?

when you say 'expanding the cards', do you mean just making the cards larger to fit the details, or dynamically routing each event to its own page? I did think about having separate pages for each event but wasn't sure how to implement it. Can look into it though

henry-li-06:
> > Should "Event Date" be an optional field? For coffee chat kinda events there probably doesn't need to be a date but not sure if there's any similar cases...
> 
> hmmm I guess you're right. do you think event date is worth having at all? not sure if saving that information even adds much.

I think it might still be worth having to put a date on certain events that have them, it's still pretty relevant metadata. Alternatively, what if we make the date for events like coffee chats to be like "repeating" or something like that to indicate that it can be whenever and can happen on multiple occasions. Not sure if repeating is the right word here tho

henry-li-06:
> > For the event details view, what are your thoughts on expanding the cards to show the details instead of a modal?
> 
> when you say 'expanding the cards', do you mean just making the cards larger to fit the details, or dynamically routing each event to its own page (keeping in mind we still need to add showing all pending/approved members for each event)? I did think about having separate pages for each event but wasn't sure how to implement it. Can look into it though

By expanding the cards, I meant when you click on the card, instead of the modal popping up, the card gets larger and shows the details. This is sorta inspired by the way Google Calendar handles looking at event details. Obviously, we wouldn't be able to completely copy that but something similar. 

It's also definitely a good point about pending/approved members. Maybe we should have dynamic routing for that. We can probably just route to like `/admin/team-events/{team-event-id}` to look at the team event with `team-event-id` and deal with updating the event details or approving requests on that page.

henry-li-06 (approved):
LGTM, the new subpage for seeing team event details looks great!

PR #204, opened 2021-11-04
Setup firestore listeners for members and approved members collection

SamChou19815:
### Summary 

By listening these data in real-time, we can avoid API calls to the backend here and there, and make the dev experience a lot better by avoiding a lot of async calls. See https://firebase.google.com/docs/firestore/query-data/listen

I migrated some of the components to use the new API, but more remains unmigrated to keep this PR small.

### Test Plan 

Affected pages (shortout, site deployer) work without issue.

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -1,7 +1,6 @@
 module.exports = {
   webpack5: true,
   typescript: { ignoreBuildErrors: true },
-  swcMinify: true,
   async rewrites() {
     return [
       {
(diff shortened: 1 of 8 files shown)


henry-li-06 (commented):
Sorry this took so long to get reviewed

Looks good for the most part but for whatever reason the site deployer page isn't loading on the Netlify preview, not sure if it has anything to do with the changes in this PR though

SamChou19815:
@henry-li-06 should be fixed now. `swcMinify` in next.js is still buggy. Removing it seems to solve the problem...

henry-li-06 (approved):
Looks good now

PR #205, opened 2021-11-09
Refactor repeated code into one MemberSearch component

airaagrawal1:
### Summary 

Refactored code so anytime user is searching for a member, the matchChecker is checking not-null data values for members. Code was repeated in 4 locations, so I replaced all of them. 

### Test Plan 

Does not break any functionality.


package.json
```diff
@@ -43,5 +43,8 @@
     "**/*.{ts,js,tsx,scss,css,html,md}": [
       "yarn prettier --write"
     ]
+  },
+  "dependencies": {
```
JacksonStaniec, line 47: Why did this need to be added to the dependencies?

package.json
```diff
@@ -43,5 +43,8 @@
...
+    "@types/node": "^16.11.4"
   }
```
JacksonStaniec, line 49: Why did this need to be added to the dependencies?

JacksonStaniec (commented):
I have also got to assume there's some other way to update the lock file -- 40,000 additions is a bit unusual.

frontend/src/components/Common/Search.tsx
```diff
@@ -48,6 +49,21 @@ type CustomSearchProps<T> = {
...
+    (member.firstName && member.firstName.toLowerCase().startsWith(queryLower)) ||
+    (member.lastName && member.lastName.toLowerCase().startsWith(queryLower)) ||
+    (member.role && member.role.toLowerCase().startsWith(queryLower)) ||
+    ((member.firstName &&
+      member.lastName &&
+      `${member.firstName.toLowerCase()} ${member.lastName.toLowerCase()}`.startsWith(
+        queryLower
+      )) as boolean)
+  );
+}
```
henry-li-06, line 65: I think this can be made into a React component instead of just a function for the match checker for the `CustomSearch` component.

i.e.
```
const MemberSearch = ({query, member}) => {
    return <CustomSearch ... />
}
```

package.json
henry-li-06: +1

SamChou19815 (changes requested):
package-lock.json shouldn't be commited, because we should be using  yarn instead of NPM

package.json
airaagrawal1: @henry-li-06 I wasn't sure about this dependency and the lock file - these were added after we tried fixing my npm and yarn modules.

frontend/src/components/Common/Search.tsx
airaagrawal1: image

I wasn't sure if I could make it a React component because the similar code is just in the matchChecker piece of the component, but source, resultRenderer, and selectCallback differ in each instance.

package.json
henry-li-06: I think this dependency can probably be deleted, `package-lock.json` should definitely be deleted though

frontend/src/components/Common/Search.tsx
henry-li-06: Oh good catch, maybe rename it to something like `memberMatchChecker` then to better follow naming conventions since it'd just be function

yarn.lock
```diff
@@ -4,26 +4,26 @@
 
 "@babel/code-frame@7.12.11":
   version "7.12.11"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.11.tgz#f4ad435aa263db935b8f10f2c552d23fb716a63f"
+  resolved "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.11.tgz"
```
SamChou19815, line 7: all changed here should also be reverted, since they look like the result of the latest npm trying to generate `yarn.lock`

package.json
airaagrawal1: Sounds good, what about the edits to `yarn.lock`?

frontend/src/components/Admin/EditTeam.tsx
```diff
@@ -291,22 +278,9 @@ class EditTeam extends React.Component<Record<string, unknown>, EditTeamState> {
...
-                            ((member.firstName &&
-                              member.lastName &&
-                              `${member.firstName.toLowerCase()} ${member.lastName.toLowerCase()}`.startsWith(
-                                queryLower
-                              )) as boolean)
-                          );
-                        }}
+                        matchChecker={(query: string, member: Member) =>
+                          MemberSearch(query, member)
+                        }
```
SamChou19815, line 283: you can even shorten it to `matchChecker={MemberSearch}`

SamChou19815 (approved):

PR #207, opened 2021-11-10
Setup firestore listeners for admin emails collection

SamChou19815:
### Summary 

Continues from #204. It makes the admin navigation cards much smoother (no more sudden refresh).

### Test Plan 

Visit admin cards page and various other admin pages. No permission errors should be observed.


frontend/src/components/Admin/SiteDeployer.tsx
```diff
@@ -30,21 +26,8 @@ const SiteDeployer: React.FC = () => {
...
-            loadPullRequests();
-          } else {
-            Emitters.generalError.emit({
-              headerMsg: 'Access Denied',
-              contentMsg: `Insufficient permissions.`
-            });
-          }
-        });
-    }
-  }, [selfMember]);
```
SamChou19815, line 47: We guard at the stage of navigation cards, so non-admin won't even enter the page.

henry-li-06: Are nav cards that lead to admin pages just not shown to non-admin users now?

SamChou19815: Yep. See changes in frontend/src/pages/admin/index.tsx

henry-li-06 (approved):

PR #213, opened 2021-11-17
Fix frontend tests under Node 16

SamChou19815:
### Summary 

Node 16 starts to fail program when there is an unhandled promise. This happens in our frontend tests a lot since we are trying to fetch some API that we don't have enough permission, or when we are missing the context to provide info. This PR fixes all of the problems and make the tests to pass on Node 16.

### Test Plan 

`yarn test` on Node 16


riyajaggi (approved):
LGTM! All tests pass.

PR #217, opened 2021-11-19
Hide Admin NavItem if the user is not admin

SamChou19815:
### Summary 

Hide the admin navbar item so users without admin permission won't enter a confusing white page.

### Test Plan 

Login as non-admin to check.


henry-li-06 (approved):
LGTM

PR #220, opened 2021-11-24
Remove `getMember` and `APICache` API

SamChou19815:
## Summary

Remove these legacy APIs with firestore listeners.

## Test Plan

Affected page of user profile and team event credit form works fine.

### DIFF
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -7,7 +7,6 @@ import { app as adminApp } from './firebase';
 import {
   allMembers,
   allApprovedMembers,
-  getMember,
   setMember,
   deleteMember,
   updateMember,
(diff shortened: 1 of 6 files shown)


henry-li-06 (commented):
Updating profile doesn't seem to be working, I just tried changing my graduation date and updated graduation didn't appear after reloading page

EDIT: doesn't seem to be working in prod for me either

SamChou19815:
> Updating profile doesn't seem to be working, I just tried changing my graduation date and updated graduation didn't appear after reloading page
> 
> EDIT: doesn't seem to be working in prod for me either

I suspect that's because the pronoun is empty so the frontend doesn't actually allow you to submit? I also noticed that, but adding some value for pronoun fixed it.

henry-li-06:
> I suspect that's because the pronoun is empty so the frontend doesn't actually allow you to submit? I also noticed that, but adding some value for pronoun fixed it.

Oh good catch, working now. We should probably add the required tag to the pronouns input, doesn't seem to have one atm
image

There also doesn't seem to be an error modal for when the profile data isn't valid. We can probably address both of these in a different PR.

henry-li-06 (approved):

PR #223, opened 2021-11-29
Test Team Events API 

airaagrawal1:
### Summary 

Tested API endpoints for Team Events. 

### Test Plan 

Included tests for admin users with permissions and users without permissions so that errors were thrown as needed.

### DIFF
--- a/backend/tests/TeamEventsAPI.test.ts
+++ b/backend/tests/TeamEventsAPI.test.ts
@@ -9,7 +9,7 @@ import {
 import jaggerData from './data/jagger-profile.json';
 
 const adminUser = { email: 'hl738@cornell.edu' } as IdolMember;
-const nonAdminUser = { email: 'aa2235@cornell.edu' } as IdolMember;
+const nonAdminUser = { email: 'pk457@cornell.edu' } as IdolMember;
 
 const testTeamEvent = {
   name: 'test',
@@ -30,12 +30,7 @@ test('created team event with permission', async () => {
 });
 
 test('does not create team event because user does not have permission', async () => {
-  try {
-    await createTeamEvent(testTeamEvent, nonAdminUser);
-    throw new Error();
-  } catch (error) {
-    expect(error).toEqual(new PermissionError('does not have permissions to create team event'));
-  }
+  expect(createTeamEvent(testTeamEvent, nonAdminUser)).rejects.toThrow(PermissionError);
 });
 
 const updatedTestTeamEvent = {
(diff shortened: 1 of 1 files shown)


backend/tests/TeamEventsAPI.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+test('does not get all team events because user does not have permission', async () => {
+  try {
+    await getAllTeamEvents(nonAdminUser).then((events) => {
+      expect(events.pop.name).toEqual('test');
+    });
+  } catch (error) {
+    expect(error).toEqual(new PermissionError('does not have permissions'));
+  }
+});
```
SamChou19815, line 85: Are we asserting the code in try throws or not? If we expect to throw the expected error, better add `throw new Error()` to the end of try block to make the intention clear.

```js
try {
  codeThatMayThrow();
  throw new Error(); // if code above not throws, will throw this
} catch (error) {
  // ensures error has the right type.
  expect(error).toEqual(new PermissionError('does not have permissions'));
}
```

SamChou19815: Or even better: https://jestjs.io/docs/expect#tothrowerror

airaagrawal1: I added `throw new Error()` where needed. For some reason, the implementation in the link was not identifying the error.

backend/tests/TeamEventsAPI.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+test('does not create team event because user does not have permission', async () => {
+  try {
+    await createTeamEvent(testTeamEvent, nonAdminUser);
+    throw new Error();
+  } catch (error) {
+    expect(error).toEqual(new PermissionError('does not have permissions to create team event'));
+  }
+});
```
henry-li-06, line 39: I think you can use the `toThrow()` function from jest to test for expected errors
ex.
```
expect(Promise.reject(createTeamEvent(testTeamEvent, nonAdminUser))).rejects.toThrow(
    Errors.PermissionError
);
```

backend/tests/TeamEventsAPI.test.ts
```diff
@@ -0,0 +1,88 @@
...
+import {
+  createTeamEvent,
+  deleteTeamEvent,
+  getAllTeamEvents,
+  updateTeamEvent
+} from '../src/team-eventsAPI';
+import jaggerData from './data/jagger-profile.json';
+
+const adminUser = { email: 'hl738@cornell.edu' } as IdolMember;
+const nonAdminUser = { email: 'aa2235@cornell.edu' } as IdolMember;
```
henry-li-06, line 12: So you should actually be an admin member (I forgot to do that when you joined the team). Probably should change to some other user

henry-li-06 (approved):
nice work! LGTM

Thanks for making the changes!!

PR #224, opened 2021-12-01
Add backend api for handling event proof images

riyajaggi:
### Summary 

This pull request implements the backend for handling proof images for Team Event Credit. It does the following:
- sets proof image
- get proof image
- get all proof images for a member
- delete a proof image
- delete all proof images for a member

The images are stored in Firebase as `eventProofs/${netID}/{$name}` where `netID` is for identifying the user and `name` is the event name for which proof is uploaded.

### Test Plan 

Not tested yet.

### DIFF
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -39,6 +39,13 @@ import {
   getAllTeamEvents,
   updateTeamEvent
 } from './team-eventsAPI';
+import {
+  allProofImagesForMember,
+  deleteProofImage,
+  deleteProofImagesForMember,
+  getProofImage,
+  setProofImage
+} from './team-events-imageAPI';
 
 // Constants and configurations
 const app = express();
@@ -205,6 +212,23 @@ loginCheckedGet('/getAllTeamEvents', async (_, user) => ({ events: getAllTeamEve
 loginCheckedPost('/updateTeamEvent', async (req, user) => updateTeamEvent(req.body, user));
 loginCheckedPost('/deleteTeamEvent', async (req, user) => ({
   team: await deleteTeamEvent(req.body, user)
+}));
+
+// Team Events Proof Images
+loginCheckedGet('/getEventProofImageSignedURL', async (req, user) => ({
+  url: await setProofImage(user, req.body.name)
+}));
+loginCheckedGet('/getEventProofImage', async (req, user) => ({
+  url: await getProofImage(user, req.body.name)
+}));
+loginCheckedGet('/getAllEventProofsForMember', async (_, user) => ({
+  url: await allProofImagesForMember(user)
+}));
+loginCheckedPost('/deleteEventProofImage', async (req, user) => ({
+  image: await deleteProofImage(user, req.body.name)
+}));
+loginCheckedPost('/deleteAllEventProofsForMember', async (_, user) => ({
+  image: await deleteProofImagesForMember(user)
 }));
 
 app.use('/.netlify/functions/api', router);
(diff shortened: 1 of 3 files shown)


backend/src/team-events-imageAPI.ts
```diff
@@ -0,0 +1,73 @@
...
+      ...image,
+      fileName: image.fileName.slice(image.fileName.indexOf('/') + 1)
+    }));
+
+  return images;
+};
+
+export const deleteProofImage = async (user: IdolMember, name: string): Promise<void> => {
+  const netId: string = getNetIDFromEmail(user.email);
+  const imageFile = bucket.file(`eventProofs/${netId}/${name}.jpg`);
```
JacksonStaniec, line 61: Should there be a quick check to make sure the file exists first? I'm never sure what the best practice is since if the file doesn't exist, then the outcome of "deleteProofImage" is the same.

JacksonStaniec (approved):
Looks great!

backend/src/api.ts
```diff
@@ -207,6 +214,23 @@ loginCheckedPost('/deleteTeamEvent', async (req, user) => ({
...
+}));
+loginCheckedGet('/getAllEventProofsForMember', async (_, user) => ({
+  url: await allProofImagesForMember(user)
+}));
+loginCheckedPost('/deleteEventProofImage', async (req, user) => ({
+  image: await deleteProofImage(user, req.body.name)
+}));
+loginCheckedPost('/deleteAllEventProofsForMember', async (_, user) => ({
+  image: await deleteProofImagesForMember(user)
+}));
```
henry-li-06, line 232: I don't think we need these as API endpoints. Images should only get deleted when a request is either approved or rejected which is handled by the team event API.

backend/src/team-events-imageAPI.ts
henry-li-06: Should be fine without the check, the purpose of deleting the proof image is just to save memory. From my other comment, proof images should only be deleted if request if approved/rejected. If it exists then we get rid of it, if it doesn't then it's also fine.

common-types/index.d.ts
```diff
@@ -81,3 +81,8 @@ interface TeamEvent {
   readonly attendees: IdolMember[];
   readonly uuid: string;
 }
+
+interface EventProofImage {
+  readonly url: string;
+  readonly fileName: string;
+}
```
henry-li-06, line 88: Hmm this is the same as the `ProfileImage` type... Does it make sense to make these the same type?

backend/src/api.ts
riyajaggi: Done!

common-types/index.d.ts
riyajaggi: I am not sure what fields would be needed for frontend integration other than these two. In that case, should I just combine both types together, or if there's a better way of doing it?

riyajaggi: Also, code for image processing is similar for both profile images and event images so I think that can be refactored out once the team events are integrated

henry-li-06:
LGTM, nice work! Should already be approved 🚢🚢🚢

common-types/index.d.ts
morganbelous:
```
 type TeamEventAttendance = {
  memberEmail: string;
  teamEventName: string;
  hoursAttended?: string;
  image: string;
};
```

The way I implemented it on the frontend was when the form for requesting a team event credit is submitted, an object of type TeamEventAttendance is sent to the backend. I feel like information should be attached to the image, like the name of the team event name, hours attended (if it applies), etc. Maybe we can make the image attribute of the type TeamEventAttendance be an object that contains the url and filename? Guess we can figure it out when we integrate the frontend and backend

PR #238, opened 2022-02-10
Candidate Decider admin page 

henry-li-06:
### Summary 

This pull request is the first step towards integrating Candidate Decider into IDOL. It adds an admin page for creating new "instances" of Candidate Decider where IDOL admins can upload a csv file of the application results spreadsheet to create a new instance. 

image

The uploaded csv file is parsed into `CandidateDeciderInstance` type: 
```
interface CandidateDeciderRating {
  readonly reviewer: IdolMember;
  readonly rating: number;
}

interface CandidateDeciderComment {
  readonly reviewer: IdolMember;
  readonly comment: string;
}

interface CandidateDeciderCandidate {
  readonly responses: any[];
  readonly id: number;
  ratings: CandidateDeciderRating[];
  comments: CandidateDeciderComment[];
}

interface CandidateDeciderInstance {
  readonly name: string;
  readonly headers: string[];
  readonly candidates: CandidateDeciderCandidate[];
  readonly uuid: string;
  isOpen: boolean;
}

```

### Test Plan 
👀

### DIFF
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -86,3 +86,11 @@ interface EventProofImage {
   readonly url: string;
   readonly fileName: string;
 }
+
+interface CandidateDeciderInstance {
+  readonly name: string;
+  readonly headers: string[];
+  readonly candidates: any[];
+  readonly uuid: string;
+  isOpen: boolean;
+}
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -87,10 +87,27 @@ interface EventProofImage {
   readonly fileName: string;
 }
 
+interface CandidateDeciderRating {
+  readonly reviewer: IdolMember;
+  readonly rating: number;
+}
+
+interface CandidateDeciderComment {
+  readonly reviewer: IdolMember;
+  readonly comment: string;
+}
+
+interface CandidateDeciderCandidate {
+  readonly responses: any[];
+  readonly id: number;
+  ratings: CandidateDeciderRating[];
+  comments: CandidateDeciderComment[];
+}
+
 interface CandidateDeciderInstance {
   readonly name: string;
   readonly headers: string[];
-  readonly candidates: any[];
+  readonly candidates: CandidateDeciderCandidate[];
   readonly uuid: string;
   isOpen: boolean;
 }
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -98,7 +98,7 @@ interface CandidateDeciderComment {
 }
 
 interface CandidateDeciderCandidate {
-  readonly responses: any[];
+  readonly responses: string[];
   readonly id: number;
   ratings: CandidateDeciderRating[];
   comments: CandidateDeciderComment[];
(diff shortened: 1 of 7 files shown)


common-types/index.d.ts
```diff
@@ -86,3 +86,28 @@ interface EventProofImage {
...
+  readonly rating: number;
+}
+
+interface CandidateDeciderComment {
+  readonly reviewer: IdolMember;
+  readonly comment: string;
+}
+
+interface CandidateDeciderCandidate {
+  readonly responses: any[];
```
riyajaggi, line 101: I see a GitHub warning for this type. Why do we choose to use `any[]`?

frontend/src/components/Admin/AdminCandidateDecider.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const instance = {
+      uuid: '',
+      name,
+      headers,
+      candidates: responses.map((res, i) => ({ id: i, responses: res, comments: [], ratings: [] })),
+      isOpen: true
+    };
+    setSuccess(true);
+    console.log(instance);
+    console.log('FORM SUBMITTED');
```
riyajaggi, line 62: There are few console.log() statements in this PR. Should these be removed/commented out before merging?

frontend/src/components/Admin/AdminCandidateDecider.tsx
```diff
@@ -0,0 +1,140 @@
...
+      ) : (
+        <div id={styles.itemsContainer}>
+          <Card.Group>
+            {instances.map((instance) => (
+              <Card
+                color={instance.isOpen ? 'green' : 'red'}
+                // href={`candidate-decider/${instance.uuid}`}
+                key={instance.uuid}
+              >
+                {console.log(instance.isOpen)}
```
riyajaggi, line 121: ^

common-types/index.d.ts
henry-li-06: This field is the responses from the candidates so the questions depend from form to form. I wasn't sure if there was a better way to do this to avoid `any`

frontend/src/components/Admin/AdminCandidateDecider.tsx
henry-li-06: These are placeholders for now since the API hasn't been implemented yet

frontend/src/components/Admin/AdminCandidateDecider.tsx
henry-li-06: I'll get rid of this once, I was testing something and forgot to delete

common-types/index.d.ts
riyajaggi: got it! In this case, `any` seems suitable

riyajaggi (approved):
LGTM!

common-types/index.d.ts
Enochen: These *should* be all strings in my experience, but I recommend creating a union type rather than using `any` to narrow down all possibilities if we add something other than string.

PR #241, opened 2022-02-14
Testing as non-admin user locally

henry-li-06:
### Summary 

This PR adds a feature that allows testing as a non-admin user locally. You can do this by setting `allowAdmin` in `frontend/src/environmen.ts` to `false` and you will effectively be a non-admin user on IDOL. 

### Test Plan 

N/A


airaagrawal1 (approved):
LGTM!

PR #242, opened 2022-02-15
Non admin testing locally 

henry-li-06:
### Summary 

This PR addresses the issue from the work session today.  When testing locally (aka when `isProduction` is false), all nav cards including those with `adminOnly` set to true will show up even if `allowAdmin` is false. This is to allow local testing as a non-admin user for incomplete features but does not affect prod. 

### Test Plan 
N/A


riyajaggi (approved):
Perfect!

PR #247, opened 2022-02-23
Generate fake data for backend testing.

JacksonStaniec:
### Summary 

This PR adds `createData.js` that uses Faker.js to create mock `IdolMembers`, `TeamEventAttendance`, and `TeamEvent` objects for the backend test suite. Hardcoded mock data is replaced by generated mock data wherever possible.

### Test Plan 
Test cases behave as expected.


backend/tests/MembersDao.test.ts
```diff
@@ -26,8 +31,8 @@ test('Add new member', () => {
 });
 
 test('Get member from past semester', () =>
-  MembersDao.getCurrentOrPastMemberByEmail(jaggerData.email).then((pastMember) =>
-    expect(pastMember).toEqual(jaggerData)
+  MembersDao.getCurrentOrPastMemberByEmail(jaggerData.email).then(
+    (pastMember) => expect(pastMember).toEqual(jaggerData) // cannot deprecate jaggerData yet
```
JacksonStaniec, line 35: This is the only place left that uses the hardcoded mock user `jaggerData`. This test for the `getCurrentOrPastMemberByEmail` function relies on accessing a json archive of previous members. Is it worth it to automatically add a generated user to the archive and remove it after the test runs, or is it fine to keep the hardcoded mock data in this case?

henry-li-06 (commented):
did you get a chance to look into fixing `update existing team event with permission` test

JacksonStaniec:
> did you get a chance to look into fixing `update existing team event with permission` test

It seems like since the Idol member in `TeamEventAttendance` objects is stored in Firestore as a document reference to the members collection, and since we have been using fake members, its failing because there is no actual member in the database and so the retrieved object from the database has `member: undefined`. This is probably the case for `TeamEventsDao` tests too. We could either:
- first put the fake Idol member in the database, and remove it afterwards like what happens in the `MembersDao` tests
- compare specific fields of the expected and received object to check for equality, instead of the objects themselves.

henry-li-06:
> * database

I think the first option seems pretty reasonable

henry-li-06 (approved):
let's forget about fixing that last test for now

PR #255, opened 2022-03-03
Candidate Decider - Fix csv parsing 

henry-li-06:
### Summary 
Fix csv parsing for Candidate Decider so that questions and responses are properly aligned. 

image

### Test Plan 
👀


airaagrawal1 (commented):
An answer and its corresponding bolded question have a space between them. However, the answer does not have space after it and before the next question. I think it would be less confusing to read if we put another space after each answer.

henry-li-06:
> An answer and its corresponding bolded question have a space between them. However, the answer does not have space after it and before the next question. I think it would be less confusing to read if we put another space after each answer.

@airaagrawal1 
image

airaagrawal1 (approved):
LGTM!

PR #268, opened 2022-03-31
Refactor `SignInFormDAO` and `signinformAPI`

JacksonStaniec:
### Summary 

This PR moves validation logic from the sign-in form DAO to its API. 
- [x] checking if the user exists
- [x] checking if the form exists
- [x] checking if the form is expired

There is also a bug fix with boolean Promises that resulted in always-true or always-false values (explanation [here](https://www.aleksandrhovhannisyan.com/blog/async-functions-that-return-booleans/)).

### Test Plan 

Revised unit test suite to account for removed validation from DAO. Manually tested API for edge and typical cases:

image


henry-li-06 (approved):
LGTM!

PR #270, opened 2022-04-06
Dev Portfolio Automation GitHub API

JacksonStaniec:
### Summary 

This PR is the first iteration of the calls to the GitHub API to get information necessary to validate a dev portfolio submission.
- `getReviewComments` retrieves all review comments under a PR. Relevant information about each comment: 
   - date created
   - comment author
   - text content
- `getReviewedPR` retrieves relevant information about a PR being used for the review requirement:
   - url (for easy reference to PR)
   - PR author
   - review comments (see previous bullet point)
- `getOpenedPR` retrieves relevant information about a PR being used for the open requirement
   - url
   - PR author
   - date PR opened

There is also a template for the validation logic. The validation process is structured using error handling, so that if an error is encountered at any point throughout the process, validation halts and returns as "invalid." 

### Test Plan 

No changes to functionality. 

### Notes 

~We need to add a required field of `IdolMember` to be the member's GitHub username for the validation process.~ Only needs to be required for devs.

How granular do we want to get with feedback for each PR submitted for consideration? For example, if a user submits a list of links for reviewed and opened PRs:
```
reviewedPRs: ["link1", "link2"]
openedPRs: ["link1", "link2"]
```
Then do we want to give feedback:
```
reviewedPRs: valid
     "link1": valid
     "link2": valid
openedPRs: invalid
     "link1": invalid - reason ....
     "link2": invalid - reason ....
```
Or just an overall determination of 'valid' or 'invalid' for the whole submission?


henry-li-06:
Good point about the granularity... I'm thinking that it might be good to provide feedback for each PR that's submitted for greater transparency to the user.

henry-li-06:
I don't think we can make the field required since this doesn't apply to all members on DTI (i.e. designers/business/pms might not have GitHubs). But maybe we can just make a submission invalid if the member's optional GitHub username field is empty?

backend/src/github-util.ts
```diff
@@ -0,0 +1,242 @@
...
+  created_by: string;
+  comments: ReviewComment[];
+};
+
+type OpenedPR = {
+  url: string;
+  created_by: string;
+  created_at: number;
+  diff_size: number;
+};
```
henry-li-06, line 27: use camel case for the fields

backend/src/github-util.ts
```diff
@@ -0,0 +1,242 @@
...
+    // placeholder logic: valid if at least 10 changes total
+    if (open.diff_size < 10) {
+      throw new Error('Trivial PR.');
+    }
+  });
+};
+
+/** ="at least one of `results` is valid" */
+const atLeastOneValid = (results: ValidationResult[]) =>
+  results.reduce((isValid, result) => isValid || result.status === 'valid', false);
```
henry-li-06, line 222: [`some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) might be more suitable here

backend/src/github-util.ts
```diff
@@ -0,0 +1,242 @@
...
+  const end = portfolio.deadline;
+  const username = submission.member.github || ''; // must be github username
+
+  return createValidationResult(async () => {
+    // get open object
+    const open = await getOpenedPR(parseGithubUrl(open_url));
+
+    if (open.created_by !== username) {
+      throw new Error(`User ${username} did not open the pull request ${open.url}.`);
+    }
```
henry-li-06, line 203: I think as long as the member has commits in the PR then it's valid (i.e. they don't necessarily have to be the one opening the PR). I'm not 100% about this tho so let's leave this as-is for now.

backend/src/github-util.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  return octokit.rest.pulls.get(pull_request).then((res) => {
+    // handle non 200 response?
+
+    const pr = res.data;
+    return {
+      url: pr.html_url,
+      created_by: pr.user?.login || '',
+      created_at: Date.parse(pr.created_at),
+      diff_size: pr.additions + pr.deletions
```
henry-li-06, line 132: we might want to consider getting this info from the big diff bot -- i believe it does the calculation in a slightly smarter way that isn't just additions + deletions

henry-li-06:
blanket comment -- use camel case for naming variables

really nice work though! this is a really good start

backend/src/github-util.ts
JacksonStaniec: Ive been using Python too much....

backend/src/github-util.ts
JacksonStaniec: Didn't know that existed!

backend/src/github-util.ts
JacksonStaniec: > we might want to consider getting `diff_size` from the big diff bot -- i believe it does the calculation in a slightly smarter way that isn't just additions + deletions

While implementing this, I realized that Github treats thread comments (like the one im writing now) differently than comments made for a review (like the one im quoting). While the PR requirement is for "reviews," should we take both review comments and non-review comments into consideration? It seems like a technicality if someone's review PR is determined to be invalid just because they made only "regular" comments instead of review comments.

JacksonStaniec:
> we might want to consider getting diff_size from the big diff bot -- i believe it does the calculation in a slightly smarter way that isn't just additions + deletions

While implementing this, I realized that Github treats thread comments (like the one im writing now) differently than comments made for a review (like the one im quoting). While the PR requirement is for "reviews," should we take both review comments and non-review comments into consideration? It seems like a technicality if someone's review PR is determined to be invalid just because they made only "regular" comments instead of review comments.

backend/src/github-util.ts
```diff
@@ -151,17 +183,35 @@ const createValidationResult = async (validation_function): Promise<ValidationRe
   return { status: 'valid' };
 };
 
+/** Parse information from portfolio and submission. Returns tuple of earliest
+ *  valid time, deadline, and Idol member github username.
+ *  Raises an error if the Idol member from `submission` does not have a github username. */
+const parsePortfolioSubmission = (portfolio: DevPortfolio, submission: DevPortfolioSubmission) => {
+  const start = portfolio.earliestValidDate;
+  const end = portfolio.deadline;
+  const username = submission.member.github; // must be github username
```
henry-li-06: the `github` field will be a github URL so we'll probably need to parse the username from the URL

riyajaggi:
Should we also rename the file as `/githubUtil.ts` instead of `/github-util.ts` for consistent case? also, we could add a sub-directory for utils in the backend?

henry-li-06:
> > we might want to consider getting diff_size from the big diff bot -- i believe it does the calculation in a slightly smarter way that isn't just additions + deletions
> 
> While implementing this, I realized that Github treats thread comments (like the one im writing now) differently than comments made for a review (like the one im quoting). While the PR requirement is for "reviews," should we take both review comments and non-review comments into consideration? It seems like a technicality if someone's review PR is determined to be invalid just because they made only "regular" comments instead of review comments.

Hmm that's interesting... it probably makes sense to also consider "non-review" comments as a valid review

backend/src/github-util.ts
```diff
@@ -0,0 +1,283 @@
...
+    ([pr, comments]) => {
+      // get any comments by dti bot
+      const botComments = filterComments(comments, 'dti-github-bot', 0, Number.MAX_SAFE_INTEGER);
+
+      let diffSize = 0;
+      botComments.forEach((comment) => {
+        const match = comment.content.match(/.*\[diff-counting\] Significant lines: ([0-9]+).*/);
+        if (match == null) {
+          throw new Error(`No significant line count by dti-github-bot for PR ${pr.data.url}.`);
+        }
```
henry-li-06, line 170: hmm i think we need to be careful about if the big diff bot hasn't finished calculating the diff yet (i.e. the comment isn't there yet)

henry-li-06: might be fine to leave as it is for now since using diff size to determine if PR is valid is only temporary

JacksonStaniec:
> Should we also rename the file as `/githubUtil.ts` instead of `/github-util.ts` for consistent case? also, we could add a sub-directory for utils in the backend?

I like `githubUtil.ts`, but there does seem to be a lot of varying styles in the backend repo (i.e. `site-integration.ts`, `signinformAPI.ts`, `team-eventsAPI.ts`) that we should change too if we want to make camelcase the one to be consistent with. I think a utils directory is a good idea too

backend/src/util/githubUtil.ts
```diff
@@ -0,0 +1,310 @@
...
+    // works for both review comments and approval comments
+    const toComment = (data): Comment => ({
+      commentUrl: data.html_url,
+      createdBy: data.user.login,
+      createdAt: Date.parse(data.created_at) || Date.parse(data.submitted_at),
+      content: data.body
+    });
+
+    const reviewComments = reviewCommentsRes.data;
+    const approvalComments = approvalCommentsRes.data.filter((comment) => comment.body !== ''); // most are empty
```
JacksonStaniec, line 102: There is yet another different type of comment that requires a separate API call: the comments attached to PR approvals/rejections. (Henry's "LGTM!" comment on [this PR](https://github.com/cornell-dti/idol/pull/268) for example.)

Review comments are classified as a type of approval/rejection that have a status of "COMMENTED" (as opposed to "APPROVED" or "REJECTED"), but they have a blank body for some reason. So, most of the responses from this second API call are useless, so I filter them first.

riyajaggi:
> > Should we also rename the file as `/githubUtil.ts` instead of `/github-util.ts` for consistent case? also, we could add a sub-directory for utils in the backend?
> 
> I like `githubUtil.ts`, but there does seem to be a lot of varying styles in the backend repo (i.e. `site-integration.ts`, `signinformAPI.ts`, `team-eventsAPI.ts`) that we should change too if we want to make camelcase the one to be consistent with. I think a utils directory is a good idea too. On this note we could also maybe use an `api` directory like the `dao` one to keep everything better organized.

Great point! #276 aims to resolve this

henry-li-06 (approved):
Nice work, LGTM!

backend/src/util/githubUtil.ts
JacksonStaniec: The goal is to get every comment made by the user under a PR, and consider them all together as a "review." So, stuff like "LGTM" is considered for the review requirement. However, with PRs like [this](https://github.com/cornell-dti/idol/pull/268), the sole "LGTM" comment doesn't satisfy the minimum length requirement, and so this is a "trivial" review. I think it is good to include these approval comments because most times they are just "LGTM" or something short, so they don't contribute much to a review, but it could also possibly be something constructive about future PRs to make, things to improve, etc.

PR #271, opened 2022-04-11
Archive FA21 Members 

henry-li-06:
### Summary 

This PR archives Fall 2021 members in order to update the website with the current team roster. 

### Test Plan 

N/A

### Notes 
The archived members aren't used for anything at the moment but seemed reasonable to at least archive them.


riyajaggi (changes requested):
I think we would also have to update this file https://github.com/cornell-dti/idol/blob/archive-members/backend/src/members-archive/index.ts to completely archive the fall members

henry-li-06:
> I think we would also have to update this file https://github.com/cornell-dti/idol/blob/archive-members/backend/src/members-archive/index.ts to completely archive the fall members and rename either json file to have consistent naming

You're definitely right that we'd need to update `backend/src/members-archive/index.ts` to properly archive the members but that requires some other annoying changes that aren't really worth it atm. For this PR i just wanted to store the previous list of members in a separate file for now to be able to make those changes in the future.

riyajaggi (approved):
That sounds good, thank you! 

The PR LGTM!!

PR #275, opened 2022-04-12
Fix get member test

henry-li-06:
### Summary 

Fixed get member test by changing the test to compare email for equality instead of the entire object 

The test was failing because the `formerSubteams` field was changed 
image

The test shouldn't fail because non-identifying fields were changed. 

### Test Plan 

Test should pass now


riyajaggi (approved):
LGTM 🚀

PR #280, opened 2022-04-17
[bug-fix] Fix admin shoutouts for former members 

henry-li-06:
### Summary 

This PR fixes the issue on the admin shoutouts page that @riyajaggi discovered. Since shoutouts use `DocumentReference` for the givers and receivers, we get an error for former members (i.e. members that are no longer n the `members` collection). 

The solution implemented in this PR is the replace the name of the giver/receiver with `(Former member)` if the giver/receiver is `undefined.`

image

### Test Plan 
Check deploy preview to ensure that the admin shoutouts page is loading 

### Notes
In the future, we should probably consider implementing a more robust system in the backend for dealing with document references to former members (possibly by looking through archived members).


riyajaggi (approved):
LGTM! Thank you for the fix! 🚀

PR #281, opened 2022-04-17
[Refactor] Search archived members for document references 

henry-li-06:
### Summary 
Currently, any document references for members are directly accessed using `(await docRef.get()).data()`. This causes issues for document references for members who are no longer current members since the member will not be in the `members` collection. This PR adds a utility function for getting member data from a document reference that searches through archived members if the member is not a current member. 

### Test Plan 
Check that all affected pages display data for former members


backend/src/utils/memberUtil.ts
```diff
@@ -13,6 +15,16 @@ export const filterImagesResponse = (
       fileName: image.fileName.slice(image.fileName.indexOf('/') + 1)
     }));
 
+export const getMemberFromDocumentReference = async (
+  docRef: firestore.DocumentReference
+): Promise<IdolMember> => {
+  const snapshot = await docRef.get();
+  if (!snapshot.exists) {
+    return archivedMembersByEmail[docRef.id];
```
airaagrawal1, line 23: Looks like archived members are manually being put in. This means if someone is removed from Idol, they are not automatically archived which could cause future bugs. Should we look into automating archived members?

henry-li-06: Actually I'm not sure if it's possible to do this in any automated way with the current way archiving works. Archived members are just stored in a JSON file and I don't believe it'd be possible to write to this file since we have a serverless backend. 

I also think it might be unnecessary -- we shouldn't be deleting current members for no reason and as long as archiving happens at the end of every semester/beginning of a new semester then this shouldn't be an issue.

riyajaggi (approved):
LGTM!!

PR #267, opened 2022-03-31
Implement TEC review on the admin side

riyajaggi:
### Summary 

This pull request implements the Team Event Credit review on the admin side. The admins can now view the details (member, number of hours and event proof image) of the pending requests and either accept or reject the request for each team event. Accepting the request adds the `TeamEventAttendance` to the `attendees` and rejecting the request, deletes the `TeamEventAttendance` (along with the image).

The admin dashboard was refactored to support the above and to make the layout concise. The TEC review takes place in the details page instead of the main page.

### Test Plan 

- [x] display pending requests and their details
- [x] approve request and move it to attendees list
- [x] reject request and delete it 

### Video

https://user-images.githubusercontent.com/55467524/160969048-b15e87d2-1f2d-437d-a504-2d8eee7aaba9.mov

### Screenshot of the review modal

Screen Shot 2022-03-30 at 10 54 01 PM
_The idol image represents the event proof image_

PR #284, opened 2022-04-21
Determine whether PR or review is trivial for dev portfolio automation

JacksonStaniec:
### Summary 

This PR revises the placeholder logic for determining whether a PR or review is trivial. 
- There is no way that an opened PR can be trivial. This allowed for some unnecessary API calls for validating an opened PR to be removed 
- A review can be trivial if it has less than 10 words. This is as opposed to a minimum character count so that things like [Henry's beautiful artworks](https://github.com/cornell-dti/idol/pull/267#pullrequestreview-929767102) dont count.
- Determining whether a collection of comments is trivial was factored out in to a separate function to allow for easier modification later if course policy changes

### Test Plan 

No changes to functionality.

### DIFF
--- a/backend/src/utils/githubUtil.ts
+++ b/backend/src/utils/githubUtil.ts
@@ -24,7 +24,6 @@ type OpenedPR = {
   url: string;
   createdBy: string;
   createdAt: number;
-  diffSize: number;
 };
 
 // useful to see reason why if not valid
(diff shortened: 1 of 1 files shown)


JacksonStaniec:
While testing, I ran into the API rate limit for the first time, which seems to be 60 requests per hour for unauthenticated calls. The [docs](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting) are a little confusing, but 
 I believe we can up this limit to 1 000 requests per hour if we send a Github token with the request. Can we use the `dti-github-bot` for this?

henry-li-06:
> While testing, I ran into the API rate limit for the first time, which seems to be 60 requests per hour for unauthenticated calls. The [docs](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting) are a little confusing, but I believe we can up this limit to 1 000 requests per hour if we send a Github token with the request. Can we use the `dti-github-bot` for this?

I think I'd have to ask the dev leads for the `dti-github-bot` token since I don't have access to it. Alternatively we could create a new token with the IDOL bot account if need be.

henry-li-06 (approved):
LGTM

backend/src/utils/githubUtil.ts
```diff
@@ -246,16 +228,32 @@ const validateReview = async (
...
 
+/** Returns `comments` with duplicate content from quote replies removed. */
+const removeQuoted = (comments: Comment[]): Comment[] => {
+  // quote replies are of the form "> (original comment)\r\n(new comment)"
+  const quotePattern = />(.*?)\r\n/gs;
+  return comments.map((comment) => ({
+    ...comment,
+    content: comment.content.replace(quotePattern, '').trim()
+  }));
+};
```
JacksonStaniec, line 244: This isolates the comment that was made by the user, so that the content that is being quoted doesnt count towards the requirement.

henry-li-06:
For the record, I think my beautiful artworks should absolutely count XD

PR #288, opened 2022-04-27
Implement BE support and add endpoint for email notifications

riyajaggi:
### Summary 

This PR sets up the infrastructure for sending emails through the backend and adds email notifications for updating member profiles. 

### Test Plan 

No functionality changed. The endpoint can be tested using dummy email once the service is authenticated and `.env` is updated with the required tokens.

### Notes 

New packages:
- nodemailer 
- googleapis


backend/src/API/mailAPI.ts
```diff
@@ -0,0 +1,16 @@
+import transporter from '../nodemailer';
+
+require('dotenv').config();
```
henry-li-06, line 3: I think this is unnecessary since `require('dotenv').config()` already gets called in `firebase.ts`

backend/src/api.ts
```diff
@@ -123,6 +124,12 @@ const loginCheckedDelete = (
   handler: (req: Request, user: IdolMember) => Promise<Record<string, unknown>>
 ) => router.delete(path, loginCheckedHandler(handler));
 
+// Email notifs
+router.post('/sendMail', async (req, res) => {
+  sendMail(req.body.to, req.body.subject, req.body.text);
+  res.status(200);
+});
```
henry-li-06, line 131: I don't think we necessarily need this to be an endpoint. For something that needs an email notification like `/updateMember`, we can probably just call `sendMail()` directly from `updateMember()` in `membersAPI.ts`.

henry-li-06:
Also I'll try to setup the OAuth stuff later today and I'll lyk when that gets done

backend/src/api.ts
riyajaggi: Sounds good! yeah, I was initially going to draft the email and call this in front end, but I see how it is just easier to do it all in backend and avoids unnecessary calls :)

henry-li-06: Actually it might not be a bad idea to leave this in there for now just for testing purposes... sorry lol

riyajaggi: np, just reverted this

riyajaggi:
Ques: Should I add any error handling for emails (ie. 500)? Also, I didn't put in any permission check since the email is on behalf of idol, but wanted to confirm if there's any specific permissions?

henry-li-06:
> Ques: Should I add any error handling for emails (ie. 500)? Also, I didn't put in any permission check since the email is on behalf of idol, but wanted to confirm if there's any specific permissions?

I think it might be ok to not do any error handling -- it's not really the end of the world if the email notif failed to send. Also probably no need for permissions, if these are only for notifications then whatever endpoint handler is calling this API will already check permissions.

backend/src/dao/AdminsDao.ts
```diff
@@ -0,0 +1,8 @@
+import { adminCollection } from '../firebase';
+
+export default class Adminsdao {
```
JacksonStaniec, line 3: Camel case for consistency
```suggestion
export default class AdminsDao {
```

backend/src/api.ts
```diff
@@ -261,6 +280,9 @@ loginCheckedPost('/updateCandidateDeciderRating', (req, user) =>
 loginCheckedPost('/updateCandidateDeciderComment', (req, user) =>
   updateCandidateDeciderComment(user, req.body.uuid, req.body.id, req.body.comment).then(() => ({}))
 );
+loginCheckedPost('/sendMail', async (req, user) => ({
+  info: await sendMail(req.body.to, req.body.subject, req.body.text)
+}));
```
JacksonStaniec, line 285: Should we add a comment that this just for testing at the moment?

backend/src/API/mailAPI.ts
```diff
@@ -0,0 +1,52 @@
+import axios from 'axios';
+import { Request } from 'express';
+import getEmailTransporter from '../nodemailer';
+import { isProd } from '../api';
+import AdminsDao from '../dao/AdminsDao';
+
+export const sendMail = async (to: string, subject: string, text: string): Promise<unknown> => {
+  // Don't send email notifications locally
+  if (!process.env.isProd) {
+    return {};
```
JacksonStaniec, line 10: Should there be some feedback still if its local to let you know its working?

JacksonStaniec (commented):
In general there are a ton of console logs and commented out portions of code, especially in `api.ts`.

backend/src/API/mailAPI.ts
henry-li-06: Good point, I'll add that.

backend/src/api.ts
JacksonStaniec: Alright thanks for explaining, I got confused reading through the conversations before when you guys were talking about leaving a testing endpoint in.

henry-li-06: yeah no problem -- i initially didn't think that we needed it but in implementing it later on (after the semester ended) this was the only way that I could get it to work

JacksonStaniec (approved):
Sorry for the wait!
```
            __/___            
          _____/______|           
  _______/_____\_______\_____     
  \              < < <       |    
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #293, opened 2022-05-02
Dev DB update workflow 

henry-li-06:
### Summary 

This PR adds a new GitHub Actions workflow that executes a script to update the dev Firebase instance. The script essentially overwrites the entire dev db with the data from the prod db. The script runs every month on the first day of the month at 3-4 AM (depending on daylight savings) or can be executed manually. 

### Test Plan 
Run the script locally

### DIFF
--- a/backend/scripts/update-dev-db.ts
+++ b/backend/scripts/update-dev-db.ts
@@ -1,10 +1,9 @@
 import admin from 'firebase-admin';
-import { configureAccount } from '../src/firebase';
+
+require('dotenv').config();
 
 const prodServiceAccount = require('../resources/idol-b6c68-firebase-adminsdk-h4e6t-40e4bd5536.json');
 const devServiceAccount = require('../resources/cornelldti-idol-firebase-adminsdk-ifi28-9aaca97159.json');
-
-require('dotenv').config();
 
 type DbData = {
   id: string;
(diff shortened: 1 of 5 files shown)


riyajaggi (commented):
There are a couple of GitHub actions check warnings on the PR. It might be cleaner to resolve them before merging this PR.

image

henry-li-06:
> There are a couple of GitHub actions check warnings on the PR. It might be cleaner to resolve them before merging this PR.
> 
> image

@riyajaggi fixed! thanks for catching this

backend/scripts/update-dev-db.ts
```diff
@@ -0,0 +1,123 @@
...
+      credential: admin.credential.cert(configureAccount(prodServiceAccount, true)),
+      databaseURL: 'https://idol-b6c68.firebaseio.com',
+      storageBucket: 'gs://cornelldti-idol.appspot.com'
+    },
+    'prod'
+  );
+  const devApp = admin.initializeApp(
+    {
+      credential: admin.credential.cert(configureAccount(devServiceAccount, false)),
+      databaseURL: 'https://idol-b6c68.firebaseio.com',
```
JacksonStaniec, line 111: Is this url supposed to be different from the one for the prod db? Looking at the firebase console, the dev db has a different url, but it seems to be a weird default one.

henry-li-06: Tbh I'm not even 100% if it needs to be included. I think this db URL is referring to firebase's realtime database which is different than the firestore database. We don't even ever use this feature of firebase.

JacksonStaniec (approved):
No more outdated docs messing up testing!

PR #301, opened 2022-05-10
Clear fields of admin forms after each submission

dan-ieljin:
### Summary 

On the admin side, there are forms to make a sign-in link and to create a candidate decider instance. Previously, after the form is submitted, the input in each field remains there, so the user would have to manually delete the input to submit new data. This PR clears the form fields by simply clearing the different states upon clicking the submit button.

### Test Plan 

1a
image

1b
image

2a
image

2b
image

### DIFF
--- a/frontend/src/components/Admin/AdminCandidateDecider/AdminCandidateDecider.tsx
+++ b/frontend/src/components/Admin/AdminCandidateDecider/AdminCandidateDecider.tsx
@@ -87,13 +87,15 @@ const CandidateDeciderInstanceCreator = ({
     };
     CandidateDeciderAPI.createNewInstance(instance)
       .then((newInstance) => setInstances((instances) => [...instances, newInstance]))
-      .then(() => setSuccess(true));
-    setName('');
-    setHeaders([]);
-    setResponses([[]]);
-    setAuthorizedMembers([]);
-    setAuthorizedRoles([]);
-    setFileInKey(Date.now().toString());
+      .then(() => setSuccess(true))
+      .then(() => {
+        setName('');
+        setHeaders([]);
+        setResponses([[]]);
+        setAuthorizedMembers([]);
+        setAuthorizedRoles([]);
+        setFileInKey(Date.now().toString());
+      });
   };
 
   return (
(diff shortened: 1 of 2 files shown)


frontend/src/components/Admin/AdminCandidateDecider/AdminCandidateDecider.tsx
```diff
@@ -87,6 +88,12 @@ const CandidateDeciderInstanceCreator = ({
     CandidateDeciderAPI.createNewInstance(instance)
       .then((newInstance) => setInstances((instances) => [...instances, newInstance]))
       .then(() => setSuccess(true));
+    setName('');
+    setHeaders([]);
+    setResponses([[]]);
+    setAuthorizedMembers([]);
+    setAuthorizedRoles([]);
```
henry-li-06, line 95: These should probably be in the .then() with setSuccess(true)

frontend/src/components/Admin/AdminCandidateDecider/AdminCandidateDecider.tsx
```diff
@@ -87,6 +88,12 @@ const CandidateDeciderInstanceCreator = ({
...
+    setFileInKey(Date.now().toString());
```
henry-li-06, line 96: What is this for?

henry-li-06:
Thanks for providing before and after screenshots!

frontend/src/components/Admin/AdminCandidateDecider/AdminCandidateDecider.tsx
dan-ieljin: Oh, not sure how some code got deleted - just added it back. The purpose of ```setFileInKey``` is to change the ```key``` of the file input component so that after submission, it is forced to re-render. Does this make sense?

henry-li-06: Ohhh i see i see

frontend/src/components/Admin/AdminCandidateDecider/AdminCandidateDecider.tsx
henry-li-06: +1

henry-li-06: As in 
```
...
.then(() => {
    setSuccess(true);
    setName('');
    setHeaders([]);
    ...
})
````

JacksonStaniec (approved):
LGTM!

PR #302, opened 2022-05-11
Update dev portfolio submission type to allow result for each PR submitted 

JacksonStaniec:
### Summary 

This PR changes the type for dev portfolio submissions so that each PR submitted can receive individual feedback on whether it was valid or invalid.

- Combine PR url and validation result into one type `PullRequestSubmission` 
- Add date validation helper function 
- Added function to create fake PR links for testing

### Test Plan 

Existing tests for dev portfolios were updated to use the new type and the generated fake PRs.

### DIFF
--- a/backend/src/utils/githubUtil.ts
+++ b/backend/src/utils/githubUtil.ts
@@ -25,6 +25,11 @@ type OpenedPR = {
   createdAt: number;
 };
 
+type ValidationResult = {
+  status: 'valid' | 'invalid' | 'pending';
+  reason?: string;
+};
+
 /** Parses GitHub PR `url` for information necessary to make API calls.
  *  Raises an error if URL is malformed. */
 const parseGithubUrl = (url: string): PullRequest => {
@@ -281,14 +286,19 @@ export const validateSubmission = async (
   submission: DevPortfolioSubmission
 ): Promise<DevPortfolioSubmission> => {
   const reviewedResults = await Promise.all(
-    submission.reviewedPRs.map(async (url) => validateReview(portfolio, submission, url))
+    submission.reviewedPRs.map(async (pr) => ({
+      ...pr,
+      ...(await validateReview(portfolio, submission, pr.url))
+    }))
   );
 
   const openedResults = await Promise.all(
-    submission.openedPRs.map(async (url) => validateOpen(portfolio, submission, url))
+    submission.openedPRs.map(async (pr) => ({
+      ...pr,
+      ...(await validateOpen(portfolio, submission, pr.url))
+    }))
   );
-
-  return { ...submission, openedResults, reviewedResults };
+  return { ...submission, openedPRs: openedResults, reviewedPRs: reviewedResults };
 };
 
 export default validateSubmission;
--- a/backend/src/utils/githubUtil.ts
+++ b/backend/src/utils/githubUtil.ts
@@ -1,5 +1,5 @@
 import { Octokit } from '@octokit/rest';
-import { DevPortfolio, DevPortfolioSubmission } from '../types/DataTypes';
+import { DevPortfolio, DevPortfolioSubmission, ValidationResult } from '../types/DataTypes';
 
 type PullRequest = {
   owner: string;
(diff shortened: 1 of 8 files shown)


backend/src/utils/githubUtil.ts
```diff
@@ -127,6 +120,10 @@ const getNonReviewComments = async (pull: PullRequest): Promise<Comment[]> => {
     });
 };
 
+/** ="`time` is between the date range `start` to `end`." */
+export const isWithinDates = (time: number, start: number, end: number): boolean =>
+  start <= time && time <= end;
+
```
henry-li-06, line 126: I think this can go in a more general util file, this isn't really that specific to GitHub

backend/src/utils/githubUtil.ts
```diff
@@ -271,33 +268,31 @@ const validateOpen = async (
...
+    if (!isWithinDates(open.createdAt, start, end)) {
       const startDate = new Date(start).toDateString();
       const endDate = new Date(end).toDateString();
       throw new Error(`Pull request ${open.url} was not made between ${startDate} and ${endDate}.`);
     }
   });
 
 /** ="at least one of `results` is valid" */
-const atLeastOneValid = (results: ValidationResult[]) =>
+export const atLeastOneValid = (results: ValidationResult[]): boolean =>
```
henry-li-06, line 279: I believe this doesn't get called anywhere anymore (?)

common-types/index.d.ts
```diff
@@ -144,5 +149,6 @@ interface DevPortfolioSubmission {
   member: IdolMember;
   openedPRs: string[];
   reviewedPRs: string[];
-  status: 'valid' | 'invalid' | 'pending';
+  openedResults: ValidationResult[];
+  reviewedResults: ValidationResult[];
```
henry-li-06, line 153: Maybe `reviewedPRs` and `openedPRs` can be defined by a list of another custom type instead of adding extra fields to `DevPortfolioSubmission`. Something like:
```
interface PullRequestSubmission {
    status: 'valid' | 'invalid' | 'pending',
    link: string,
    reason?: string
}
```
What are your thoughts on this?

backend/src/utils/githubUtil.ts
JacksonStaniec: Looking in `memberUtil.ts`, the only function there that is technically not specifically related to members is `filterImagesResponse()`. Should we make a `util.ts` file for these two?

backend/src/utils/githubUtil.ts
JacksonStaniec: Yea, I was thinking they could call that on the `openResults` or `reviewedResults` array of results to get the "overall result" easily, but since those fields dont exist anymore, this won't be directly useful.

common-types/index.d.ts
JacksonStaniec: It was kinda awkward to separate the results from the PR it corresponds to, so I like this better. It also gives the chance to actually use the `'pending'` status.

backend/src/utils/githubUtil.ts
henry-li-06: sure sure, a general `util.ts` file will always be useful

henry-li-06 (approved):
```
 /\/\/\                            /  \
| \  / |                         /      \
|  \/  |                       /          \
|  /\  |----------------------|     /\     |
| /  \ |                      |    /  \    |
|/    \|                      |   /    \   |
|\    /|                      |  | (  ) |  |
| \  / |                      |  | (  ) |  |
|  \/  |                 /\   |  |      |  |   /\
|  /\  |                /  \  |  |      |  |  /  \
| /  \ |               |----| |  |      |  | |----|
|/    \|---------------|    | | /|   .  |\ | |    |
|\    /|               |    | /  |   .  |  \ |    |
| \  / |               |    /    |   .  |    \    |
|  \/  |               |  /      |   .  |      \  |
|  /\  |---------------|/        |   .  |        \|
| /  \ |              /   NASA   |   .  |  NASA    \
|/    \|              (          |      |           )
|/\/\/\|               |    | |--|      |--| |    |
------------------------/  \-----/  \/  \-----/  \--------
                        \\//     \\//\\//     \\//
                         \/       \/  \/       \/
```

PR #309, opened 2022-06-27
Implement Footer 

henry-li-06:
### Summary 

Implement footer 

### Figma File 

[Nav bar and Footer](https://www.figma.com/file/6hSXHwtXblzSaR6rFV2b8p/SUMMER22-Website-Dev-Handoff?node-id=5%3A224)

### Test Plan 

- Check that implementation is consistent with design
- Check that links link to the correct URL 
### Notes 
**Unrelated changes**
- Added `dti-website-redesign` as a yarn workspace
- Added `/// <reference types="next/types/global" />` to `dti-website-redesign/next-env.d.ts` to get rid of errors for `.module.css` imports


dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const icons: Icon[] = [
+  {
+    src: '/static/icons/app-store.png',
+    link: 'https://www.example.com',
+    alt: 'app store icon'
+  },
+  {
+    src: '/static/icons/email.png',
+    link: 'emailto: hello@cornelldti.org',
```
JacksonStaniec, line 17: I think this should be `mailto:hello@cornelldti.org`. Currently theres this error:
image
and clicking the icon tries to take you to another page. 

This "mailto" thing also opens some default email app (e.g. not gmail) and picks a random email address to send it from. Maybe this is just on my end, but do you think it would be a better user experience if it just copied the email address to your clipboard or something?

JacksonStaniec:
Also there should be a linear gradient on the footer according to the figma. 
```
background: linear-gradient(180deg, rgba(237, 237, 237, 0) 0%, #CECECE 100%);
```

dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,55 @@
+import styles from './Footer.module.css';
+
+type Icon = {
+  link: string;
+  src: string;
+  alt: string;
+};
+
+const icons: Icon[] = [
```
JacksonStaniec, line 9: Idk if this is too important, but the icons are in a different order from the design.

JacksonStaniec:
Since they're icons we can also download them as svg's instead of png's and they look a bit sharper. They'll also scale better when we make this site responsive.
image
(svg vs. png)

dti-website-redesign/src/components/Footer.tsx
henry-li-06: Hmm not sure... I feel like that's the typical behavior of clicking on an email icon. I think if we did do something like copy it to the clipboard there'd need to be some kind of feedback to the user that the action happened which might detract from the website. We can always see what Tracy thinks about it.

JacksonStaniec (approved):

PR #314, opened 2022-07-21
Email copy notification

andrew032011:
### Summary 

This pull request makes clicking the email icon copy "hello@cornelldti.org" onto the clipboard and displays a notification component as feedback.

### Test Plan 

-Manual testing (video attached)

https://user-images.githubusercontent.com/59291082/180226998-d5748899-813f-4080-9126-3cf04d67b151.mov

### Notes 

I feel there's gotta be a way to change the CSS, under the "@keyframes fadeAndBounce" animation, so that there isn't so much code duplication with translating the notification by -105px, -50px every time. I tried a few things already but to no avail. Suggestions appreciated.


dti-website-redesign/src/components/Footer.tsx
```diff
@@ -1,3 +1,6 @@
+import { useState } from 'react';
+import OverlayTrigger from 'react-bootstrap/OverlayTrigger';
+import Tooltip from 'react-bootstrap/Tooltip';
```
JacksonStaniec, line 3: Do we need to use a `Tooltip` component as opposed to any other div for the overlay? I think that component is used for text tooltips, and we are creating our own "tooltip" with the notification svg.

dti-website-redesign/src/components/Footer.tsx
```diff
@@ -29,10 +27,43 @@ const icons: Icon[] = [
...
+        className={styles.socialIcon}
+        src={'/static/icons/email.svg'}
+        alt={'email icon'}
+        onClick={() => {
+          navigator.clipboard.writeText('hello@cornelldti.org');
+          setIsNotificationVisible(true);
+        }}
+      />
+    </OverlayTrigger>
+  );
```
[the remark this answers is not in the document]
andrew032011: Huh, I get "TypeError: Cannot read property 'useRef' of null" from using `userRef(null)`. Suggestions?

andrew032011:
@JacksonStaniec Made the changes, thanks for the comments! One issue I have now though is this. Not entirely sure how to get rid of this glitch. Was wondering if you had any suggestions.

https://user-images.githubusercontent.com/59291082/181692155-d45d5c4d-66ce-4890-b6b7-e611c3b95244.mov

JacksonStaniec:
> @JacksonStaniec Made the changes, thanks for the comments! One issue I have now though is this. Not entirely sure how to get rid of this glitch. Was wondering if you had any suggestions. (Watch till the last second)
> 
>  Screen.Recording.2022-07-29.at.1.54.27.AM.mov

I believe whats happening is that once the animation is finished, the default position of the notification is being rendered before the state changes from the `onAnimationEnd` function can take effect. That's why it flashes for just a split second before disappearing. You can add `animation-fill-mode: forwards;` after `animation: fadeAndBounce 4s 1;` in the css so that it remains in its final keyframe of the animation instead of "resetting" afterwards.

dti-website-redesign/src/components/Footer.tsx
```diff
@@ -29,10 +26,40 @@ const icons: Icon[] = [
...
+        src={'/static/icons/email.svg'}
+        alt={'email icon'}
+        ref={target}
+        onClick={() => {
+          navigator.clipboard.writeText('hello@cornelldti.org');
+          setIsNotificationVisible(true);
+        }}
+      />
+      <Overlay placement={'top'} show={isNotificationVisible} target={target.current}>
+        <div style={{ position: 'absolute' }} id={`tooltip-top`}>
```
JacksonStaniec, line 45: I dont think we need this id anymore
```suggestion
        <div style={{ position: 'absolute' }}>
```

JacksonStaniec:
To answer your question from the PR, there is a way to get rid of all the `translate` functions in the css. The benefit of using the `Overlay` component is that it takes care of the positioning for you, since we just pass it the prop `placement={'top'}`. So, i think you can just remove all the translating, and remove `position: absolute` from `.emailCopyNotification` so that the `Overlay` prop can position it for you.

dti-website-redesign/src/components/Footer.module.css
```diff
@@ -24,4 +24,38 @@
 .socialIcon {
   height: 35px;
   width: 35px;
+  cursor: pointer;
+}
+
+.emailCopyNotification {
+  animation: fadeAndBounce 4s 1;
```
JacksonStaniec, line 31: I'm sorry, i was wrong before. When i removed the translate functions locally it still worked, but I see in the deploy preview that its wrong the first time you click the icon, but right every time you click afterwards?? Im not sure why that is, but I think its better to go back to manually adjusting the notification position so that it works every time. You can still reduce duplication tho if you adjust the position of the notification itself instead of during the animation:
```suggestion
  position: absolute;
  left: -105px;
  top: -50px;
  animation: fadeAndBounce 4s 1;
```

andrew032011: No worries! Yeah I saw something earlier which is why I think I did that position absolute thing...idk if it helps but I found initially that when the notification is in the middle, and you resize the window, the notification actually snaps into place. Maybe it has to do with its position being "flex". Anyways, made the change.

JacksonStaniec (approved):
Great work Andrew!

PR #319, opened 2022-09-13
add name

andrew032011:
### Summary 

Add name to readme, create Fall 2022

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ All the common types used by the packages are defined [here](./common-types/inde
 
 ## Contributors
 
+### Fall 2022
+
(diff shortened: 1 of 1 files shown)


henry-li-06:
Can you just add everyone?

andrew032011:
> 

Sure

andrew032011:
Do we keep the tradition of leaving Devs' names only or should we add design + PM?

henry-li-06:
> Do we keep the tradition of leaving Devs' names only or should we add design + PM?

It seems like other repos also include PM + design so we should probably do that as well

henry-li-06 (approved):

PR #328, opened 2022-09-23
Andrewchen devportfolio frontend backend integration

andrew032011:
### Summary 

Implemented frontend-backend integration of Dev Portfolio on both the member side and admin side. Users can now submit dev portfolio submissions, admins can create and delete dev portfolios. Admins can also view current dev portfolio submissions.

### Test Plan

Manual testing, video attached

### Notes 
N/A

https://user-images.githubusercontent.com/59291082/191875922-f0544295-db59-4d92-a8f4-771881087431.mov

### DIFF
--- a/frontend/src/API/DevPortfolioAPI.ts
+++ b/frontend/src/API/DevPortfolioAPI.ts
@@ -2,7 +2,6 @@ import APIWrapper from './APIWrapper';
 import { backendURL } from '../environment';
 
 type DevPortfolioSubmissionResponseObj = {
-  submission?: DevPortfolioSubmission;
   error?: string;
 };
 
--- a/frontend/src/API/DevPortfolioAPI.ts
+++ b/frontend/src/API/DevPortfolioAPI.ts
@@ -23,7 +23,7 @@ export default class DevPortfolioAPI {
     devPortfolioSubmission: DevPortfolioSubmission
   ): Promise<DevPortfolioSubmissionResponseObj> {
     return APIWrapper.post(`${backendURL}/makeDevPortfolioSubmission`, {
-      uuid: uuid,
+      uuid,
       submission: devPortfolioSubmission
     }).then((res) => res.data);
   }
--- a/frontend/src/API/DevPortfolioAPI.ts
+++ b/frontend/src/API/DevPortfolioAPI.ts
@@ -1,11 +1,6 @@
 import APIWrapper from './APIWrapper';
 import { backendURL } from '../environment';
 
-type DevPortfolioResponseObj = {
-  devPortfolio: DevPortfolio;
-  error?: string;
-};
-
 type DevPortfolioSubmissionResponseObj = {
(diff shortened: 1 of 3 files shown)


frontend/src/API/DevPortfolioAPI.ts
```diff
@@ -1,8 +1,8 @@
 import APIWrapper from './APIWrapper';
 import { backendURL } from '../environment';
 
-type DevPortfolioResponseObj = {
-  devPortfolio: DevPortfolio;
+type DevPortfolioSubmissionResponseObj = {
+  submission?: DevPortfolioSubmission;
   error?: string;
 };
```
henry-li-06, line 7: I don't think this is needed

andrew032011: Originally was trying to use that for error-handling, if there's an error display it, else use the submission. Is there a better way?

henry-li-06: I could be wrong but I think [this](https://github.com/cornell-dti/idol/blob/main/frontend/src/API/APIWrapper.ts#L47-L57) should already deal with error handling.

frontend/src/components/Forms/DevPortfolioForm/DevPortfolioForm.tsx
```diff
@@ -12,58 +12,34 @@ const DevPortfolioForm: React.FC = () => {
...
-        Emitters.generalSuccess.emit({
-          headerMsg: 'Dev Portfolio Assignment submitted!',
-          contentMsg: `The leads were notified of your submission and your submission will be graded soon!`
-        });
+    DevPortfolioAPI.makeDevPortfolioSubmission(devPortfolio.uuid, devPortfolioRequest).then(
+      (val) => {
+        if (val.error) {
+          Emitters.generalError.emit({
+            headerMsg: "Couldn't submit dev assignment!",
+            contentMsg: val.error
```
andrew032011, line 33: @henry-li-06 Does that mean all of this should go too?

henry-li-06: I think it should be able to stay. I believe `responseMiddleware` just returns the response as long as it's not a 440 status code. And in the case of an error, the backend's response is always something like `{"error": <some-error-message>}`. 

You can see the error handling on the backend [here](https://github.com/cornell-dti/idol/blob/main/backend/src/api.ts#L104-L115). 

You can see [here](https://github.com/cornell-dti/idol/blob/main/frontend/src/components/Forms/UserProfile/UserProfile.tsx#L30-L42) that there's similar error handling on the frontend to what's in the dev portfolio form.

andrew032011:
Resolved offline. We will keep the type to display the error if there is one, but no need to collect the submission in the response right now, so I've removed that field.

henry-li-06 (approved):
LGTM

PR #329, opened 2022-09-25
Change meta tag theme-color to white

henry-li-06:
### Summary 
Changed `<meta>` tag for `theme-color` to be white (#FFFFFF) for the frontend. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

### Notion/Figma Link 

[Notion link](https://www.notion.so/cornelldti/Frontend-Make-the-theme-color-white-5c8939bd0a8b49af8ee8c45f36629e30)

### Test Plan 

On a mobile device, the top bar of screen around the notch should be white.


JacksonStaniec:
I still get the black bars testing this on my phone. I see in the deploy preview that the header still has `content=#000000` for some reason. Do you think this could be something weird going on in Netlify?
image

henry-li-06:
> I still get the black bars testing this on my phone. I see in the deploy preview that the header still has `content=#000000` for some reason. Do you think this could be something weird going on in Netlify? image

That's a possibility... another possibility is that it has something to do with next.js

I tried updating the `<meta>` tag in `_app.tsx` to see if that's the issue. We'll see once the preview finishes building.

andrew032011:
Looks like that _was_ the issue. Deploy preview looks correct now after that last commit.

Screen Shot 2022-09-26 at 12 27 45 AM

JacksonStaniec:
Maybe we can remove the `<head/>` tag in `public\index.html` if `_app.tsx` overrides it anyways. None of our other Netlify projects have that `index.html` file so maybe it's a remnant of migrating to Next.js?

henry-li-06:
> Maybe we can remove the `<head/>` tag in `public\index.html` if `_app.tsx` overrides it anyways. None of our other Next projects have that `index.html` file so maybe it's a remnant of migrating to Next.js?

I think you're right -- it seems like the `index.html` is from when the frontend was build from CRA

JacksonStaniec (approved):
Looks good!

PR #331, opened 2022-09-27
Add Date Validation to Dev Portfolio Submissions

airaagrawal1:
### Summary 

- Added due date to dev portfolio assignment so members can easily identify which is the current assignment.
Screen Shot 2022-09-27 at 5 14 59 PM

- Checked if current submission date is past the selected dev portfolio assignment due date and improved the UI error message to the member. 
Screen Shot 2022-09-27 at 5 11 19 PM

- Checked if current submission date is before the selected dev portfolio assignment starts accepting submissions and improved the UI error message to the member. 
Screen Shot 2022-09-27 at 5 13 24 PM


JacksonStaniec (approved):
Great work! I think just including the due date is good, because the beginning of a dev portfolio's submission window should be the day after the previous dev portfolio's deadline, so users can figure out the submission window for any portfolio just by looking at the due dates.

PR #335, opened 2022-09-29
Increase GitHub API Rate Limit

JacksonStaniec:
### Summary 

This PR uses a GitHub access token to authenticate API calls during made during dev portfolio validation, increasing the rate limit from 60 calls per hour to 5000 calls per hour. 

We can estimate that each dev portfolio submission has 2 opened PR links and 2 reviewed PR links (even though only 1 of each is required by the assignment). It costs 1 API call to validate an opened PR, and 4 API calls to validate a reviewed PR. So, this new rate limit allows the dev portfolio to validate up to 500 submissions per hour. 

### Notion/Figma Link 

[Notion](https://www.notion.so/cornelldti/Idol-FA22-5a158866a8a1472c98b1f3c042b480f8?p=2e4df035eb8d41c98fde53062e110baa&pm=s)

### Test Plan 

Check the API rate limit by running:
```typescript
octokit.rest.rateLimit.get().then((res) => console.log(res.data.rate));
```
with `octokit` initialized using the token stored as an environment variable. (You can do this at the end of `githubUtil.ts` for example). You should see that:
```typescript
limit: 5000
```

Then, make a dev portfolio submission.

Checking the API limit again, the `limit` should still be 5000, and the `used` field should have incremented by however many API calls it took to validate the submission.

### DIFF
--- a/backend/src/utils/githubUtil.ts
+++ b/backend/src/utils/githubUtil.ts
@@ -1,5 +1,11 @@
 import { Octokit } from '@octokit/rest';
 
+require('dotenv').config();
+// authenticated with PAT so rate limit increased
+const octokit = new Octokit({
+  auth: process.env.GITHUB_VALIDATION_TOKEN
+});
+
 type PullRequest = {
   owner: string;
   repo: string;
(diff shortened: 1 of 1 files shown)


JacksonStaniec:
Reopened because I mistyped during a rebase and made a mess.

JacksonStaniec:
@henry-li-06 You were right in that the extra `require('dotenv').config()` was redundant. I thought it was necessary because it didn't work without it when locally testing that specific file, but it works when running the entire backend.

henry-li-06 (approved):
LGTM

PR #338, opened 2022-10-02
Members Can View Past DP Submissions

andrew032011:
### Summary 

- Users were previously not able to see their previous submissions. Now, users can see dev portfolios that they’ve submitted to, and a table of their submissions, using analogous components to the admin dashboard.
- Users are not able to see the validation results, just the submissions. We've hidden their results for now. The results are only available from the admin side.
- Added getAllDevPortfoliosNonAdmin and getDevPortfolioNonAdmin routes, which only return information about the dev portfolios that are accessible to a non-admin (aka, deadline, earliest date, and that member's submissions ONLY). Trying to access the admin-only routes as a non-admin still throws a permissions error. Admins can call the nonAdmin route, which returns information only as if they were a non-admin.
- Architecturally, the admin dashboard and the non-admin form submission page now share their analogous components that display dev portfolios and the submission details

### Notion/Figma Link 

[https://www.notion.so/cornelldti/DP-Form-See-Past-Submissions-00067ed77b394c1ca8d552c37fd0df5a](https://www.notion.so/cornelldti/DP-Form-See-Past-Submissions-00067ed77b394c1ca8d552c37fd0df5a)

### Test Plan 

- Manual testing, video attached

### Breaking Changes 
- As you can see in the video, the submit button is below the divider. I haven't been able to figure out how to get it to be above.

https://user-images.githubusercontent.com/59291082/193474819-dc47d77d-ea03-40b5-8b44-eba0719d72e9.mov

### DIFF
--- a/backend/src/API/devPortfolioAPI.ts
+++ b/backend/src/API/devPortfolioAPI.ts
@@ -5,7 +5,7 @@ import { validateSubmission, isWithinDates } from '../utils/githubUtil';
 
 export const getAllDevPortfolios = async (
   user: IdolMember,
-  isAdminRequest
+  isAdminRequest: boolean
 ): Promise<DevPortfolio[]> => {
   if (!isAdminRequest) {
     return DevPortfolioDao.getAllInstances(false, user);
--- a/backend/src/API/devPortfolioAPI.ts
+++ b/backend/src/API/devPortfolioAPI.ts
@@ -3,14 +3,36 @@ import PermissionsManager from '../utils/permissionsManager';
 import { PermissionError, BadRequestError } from '../utils/errors';
 import { validateSubmission, isWithinDates } from '../utils/githubUtil';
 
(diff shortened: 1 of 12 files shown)


JacksonStaniec (approved):
Great work! Thank you for the quick turnaround on this task. By the way, if you want to move the divider below the submit button, you'd probably have to put the `<Divider/>` and ` <DevPortfolioDashboard/>` components after the `<Form/>` component, like its done in the admin side. But, we can address this in a different PR if youd like.

andrew032011:
Thanks! Yep, I tried that and a couple of other things, here's what happens when I move the two components down (screenshot attached, basically it's missing a bottom margin and the divider fills up the entire width). I think it'll just take some more fighting with CSS to get it to look nice. I'll leave it out for now. Definitely down to fix it after launch. 

Screen Shot 2022-10-02 at 10 49 38 PM (2)

PR #343, opened 2022-10-07
Bug fix: add . to regex

andrew032011:
### Summary 

Bug fix

Regex currently doesn't support URLs that contain . (periods).

i.e. https://github.com/cornell-dti/course-reviews-react-2.0/pull/362 did not work.

### Test Plan 

Manually tested and successfully submitted an example URL

### Notes 

We should probably look into adding other symbols eventually too, pushing this one for now because it's urgent.


JacksonStaniec (approved):
Thank you for the quick response to this!

PR #347, opened 2022-10-15
Populate `reason` only if PR is invalid.

JacksonStaniec:
### Summary 

This PR ensures that there is a `reason` associated with a submission result only if it is determined to be `invalid`

### Test Plan 

Only invalid results should have the reason displayed under them.


henry-li-06 (approved):
LGTM

PR #345, opened 2022-10-13
Export Dev Submissions to CSV

airaagrawal1:
### Summary 

This pull request implemented "export to csv" functionality for dev leads to easily grade dev portfolio submissions. It includes the columns `name`, `netid`, `opened_score`, `reviewed_score`, and `total_score`. Submissions are sorted by netid and the csv only includes one submission from each member. 

If a dev portfolio assignment has no submissions, an error message will pop-up. 

Screen Shot 2022-10-13 at 1 01 53 PM

### Test Plan 

- Manually tested and checked that all submissions were present in CSV with correct values.
- All submissions were correctly sorted by netID.
- Tested dev portfolio with multiple submissions from one member. 
- Tested dev portfolio with no submissions.

PR #348, opened 2022-10-16
Add late deadline feature 

henry-li-06:
### Summary 
Added support for late deadline for dev portfolios. A late deadline is optional when creating a new dev portfolio.

### Notion/Figma Link 

[Notion link](https://www.notion.so/cornelldti/Late-Submissions-for-Dev-Portfolios-30da8b9dbefb4f6298906f4023b6f38c)

### Test Plan 
- Updated backend unit testing 
- Test validation for creating a new portfolio (i.e. late deadline is after the normal deadline, etc.) 
- Test that late submissions are marked as late (and that submission before the deadline isn't marked late)
- Test that late submissions have a `*` by the submission status 
- Test that a dev portfolio should behave normally if the late deadline wasn't set 
- Test that you can't submit a portfolio past the late deadline 
- Test that late deadline is used (only if it was set) as the "end date" for submission validation 

### Notes 
- ~~CSV exporting isn't implemented in this PR. This is currently blocked by #345.~~
- [This issue](https://www.notion.so/cornelldti/DP-Admin-Add-validation-for-creating-new-DP-assignment-66f97764abbc4bed89900ce5b8416e3b) should be addressed in a separate PR still

### DIFF
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -144,10 +144,12 @@ interface DevPortfolio {
   earliestValidDate: number;
   submissions: DevPortfolioSubmission[];
   readonly uuid: string;
+  lateDeadline?: number;
 }
 
 interface DevPortfolioSubmission {
   member: IdolMember;
   openedPRs: PullRequestSubmission[];
   reviewedPRs: PullRequestSubmission[];
+  isLate?: boolean;
 }
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -144,7 +144,7 @@ interface DevPortfolio {
   earliestValidDate: number;
   submissions: DevPortfolioSubmission[];
   readonly uuid: string;
-  lateDeadline?: number;
+  lateDeadline: number | null;
 }
 
 interface DevPortfolioSubmission {
--- a/backend/src/API/devPortfolioAPI.ts
+++ b/backend/src/API/devPortfolioAPI.ts
@@ -48,11 +48,16 @@ export const createNewDevPortfolio = async (
 
   const updatedDeadline = new Date(instance.deadline);
   const updatedEarliestValidDate = new Date(instance.earliestValidDate);
+<<<<<<< HEAD
   const updatedLateDeadline = instance.lateDeadline ? new Date(instance.lateDeadline) : null;
+=======
+  const updatedLateDeadline = instance.lateDeadline ? new Date(instance.lateDeadline) : undefined;
+>>>>>>> 286cb9a (add late deadline support)
   const modifiedInstance = {
     ...instance,
     deadline: updatedDeadline.setHours(23, 59, 59),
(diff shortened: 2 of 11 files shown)


common-types/index.d.ts
```diff
@@ -144,10 +144,12 @@ interface DevPortfolio {
...
   submissions: DevPortfolioSubmission[];
   readonly uuid: string;
+  lateDeadline?: number;
 }
 
 interface DevPortfolioSubmission {
   member: IdolMember;
   openedPRs: PullRequestSubmission[];
   reviewedPRs: PullRequestSubmission[];
+  isLate?: boolean;
```
JacksonStaniec, line 154: Will this run into the same issue as #347 where the db can't handle fields that are `undefined`?

JacksonStaniec (commented):
I think that the submission form validation still needs to be modified to allow submissions after the deadline and to actually populate `isLate` in order to see whether `*` is next to each late submission.

JacksonStaniec:
Also there is an error that prevents creating an assignment without a late deadline. 

> TypeError: Cannot read properties of undefined (reading 'uuid')

We could also not make it "optional," and just default set the late deadline to be the same as the regular deadline.

henry-li-06:
> I think that the submission form validation still needs to be modified to allow submissions after the deadline and to actually populate `isLate` in order to see whether `*` is next to each late submission.

Regarding populating the `isLate` field, it's done on the backend [here](https://github.com/cornell-dti/idol/blob/dp/late-submissions/backend/src/API/devPortfolioAPI.ts#L90-L93).

common-types/index.d.ts
henry-li-06: On the backend, `isLate` is always set [here](https://github.com/cornell-dti/idol/blob/dp/late-submissions/backend/src/API/devPortfolioAPI.ts#L90-L93) so we would never try to set `undefined` into the db.

JacksonStaniec:
Since we put the "late" asterisk next to the status, and the status doesn't show for users currently, there is no feedback for the user on whether their assignment is late. Maybe we could put it next to their name in the user view?
image

henry-li-06:
> Since we put the "late" asterisk next to the status, and the status doesn't show for users currently, there is no feedback for the user on whether their assignment is late. Maybe we could put it next to their name in the user view?
> 
> image
> 
> 

Are planning on showing submission status for portfolio 2?

JacksonStaniec:
> > Since we put the "late" asterisk next to the status, and the status doesn't show for users currently, there is no feedback for the user on whether their assignment is late. Maybe we could put it next to their name in the user view?
> > image
> 
> Are we planning on showing submission status for portfolio 2?

Yes, if the admin override feature is working as expected. In the PR where we switch the view of the users, we can change the asterisk to show like it does for the admin view.

JacksonStaniec (approved):
Great work!

PR #351, opened 2022-10-19
Fixed Anonymous Bug in Shoutouts Page

airaagrawal1:
### Summary 
Fixed bug in Shoutouts page so that shoutouts marked as anonymous show up as "From: Anonymous" in the admin Shoutouts page and the member's Shoutout page. 

### Test Plan 
Manually tested that shoutouts were showing up with correct and matching information in member's view and admin view.

Here is the member view:
Screen Shot 2022-10-19 at 5 50 06 PM

Here is the admin view:
Screen Shot 2022-10-19 at 5 51 33 PM

### DIFF
--- a/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
+++ b/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
@@ -1,13 +1,35 @@
 import React, { useState, useEffect } from 'react';
 import { Item, Card } from 'semantic-ui-react';
+import { Member } from '../../../API/MembersAPI';
 import { ShoutoutsAPI, Shoutout } from '../../../API/ShoutoutsAPI';
 import styles from './AdminShoutouts.module.css';
+
+type Props =
+  | {
+      readonly giver: Member;
+      readonly receiver: Member;
+      readonly message: string;
+      readonly isAnon: false;
+    }
+  | {
+      readonly receiver: Member;
+      readonly message: string;
+      readonly isAnon: true;
+    };
 
 const AdminShoutouts: React.FC = () => {
   const [shoutouts, setShoutouts] = useState<Shoutout[]>([]);
   useEffect(() => {
     ShoutoutsAPI.getAllShoutouts().then((shoutouts) => setShoutouts(shoutouts));
   }, []);
+
+  const fromString = (shoutout: Shoutout): string => {
+    if (!shoutout.isAnon) {
+      const { giver } = shoutout;
+      return `From: ${giver?.firstName} ${giver?.lastName} (${giver.email})`;
+    }
+    return 'From: Anonymous';
+  };
 
   return (
     <div className={styles.shoutoutsContainer}>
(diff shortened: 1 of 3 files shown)


frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -27,9 +27,18 @@ const AdminShoutouts: React.FC = () => {
...
+                  {(() => {
+                    switch (shoutout.isAnon) {
+                      case true:
+                        return 'Anonymous';
+                      case false:
+                        return shoutout.giver
+                          ? `${shoutout.giver.firstName} ${shoutout.giver.lastName}`
+                          : '(Former member)';
+                      default:
+                        return 'Anonymous';
```
andrew032011, line 39: Someone else can come in on this if it was discussed offline already, but I didn't see anything in the notion issue that said we were going to anonymize shoutouts from the admin view. Is that new functionality that we've agreed upon? @JacksonStaniec 

If it is intended...

This can be rewritten shorter with an if-statement since you only have two results, non-anonymous and anonymous.

I think it makes more sense to handle this logic similarly in `ShoutoutCard.tsx` if you changed it here. Take a look at how Anonymous users were previously handled in lines 17-37 in `ShoutoutCard.tsx` and see if you can match them.

andrew032011 (changes requested):

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
JacksonStaniec: Yea this was discussed offline. This is the page that is going to be shown during all-hands, so we need to make sure that people are anonymous if they want to be.

airaagrawal1: @andrew032011 I made the logic a bit shorter. Are you also proposing changing the admin view to a `List` of type `Card` instead of type `Item` to match the type in the member view?

henry-li-06:
The shoutouts type is being defined in two separate places rn [here](https://github.com/cornell-dti/idol/blob/fix-anonymous-shoutouts/backend/src/types/DataTypes.d.ts#L18-L29) and [here](https://github.com/cornell-dti/idol/blob/fix-anonymous-shoutouts/frontend/src/API/ShoutoutsAPI.ts#L6-L11). These should probably be unified.

airaagrawal1:
> The shoutouts type is being defined in two separate places rn [here](https://github.com/cornell-dti/idol/blob/fix-anonymous-shoutouts/backend/src/types/DataTypes.d.ts#L18-L29) and [here](https://github.com/cornell-dti/idol/blob/fix-anonymous-shoutouts/frontend/src/API/ShoutoutsAPI.ts#L6-L11). These should probably be unified.

Yes, I can make a separate PR for updating the types as we also want to change the type of `receiver`. I will keep this PR focused on fixing the 'anonymous' bug.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
andrew032011: I think what you have now is fine. More or less I was referring to what Henry mentioned.

andrew032011 (approved):
Looks good!

PR #352, opened 2022-10-22
Unified Shoutouts type and removed requirement that recipient be IDOL user

airaagrawal1:
### Summary 

- Unified Shoutouts type by removing variant without the `giver` field
- Updated Shoutout `recipient` to be type `string` instead of `IdolMember`
- Removed "Received Shoutouts" section from members Shoutouts page
- Removed Shoutouts with old type from dev Firebase 
- Updated Dao tests for correct type

### Test Plan 

- Manually tested that the Shoutouts form works and collects the correct information
- Ensured that anonymous and non-anymous Shoutout functionality still works
- Verified that correct Shoutout information appears in admin view as well as member view
- Ensured that submitted Shoutouts appear in dev Firebase

### DIFF
--- a/frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
+++ b/frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
@@ -4,9 +4,7 @@ import { Shoutout } from '../../../API/ShoutoutsAPI';
 const ShoutoutCard = (props: Shoutout): JSX.Element => {
   const { giver, receiver, message, isAnon } = props;
 
-  let fromString = 'From: Anonymous';
-  if (!isAnon) {
-    fromString = `From: ${giver?.firstName} ${giver?.lastName} (${giver.email})`;
+  const fromString = isAnon ? 'From: Anonymous' :  `From: ${giver?.firstName} ${giver?.lastName} (${giver.email})` ;
   }
 
   return (
--- a/frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
+++ b/frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
@@ -1,35 +1,17 @@
 import { Card } from 'semantic-ui-react';
-import { Member } from '../../../API/MembersAPI';
+import { Shoutout } from '../../../API/ShoutoutsAPI';
 
-type Props =
-  | {
-      readonly giver: Member;
-      readonly receiver: Member;
-      readonly message: string;
-      readonly isAnon: false;
-    }
-  | {
-      readonly receiver: Member;
-      readonly message: string;
-      readonly isAnon: true;
-    };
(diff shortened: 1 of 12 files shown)


frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
```diff
@@ -1,35 +1,17 @@
...
 
-const ShoutoutCard = (props: Props): JSX.Element => {
-  const { receiver, message } = props;
-
-  let fromString;
-  if ('giver' in props) {
-    const { giver } = props;
+  let fromString = 'From: Anonymous';
+  if (!isAnon) {
     fromString = `From: ${giver?.firstName} ${giver?.lastName} (${giver.email})`;
```
JacksonStaniec, line 9: It's not a big deal, but this could be done with a tertiary operator so that `fromString` can be a const. 
```suggestion
  const fromString = isAnon ? 'From: Anonymous' :  `From: ${giver?.firstName} ${giver?.lastName} (${giver.email})` ;
```

frontend/src/components/Forms/ShoutoutsPage/ShoutoutForm.tsx
```diff
@@ -59,29 +54,15 @@ const ShoutoutForm: React.FC<ShoutoutFormProps> = ({ getGivenShoutouts }) => {
...
-        ) : undefined}
-
+        <Form.Input
+          label="Who is awesome?"
+          name="receiver"
+          value={receiver}
+          control={TextArea}
+          onChange={(event) => setReceiver(event.target.value)}
+          required
+        />
```
JacksonStaniec, line 65: Do you think this could be a simpler input component instead of a text area? It just looks a little strange to have so much space to put a name.
image

airaagrawal1: Yup that makes sense

frontend/src/components/Forms/ShoutoutsPage/ShoutoutsPage.module.css
```diff
@@ -6,18 +6,11 @@
...
-}
-
-.listContainer {
-  padding: 2em;
+.shoutoutListContainer {
   width: 50%;
+  align-self: center;
+  margin: auto;
+  height: calc(100vh - 80px - 25vh);
 }
```
JacksonStaniec, line 14: I think it might be here but the list gets cut off if theres more than 2 shoutouts.
image

airaagrawal1: I was zoomed out so didn't realize. It should be fixed now!

JacksonStaniec (approved):
This is really excellent work! You've done an amazing job of taking charge of the Shoutouts feature, and every idea you've had about it has been very well thought-out. Keep it up!

PR #349, opened 2022-10-17
Add text field for TPM dev portfolio submission.

JacksonStaniec:
### Summary 

This PR is the first step to be able to associate text with a dev portfolio submission. Just for the TPM paragraph requirement:
- [ ] added a text box and prompt area under the PR links on the submission form
- [ ] added button to view text response associated with a submission, if any

### Notion/Figma Link 

[Notion](https://www.notion.so/cornelldti/Idol-FA22-5a158866a8a1472c98b1f3c042b480f8?p=224afb83e4274d219f257c43d2f4786e&pm=s)

### Test Plan 

- text area should not show for non-TPMs
- details view should not change if no submissions have text fields

Text area on submission form:
image

View from admin details:
image

View from user details:
image

Clicking "Show Text"
image

### Notes 

This text field format may be able to be reused for non-TPMs as an area to explain why they were granted an exception from certain PR requirements ([Notion task](https://www.notion.so/cornelldti/DP-Add-Other-Category-to-Allow-Exceptions-1f4c5b41c7cd48c488ac2b22b668f62d)).

PR #354, opened 2022-10-23
Allow admins to manually set status of dev portfolio submission.

JacksonStaniec:
### Summary 

This PR adds the ability for admins to manually set the status of a submission after the automatic validation process. Any submissions with a `text` field associated with it will be set to `pending`, which an admin will manually review to determine whether it is `valid` or `invalid`. 
- (Text feature will be added by this PR) #349 

---
- Added `status` field to each dev portfolio submission.
- In admin details view, added a dropdown menu to display the status of each submission.
- Changing the dropdown menu selection will update the submission in the db immediately to reflect the new status. The menu is disabled while the change is being processed.
- Added new API endpoint `/updateDevPortfolioSubmissions` that replaces an existing portfolio's array of submissions.

### Notion/Figma Link 

[Notion](https://www.notion.so/cornelldti/Idol-FA22-5a158866a8a1472c98b1f3c042b480f8?p=dd7214b6d48d48c783661139ba3c970f&pm=s)

### Test Plan 

https://user-images.githubusercontent.com/81320443/197406266-a8a4c4ab-ec94-4cdc-9734-25312bb4f449.mp4

- See that each submission's `status` matches the previous determination of valid or invalid
- Change a submission's status one or more times, and see that it persists in the db/after a refresh
- No change in the user view of submissions

### Notes 

This PR will probably require a bit more modification to be compatible with the late deadline feature (#348), since the status column now has a function instead of just being text. We could mark the name with an asterisk to denote a late submission instead of the status, or create another column that just tells whether the submission was late or not, or change the way that the status is changed from a dropdown menu to something else, etc.

### DIFF
--- a/backend/src/utils/githubUtil.ts
+++ b/backend/src/utils/githubUtil.ts
@@ -31,7 +31,7 @@ type OpenedPR = {
 };
 
 type ValidationResult = {
-  status: 'valid' | 'invalid' | 'pending';
+  status: SubmissionStatus;
   reason?: string | null;
 };
 
@@ -275,9 +275,7 @@ const validateOpen = async (
   });
 
 /** Determines the overall status of the submission. */
-export const getSubmissionStatus = (
-  submission: DevPortfolioSubmission
-): 'pending' | 'valid' | 'invalid' => {
+export const getSubmissionStatus = (submission: DevPortfolioSubmission): SubmissionStatus => {
   // if text area populated, set to pending
   if (submission.text) {
     return 'pending';
--- a/backend/src/utils/githubUtil.ts
+++ b/backend/src/utils/githubUtil.ts
@@ -274,6 +274,23 @@ const validateOpen = async (
     }
   });
 
+/** Determines the overall status of the submission. */
+export const getSubmissionStatus = (
+  submission: DevPortfolioSubmission
+): 'pending' | 'valid' | 'invalid' => {
+  // if text area populated, set to pending
+  if (submission.text) {
+    return 'pending';
+  }
+
+  // otherwise, valid if at least one valid in each category
+  const atLeastOneValid =
+    submission.openedPRs.some((pr) => pr.status === 'valid') &&
+    submission.reviewedPRs.some((pr) => pr.status === 'valid');
+
+  return atLeastOneValid ? 'valid' : 'invalid';
+};
+
 /** Determines whether submission is valid or invalid. */
 export const validateSubmission = async (
   portfolio: DevPortfolio,
@@ -292,7 +309,14 @@ export const validateSubmission = async (
       ...(await validateOpen(portfolio, submission, pr.url))
     }))
   );
-  return { ...submission, openedPRs: openedResults, reviewedPRs: reviewedResults };
+
+  const updatedSubmission = {
+    ...submission,
+    openedPRs: openedResults,
+    reviewedPRs: reviewedResults
+  };
+
+  return { ...updatedSubmission, status: getSubmissionStatus(updatedSubmission) };
 };
 
 export default validateSubmission;
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -132,9 +132,11 @@ interface CandidateDeciderInfo {
   isOpen: boolean;
 }
 
+type SubmissionStatus = 'valid' | 'pending' | 'invalid';
+
 interface PullRequestSubmission {
   url: string;
-  status: 'valid' | 'invalid' | 'pending';
+  status: SubmissionStatus;
   reason?: string | null;
 }
 
@@ -153,5 +155,5 @@ interface DevPortfolioSubmission {
   reviewedPRs: PullRequestSubmission[];
   isLate?: boolean;
   text?: string;
-  status: 'pending' | 'valid' | 'invalid';
+  status: SubmissionStatus;
 }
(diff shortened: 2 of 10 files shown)


airaagrawal1:
This logic changes how submissions are deemed `valid` and `invalid`. Are we going to modify the exported CSV data to include if an admin manually changes a submission's status?

frontend/src/components/Admin/DevPortfolio/DevPortfolioDetails.tsx
```diff
@@ -54,7 +54,10 @@ const DevPortfolioDetails: React.FC<Props> = ({ uuid, isAdminView }) => {
         netid: submission.member.netid,
         opened_score: open,
         reviewed_score: review,
-        total_score: open + review
+        total_score:
+          (submission.status === 'valid' && 2) || // in case admin gave full points
+          (submission.status === 'invalid' && '0') || // in case admin gave zero points
+          open + review
```
JacksonStaniec, line 60: > This logic changes how submissions are deemed valid and invalid. Are we going to modify the exported CSV data to include if an admin manually changes a submission's status?

I knew I forgot something :)   
Let me know what you think of this solution: A `valid` overall status gives full points. Either they got this by having one valid PR in each category (in which case `open + review` would sum to 2), or an admin overrode the results to give them full points. An `invalid` overall status gives zero points, either by an admin override or by not having at least one valid PR in each category. In the case of partial points (im not sure how common this case is), the csv still has the individual score of each category, and so the admins could manually compute this. Or, the admins could override the status by setting it to `pending` (which admittedly would be somewhat confusing to an end user), which would default to the sum of the `open` and `review` points.

JacksonStaniec: @airaagrawal1

backend/src/utils/githubUtil.ts
```diff
@@ -274,6 +274,23 @@ const validateOpen = async (
...
+    return 'pending';
+  }
+
+  // otherwise, valid if at least one valid in each category
+  const atLeastOneValid =
+    submission.openedPRs.some((pr) => pr.status === 'valid') &&
+    submission.reviewedPRs.some((pr) => pr.status === 'valid');
+
+  return atLeastOneValid ? 'valid' : 'invalid';
+};
```
henry-li-06, line 292: I think this logic is better off in the API layer

common-types/index.d.ts
```diff
@@ -151,4 +151,5 @@ interface DevPortfolioSubmission {
   openedPRs: PullRequestSubmission[];
   reviewedPRs: PullRequestSubmission[];
   text?: string;
+  status: 'pending' | 'valid' | 'invalid';
```
henry-li-06, line 154: Maybe we should create a new type for this since it's referenced in a lot of places 
```
type SubmissionStatus = 'pending' | 'valid' | 'invalid' 
```

henry-li-06 (commented):
I think under the current implementation of admin override, the overridden submissions status would be overwritten if all submissions were regarded. Just wanted to check if that was the desired behavior.

JacksonStaniec:
> I think under the current implementation of admin override, the overridden submissions status would be overwritten if all submissions were regarded. Just wanted to check if that was the desired behavior.

Yes, that is intended. I figured that the admins would like to see the effects of the new validation logic, in case it takes care of an exception they would manually grant otherwise. Also if the automatic part is being regraded due to a change in validation logic or grading rubric, i think it is not too unreasonable to expect that the manual part also should be reexamined along with it.

common-types/index.d.ts
JacksonStaniec: Yea that's a good idea

backend/src/utils/githubUtil.ts
JacksonStaniec: Why? This function is used primarily in this utils file, and its only exported for use in the DAO layer to make older submission types in the db backwards compatible with the new type. Once each submission in the db has the new type, then I can remove the use of it outside of this file.

frontend/src/components/Admin/DevPortfolio/DevPortfolioDetails.tsx
airaagrawal1: Ya the `pending` status is a bit confusing, but I think this logic is good for the purposes of exporting to csv. Do you think we should include a `status` column in the csv for admins to double check `total_score` and `status` in the future? For example, an admin thinks someone's score is 2 but they marked it as pending in Idol and forgot to change it.

JacksonStaniec: Yea in the future the leads may want some way to filter the table view to see if there are any `pending` submissions, but i'll add `status` as a column in the meantime. I'll have to get feedback from them on whether that's something theyre interested in

airaagrawal1: Sounds good

backend/src/utils/githubUtil.ts
henry-li-06: That's fair

henry-li-06:
I think LGTM but maybe @airaagrawal1 should take another took to make sure everything is resolved from her comments

airaagrawal1 (approved):
LGTM!

PR #359, opened 2022-10-29
Hide broken/in-progress pages from users.

JacksonStaniec:
### Summary 

This PR removes access to broken pages or features that are not ready for release.
- Removed "Games" tab
- Add message if user's arent supposed to see candidate decider instances
- Hide shoutouts feature until ready for use

### Test Plan 

Non-admin view
image

image

### DIFF
--- a/frontend/src/components/Candidate-Decider/CandidateDeciderBase.module.css
+++ b/frontend/src/components/Candidate-Decider/CandidateDeciderBase.module.css
@@ -1,3 +1,7 @@
 .instanceGroup {
   margin: 20px;
 }
+
+.emptyMessage {
+  margin: 5%;
+}
--- a/frontend/src/components/Candidate-Decider/CandidateDeciderBase.tsx
+++ b/frontend/src/components/Candidate-Decider/CandidateDeciderBase.tsx
@@ -1,5 +1,6 @@
 import { useState, useEffect } from 'react';
 import { Card } from 'semantic-ui-react';
+import Link from 'next/link';
 import CandidateDeciderAPI from '../../API/CandidateDeciderAPI';
 import styles from './CandidateDeciderBase.module.css';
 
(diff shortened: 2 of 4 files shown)


henry-li-06 (commented):
Candidate decider _should not_ be hidden for non-admins. Members can only see the instances that they have access to which for most non-tpms and admins should be none.

JacksonStaniec:
> Candidate decider _should not_ be hidden for non-admins. Members can only see the instances that they have access to which for most non-tpms and admins should be none.

Alright I didn't know non-admins would be using it at all. Do you think it would be a good idea to add some text saying something like "You don't have to access to any right now" if a member doesn't have access just so theres not a blank screen when they're looking on the page?

henry-li-06:
> > Candidate decider _should not_ be hidden for non-admins. Members can only see the instances that they have access to which for most non-tpms and admins should be none.
> 
> Alright I didn't know non-admins would be using it at all. Do you think it would be a good idea to add some text saying something like "You don't have to access to any right now" if a member doesn't have access just so theres not a blank screen when they're looking on the page?

Well just for example most tpms aren't even admins and would need access to be able to review dev apps. But yeah I think that's probably a great idea.

JacksonStaniec:
> > > Candidate decider _should not_ be hidden for non-admins. Members can only see the instances that they have access to which for most non-tpms and admins should be none.
> > 
> > 
> > Alright I didn't know non-admins would be using it at all. Do you think it would be a good idea to add some text saying something like "You don't have to access to any right now" if a member doesn't have access just so theres not a blank screen when they're looking on the page?
> 
> Well just for example most tpms aren't even admins and would need access to be able to review dev apps. But yeah I think that's probably a great idea.

Yea that makes sense. Let me know if you think the message is okay.

henry-li-06 (approved):
LGTM

PR #361, opened 2022-11-03
Sorting and filtering shoutouts for admins

airaagrawal1:
### Summary 

This PR added a timestamp to each shoutout submission and implemented sorting and filtering shoutouts in the admin view. 

- [x] added timestamp field to Shoutouts type
- [x] updated member and admin view to view date of shoutout submission
- [x] added date range for admins to filter shoutouts

### Test Plan 

- Verified that correct dates were showing up for each shoutout. 
- Ensured shoutouts were correctly filtered and sorted. 
- Tested that the date range feature works as expected.


frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
```diff
@@ -2,11 +2,11 @@ import { Card } from 'semantic-ui-react';
 import { Shoutout } from '../../../API/ShoutoutsAPI';
 
 const ShoutoutCard = (props: Shoutout): JSX.Element => {
-  const { giver, receiver, message, isAnon } = props;
+  const { giver, receiver, message, isAnon, timestamp } = props;
 
-  const fromString = isAnon
-    ? 'From: Anonymous'
-    : `From: ${giver?.firstName} ${giver?.lastName} (${giver.email})`;
+  let fromString = isAnon ? 'From: Anonymous' : `From: ${giver?.firstName} ${giver?.lastName}`;
```
andrew032011, line 7: Feels like the changing this to `let` isn't necessary if we just consolidate these couple lines of template Strings into one. Was the purpose of breaking the code that displays the "from" text up into two lines just for readability?

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -1,41 +1,92 @@
...
   const [shoutouts, setShoutouts] = useState<Shoutout[]>([]);
+  const [displayShoutouts, setDisplayShoutouts] = useState<Shoutout[]>([]);
   useEffect(() => {
-    ShoutoutsAPI.getAllShoutouts().then((shoutouts) => setShoutouts(shoutouts));
+    ShoutoutsAPI.getAllShoutouts().then((shoutouts) => {
+      setShoutouts(shoutouts);
+      setDisplayShoutouts(shoutouts);
+    });
   }, []);
+  const [earlyDate, setEarlyDate] = useState<Date>(new Date('2022/01/01'));
```
andrew032011, line 18: I'm thinking it might make more sense to change this to today's date, since the most common use case is to display the shoutouts at all-hands, which were made the same day. So the "customizability" isn't needed until someone, in some less common scenario wants to display shoutouts for some other date range.

frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
andrew032011: Actually, I think the Date might fit better elsewhere, like the line below the `From: ` text. Combining it with the shoutout author's name feels a little strange.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -1,41 +1,92 @@
...
+  const [lastDate, setLastDate] = useState<Date>(new Date());
 
   const fromString = (shoutout: Shoutout): string => {
+    const dateString = ` (Date: ${new Date(shoutout.timestamp).toDateString()})`;
     if (!shoutout.isAnon) {
       const { giver } = shoutout;
-      return `From: ${giver?.firstName} ${giver?.lastName} (${giver.email})`;
```
andrew032011, line 15: Just want to double-check that we agreed on deleting the email/NetID from the text as part of this change too, since I didn't see a note about that in the PR or the notion page. Feel free to ignore this if you did agree about that.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -1,41 +1,92 @@
...
+    if (lastDate < earlyDate) {
+      Emitters.generalError.emit({
+        headerMsg: 'Invalid Date Range',
+        contentMsg: 'Please make sure the latest shoutout date is after the earliest shoutout date.'
+      });
+    } else {
+      setDisplayShoutouts(
+        shoutouts.filter(
+          (shoutout) =>
+            new Date(shoutout.timestamp) >= earlyDate && new Date(shoutout.timestamp) <= lastDate
```
andrew032011, line 40: Avoiding code duplication/extra instantiating of a Date object, this should probably be best stored in a `const` variable right before the function call and used twice here.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -1,41 +1,92 @@
...
-              </Item.Content>
-            </Item>
-          ))}
-        </Item.Group>
-      )}
+    <div>
+      <Form className={styles.shoutoutForm}>
+        <h2 className={styles.formTitle}>Select date range to display shoutouts:</h2>
+        <Form.Group width="equals">
+          <DatePicker
```
andrew032011, line 51: Might be good to make a component that contains what these two DatePickers have in common, and just takes in the selected date and the setter function. Just to consolidate the two DatePicker's since they're really similar.

andrew032011 (changes requested):

frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
airaagrawal1: I was also debating putting the Date on the line below 'From:', but I was worried that it would make the page longer and be a lot to scroll through during all-hands. Thoughts on this?

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
airaagrawal1: I thought people submit shoutouts leading up to an all hands and during all hands?

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
airaagrawal1: This is connected to not making the shoutouts page too long for admins to scroll through - I exchanged the giver's email for the date, but this can be undone if we want the email. @JacksonStaniec do you know if leads definitely want the giver's email?

frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
andrew032011: Good point. Maybe it could go in the top right corner, horizontally aligned with the `To: ...` text? That seems to be a pretty common place to put it in a lot of online forums.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
andrew032011: Makes sense. I totally agree with getting rid of it by the way. I just wanted to make sure we formally agreed on it.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
andrew032011: In more recent all-hands, I remember submitting shoutouts when the form link was posted, which was usually at the very beginning of all-hands. And my shoutouts usually are the first one, so I assume they aren't opening up shoutouts any earlier than the same day of all-hands.

If they do, they I guess this is fine as is. In that case, probably just want to make the default year dynamic to the current year.

Thoughts @JacksonStaniec?

JacksonStaniec: I agree with @andrew032011 that the date should be set dynamically; it'll just avoid potentially confusing bugs in the future. But @airaagrawal1 has the right idea about what we want to do with the IDOL shoutouts system, where people can submit shoutouts leading up to and during all-hands instead of just during it. The leads are probably going to want to display the shoutouts between the last all-hands and the current one, so maybe the last 2 weeks is a good default range?

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
JacksonStaniec: We're moving to make the giver's actual identity only relevant on the backend, for content moderation purposes. So, their netid doesn't need to be displayed. Just so we're all on the same page, I've asked @airaagrawal1 to feel free to take more creative liberty with the Shoutouts feature because I think she's got good ideas for it, so it's alright if there's deviations from the original ticket. @andrew032011 I do think its a good idea to document these changes somewhere if they are major

andrew032011: Ok, I’m on board with the decision and will keep that in mind for the rest of the review. Thanks for letting me know.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
andrew032011: This sounds right to me.

airaagrawal1: Sounds good 👍

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
airaagrawal1: Sounds good, will make sure to check in next time before making a big change.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
airaagrawal1: @andrew032011 Makes sense. I believe I addressed most of your comments, so please review if you get a chance!

andrew032011: Looks great! Thoughts on moving the date to be inline with the `To: ...` text? So like, top right corner.
 
Screen Shot 2022-11-04 at 1 17 19 PM

airaagrawal1: I agree. I was having issues with formatting before, but I fixed it now!

andrew032011 (approved):
Thanks for making all those changes! Looks really really great :)

PR #365, opened 2022-11-10
Hide shoutouts from admin view

airaagrawal1:
### Summary 

This pull request is the first step towards implementing hiding shoutouts in the admin view. 

- [x] added `hidden` field to Shoutouts type
- [x] added `uuid` field to Shoutouts type
- [x] implemented admin being able to hide a specific shoutout and confirmation modal
- [x] dynamically updated shoutouts list in admin view to hide chosen shoutout 
- [x] change hidden field in backend so shoutout remains hidden for other admins/future use 

### Test Plan 

Manually tested using dev environment. Also added Dao test to test `hideShoutout()` endpoint.

### Notion Link

https://www.notion.so/cornelldti/Idol-FA22-5a158866a8a1472c98b1f3c042b480f8?p=d932c4d29e014f0481afb3016bcdeb39&pm=s

### Future

Next step will be to add 3-part button to toggle between hidden, unhidden, and all shoutouts.

### DIFF
--- a/backend/src/API/shoutoutAPI.ts
+++ b/backend/src/API/shoutoutAPI.ts
@@ -28,21 +28,13 @@ export const getShoutouts = async (
   return ShoutoutsDao.getShoutouts(memberEmail, type);
 };
 
-export const hideShoutout = async (body: string, user: IdolMember): Promise<Shoutout> => {
+export const hideShoutout = async (body: Shoutout, user: IdolMember): Promise<Shoutout> => {
   const canEdit = await PermissionsManager.canHideShoutouts(user);
   if (!canEdit) {
     throw new PermissionError(
       `User with email: ${user.email} does not have permission to hide shoutouts!`
     );
   }
-
-  const shoutout = await ShoutoutsDao.getInstance(body);
-
-  const shoutoutRef = {
-    ...shoutout,
-    hidden: true
-  };
-
-  await ShoutoutsDao.hideShoutout(shoutoutRef);
-  return shoutoutRef;
+  await ShoutoutsDao.hideShoutout(body);
+  return body;
 };
--- a/backend/src/API/shoutoutAPI.ts
+++ b/backend/src/API/shoutoutAPI.ts
@@ -28,13 +28,13 @@ export const getShoutouts = async (
   return ShoutoutsDao.getShoutouts(memberEmail, type);
 };
 
-export const hideShoutout = async (body: Shoutout, user: IdolMember): Promise<Shoutout> => {
+export const updateShoutout = async (body: Shoutout, user: IdolMember): Promise<Shoutout> => {
   const canEdit = await PermissionsManager.canHideShoutouts(user);
   if (!canEdit) {
     throw new PermissionError(
       `User with email: ${user.email} does not have permission to hide shoutouts!`
     );
   }
-  await ShoutoutsDao.hideShoutout(body);
+  await ShoutoutsDao.updateShoutout(body);
   return body;
 };
--- a/backend/src/API/shoutoutAPI.ts
+++ b/backend/src/API/shoutoutAPI.ts
@@ -28,13 +28,21 @@ export const getShoutouts = async (
   return ShoutoutsDao.getShoutouts(memberEmail, type);
 };
 
-export const hideShoutout = async (body: Shoutout, user: IdolMember): Promise<Shoutout> => {
+export const hideShoutout = async (body: string, user: IdolMember): Promise<Shoutout> => {
   const canEdit = await PermissionsManager.canHideShoutouts(user);
   if (!canEdit) {
     throw new PermissionError(
       `User with email: ${user.email} does not have permission to hide shoutouts!`
     );
   }
-  await ShoutoutsDao.hideShoutout(body);
-  return body;
+
+  const shoutout = await ShoutoutsDao.getInstance(body);
+
+  const shoutoutRef = {
+    ...shoutout,
+    hidden: true
+  };
+
+  await ShoutoutsDao.hideShoutout(shoutoutRef);
+  return shoutoutRef;
 };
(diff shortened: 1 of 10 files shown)


backend/src/dao/ShoutoutsDao.ts
```diff
@@ -1,20 +1,21 @@
...
-        })
-      )
+      shoutoutRefs.docs.map(async (doc) => {
+        const dbShoutout = doc.data() as DBShoutout;
+        const giver = await getMemberFromDocumentReference(dbShoutout.giver);
+        return {
+          ...dbShoutout,
+          giver
+        };
+      })
```
henry-li-06, line 18: What's the purpose of this change?

backend/src/API/shoutoutAPI.ts
```diff
@@ -36,3 +27,14 @@ export const getShoutouts = async (
...
+
+export const hideShoutout = async (body: Shoutout, user: IdolMember): Promise<Shoutout> => {
+  const canEdit = await PermissionsManager.canHideShoutouts(user);
+  if (!canEdit) {
+    throw new PermissionError(
+      `User with email: ${user.email} does not have permission to hide shoutouts!`
+    );
+  }
+  await ShoutoutsDao.hideShoutout(body);
+  return body;
```
henry-li-06, line 39: It seems like the `hideShoutout` handler is just updating the shoutout document with whatever is provided from the frontend. I think it makes more sense for this handler to just take the uuid of the shoutout that we want to hide and then update the document with the the new hidden field in the dao layer. e.g. 

ShoutoutAPI.ts
```
const hideShoutout = (uuid, user) => {
    ...
    await ShoutoutsDao.hideShoutout(uuid)
```
ShoutoutsDao.ts
```
static async hideShoutout(uuid) {
    ...
    await shoutousCollection.doc(uuid).update({ hidden: true })
```

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.module.css
```diff
@@ -42,9 +42,24 @@
...
+  justify-content: space-between !important;
+  margin-top: 0 !important;
+  margin-bottom: 0 !important;
+}
+
+.shoutoutMessage {
+  color: black;
+  font-weight: bold;
+  margin-top: 0 !important;
+  font-size: 15px !important;
```
henry-li-06, line 60: I think it's best to avoid `!important` if possible

henry-li-06:
Also please link the notion issue for PRs in the future! Just makes reviewing PRs a lot easier :D

backend/src/dao/ShoutoutsDao.ts
airaagrawal1: Makes the function more consistent with how other `getAll()` functions are structured. For example in `CandidateDeciderDao.ts `

airaagrawal1:
> Also please link the notion issue for PRs in the future! Just makes reviewing PRs a lot easier :D

Good point! Just added link for this PR and will continue to do in future!

backend/src/API/shoutoutAPI.ts
airaagrawal1: I was originally planning to do this, but do you think there is any use case for updating other fields of a Shoutout? Thus, `hideShoutout` could be converted to a function like `updateShoutout` and won't just update the `hidden` field.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.module.css
airaagrawal1: I just took out a few, but the rest directly impact the formatting of the page.

backend/src/API/shoutoutAPI.ts
henry-li-06: yeah that's a good point -- maybe instead the dao layer can just be an update operation that takes in the updated shoutout object. but in the API layer i think the function should still just take in just a uuid and pass in the update to the dao layer

airaagrawal1: Would this require us to have a getShoutout(uuid) function in the dao layer to check if the shoutout exists? I'm just unclear as to how having the uuid as the only parameter into the API layer will allow updated information to get to the dao layer

JacksonStaniec: You can check whether the shoutout exists by:
```
const doc = await shoutoutsCollection.doc(uuid).get();
```
If doc is `null`, then it doesnt exist. The reason we only need to the `uuid` in the API layer is because we just want to update the `hidden` field of the shoutout to be `true`, so we don't need to pass in the whole shoutout object.

airaagrawal1: @henry-li-06 For some reason passing the uuid is causing a "type error somewhere in the api-dao interaction, and the error is causing the api to fail with the internal server error" according to @JacksonStaniec. Jackson and I discussed this and we think it's best to leave the current implementation so we can use shoutouts tomorrow during all-hands, but revisit this issue after.

henry-li-06: I was referring to the gateway-api interaction. The gateway layer should just expecting the request body to be in the form of 
```
{
  uuid: string
}
```
Then the gateway layer should pass `req.body.uuid` to the API layer. The business logic in the API layer can look something like this: 
```
const hideShoutout = (uuid: string, user: IdolMember) => {
  // check permissions 
  // get the shoutout with uuid
  // check that shoutout exists 
  // create updated shoutout that's hidden 
  // call updateShoutout in the dao layer to update the shoutout 
}
```

henry-li-06: It seems like the reason there was a type error is here:
image
The frontend is passing `uuid` as the entire request body instead of `{ uuid }` so the gateway layer can't parse the `uuid` field from the request body.

airaagrawal1: @henry-li-06 hmm that still doesn't work for me and @JacksonStaniec

JacksonStaniec (approved):
I think its alright to have a general `updateShoutout` DAO function for now, but we can improve the API layer by only needing to provide a `uuid` for hiding shoutouts later. The priority right now is giving admins the ability to hide shoutouts from view in case that its needed.

PR #368, opened 2022-11-12
Fix bug with anonymous shoutouts

JacksonStaniec:
### Summary 

This PR resets the form to be anonymous by default after submitting, and makes the checkbox in sync with whether the shoutout is anonymous or not.

### Test Plan
Try submitting 2 anonymous shoutouts in a row, then uncheck the box to send a non anonymous one


airaagrawal1 (approved):
Good catch! LGTM!

PR #371, opened 2022-11-13
Add community events category for TEC

henry-li-06:
### Summary 

Add community category for TEC. All members will have required to have at least 1 community TEC and meet the total number of required TEC. 

### Notion/Figma Link 
[Notion link](https://www.notion.so/cornelldti/TEC-Add-community-category-30a26e32db4f43e79ec26e35b0791f9e)

### Test Plan 

On member-side dashboard, the remaining credits should only be 0 if the total number of approved credits if at least the required amount and the number of approved community credits is at least 1. Otherwise the remaining credits should be the higher between the required community credits and the total required credits.


backend/src/API/teamEventsAPI.ts
```diff
@@ -1,5 +1,4 @@
 import TeamEventsDao from '../dao/TeamEventsDao';
-import { TeamEvent } from '../types/DataTypes';
```
henry-li-06, line 2: For some reason the backend was still using a separate declaration of the `TeamEvent` type

backend/src/api.ts
```diff
@@ -246,7 +246,10 @@ loginCheckedPost('/signIn', async (req, user) => signIn(req.body.id, user));
 loginCheckedPost('/signInAll', async (_, user) => allSignInForms(user));
 
 // Team Events
-loginCheckedPost('/createTeamEvent', async (req, user) => createTeamEvent(req.body, user));
+loginCheckedPost('/createTeamEvent', async (req, user) => {
+  await createTeamEvent(req.body, user);
+  return {};
+});
```
henry-li-06, line 252: returning the value of `createteamEvent` was causing a type error and the frontend doesn't even read the value of the response so just returning `{}` here instead

backend/src/types/DataTypes.d.ts
```diff
@@ -56,22 +56,7 @@ export type DBTeamEvent = {
...
-
-export type TeamEvent = {
-  name: string;
-  date: string;
-  numCredits: string;
-  hasHours: boolean;
-  requests: TeamEventAttendance[];
-  attendees: TeamEventAttendance[];
-  uuid: string;
+  isCommunity: boolean;
```
henry-li-06, line 59: removing the duplicate declaration

frontend/src/components/Admin/TeamEvent/TeamEventForm.tsx
```diff
@@ -145,7 +147,23 @@ const TeamEventForm = (props: Props): JSX.Element => {
...
+            checked={isCommunity}
+            onChange={() => setIsCommunity(true)}
+          />
+        </Form.Field>
+        <Form.Field>
+          <Radio
+            label="No"
+            value="No"
+            checked={!isCommunity}
+            onChange={() => setIsCommunity(false)}
```
airaagrawal1, line 164: Screen Shot 2022-11-15 at 3 04 21 PM
Would it be better to have the 'yes' and 'no' radio buttons in line so that the form isn't unnecessarily long?

airaagrawal1 (commented):
I know that the scope of the notion task is to make a community category, but will we eventually implement the other categories? Should we implement this in a way where we can add other categories later on?

frontend/src/components/Admin/TeamEvent/TeamEventForm.tsx
henry-li-06: @airaagrawal1 both are good points (radio button and additional categories), let's discuss with @JacksonStaniec during work session today?

JacksonStaniec: Discussed offline: 
The community category itself is an experimental feature that the leads might not opt to use next semester, or might not continue to use if the next semester doesnt go as well as planned. So, we shouldnt invest the development resources into making a system of adding additional categories until we are more certain it will be used. We think the probability of having additional category requirements in the future is relatively low anyways, based on the current difficulty for many members to achieve 3 uncategorized credits. For the radio button, we decided taking up less vertical space is better.

frontend/src/components/Admin/TeamEvent/TeamEventForm.tsx
```diff
@@ -18,6 +18,7 @@ const TeamEventForm = (props: Props): JSX.Element => {
   const [teamEventDate, setTeamEventDate] = useState(teamEvent?.date || '');
   const [teamEventCreditNum, setTeamEventCreditNum] = useState(teamEvent?.numCredits || '');
   const [teamEventHasHours, setTeamEventHasHours] = useState(teamEvent?.hasHours || false);
+  const [isCommunity, setIsCommunity] = useState<boolean>(false);
 
   const submitTeamEvent = () => {
     if (!teamEventName) {
```
JacksonStaniec, line 24: I think the ability to edit the community field still needs to be added. This should be something like
```
  const [isCommunity, setIsCommunity] = useStat(teamEvent?.isCommunity || false);
```
and the community field isnt included in the team event object used for the `editTeamEvent` function later on.

JacksonStaniec (commented):
I think the inline radio buttons do look better :)

JacksonStaniec (approved):
Great work!

PR #376, opened 2022-11-16
Add backend support for sign-in form questions

henry-li-06:
### Summary 
This PR is a step towards adding optional questions to sign-in forms. It adds backend support to make it possible to create sign-in forms with an optional prompt and to require sign-in requests to have a response to the prompt if the form has a prompt. 

### Notion/Figma Link 

[Notion link](https://www.notion.so/cornelldti/Sign-in-Option-for-Adding-Questions-692bec37614d48a39b3d51982acc8858) (doesn't close this ticket) 
### Test Plan 

There isn't really a good way to test this at the moment... But check the deploy preview to ensure that the user-facing functionality hasn't broken. The sign-in backend is missing a test suite ;~;. Might be worth it to add a test suite for sign-in feature in the future.

### DIFF
--- a/backend/src/API/signInFormAPI.ts
+++ b/backend/src/API/signInFormAPI.ts
@@ -19,6 +19,7 @@ export const signInFormExists: (id: string) => Promise<boolean> = checkIfDocExis
 export const createSignInForm = async (
   id: string,
   expireAt: number,
+  prompt: string | undefined,
   user: IdolMember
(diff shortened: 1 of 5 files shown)


JacksonStaniec (approved):
Thank you for investigating this for the dev leads! This is a good base to build from if we end up reviving the sign-in feature in the future.

PR #377, opened 2022-11-17
Display message for infinite loading on sign-in 

henry-li-06:
### Summary 

Add a message with the infinite loading indicator if a user that to login into IDOL is not registered as a user. 

image

### Notion/Figma Link 

[Notion link](https://www.notion.so/cornelldti/Frontend-Gracefully-handle-login-for-non-members-a971c61611f64db5814aeae2b1e8b7b9)
### Test Plan 

Try to login to IDOL with not with cornell email -- you should see the message in the screenshot.

### DIFF
--- a/frontend/src/components/Common/FirestoreDataProvider.tsx
+++ b/frontend/src/components/Common/FirestoreDataProvider.tsx
@@ -83,7 +83,7 @@ export default function FirestoreDataProvider({ children }: Props): JSX.Element 
   const [adminEmails, setAdminEmails] = useState<readonly string[] | undefined>();
   const [members, setMembers] = useState<readonly IdolMember[] | undefined>();
   const [approvedMembers, setApprovedMembers] = useState<readonly IdolMember[] | undefined>();
-  const [isIDOLMember, setIsIDOLMember] = useState(false);
+  const [isIDOLMember, setIsIDOLMember] = useState(true);
 
   const userEmail = useUserEmail();
 
--- a/frontend/src/components/Common/FirestoreDataProvider.tsx
+++ b/frontend/src/components/Common/FirestoreDataProvider.tsx
@@ -5,6 +5,7 @@ import { adminsCollection, membersCollection, approvedMembersCollection } from '
 import { useUserEmail } from './UserProvider/UserProvider';
 import { Team } from '../../API/TeamsAPI';
 import { allowAdmin } from '../../environment';
+import { MembersAPI } from '../../API/MembersAPI';
 
 type ListenedFirestoreData = {
   readonly adminEmails?: readonly string[];
@@ -82,6 +83,9 @@ export default function FirestoreDataProvider({ children }: Props): JSX.Element 
   const [adminEmails, setAdminEmails] = useState<readonly string[] | undefined>();
   const [members, setMembers] = useState<readonly IdolMember[] | undefined>();
   const [approvedMembers, setApprovedMembers] = useState<readonly IdolMember[] | undefined>();
+  const [isIDOLMember, setIsIDOLMember] = useState(false);
+
+  const userEmail = useUserEmail();
 
   useEffect(() => {
     if (process.env.NODE_ENV === 'test') {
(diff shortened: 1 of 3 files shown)


frontend/src/components/Common/FirestoreDataProvider.tsx
```diff
@@ -112,7 +117,10 @@ export default function FirestoreDataProvider({ children }: Props): JSX.Element
         process.env.NODE_ENV === 'test' && children
       }
       {adminEmails == null || members == null || approvedMembers == null ? (
-        <Loader active={true} size="massive" />
+        <Loader style={{ fontSize: 12 }} active={true} size="massive">
+          {!isIDOLMember &&
+            'You may not be registered as a user within IDOL. Please use #idol-support to get assistance.'}
+        </Loader>
```
JacksonStaniec, line 123: I still see this error message when loading even when signing in with a cornell email. I think this is good to show when its stuck in the infinite loading, but otherwise it just looks like a mistake when users have an actual valid login and still see it.

JacksonStaniec (commented):
I do think its a small improvement to the user experience to have a message telling them to contact support if they're caught in the infinite loading screen. But, I think it would be a much more elegant solution to actually escape from the infinite loading, and give them the option to try again with their cornell email, or contact support if they need to be added to our system.

henry-li-06:
> I do think its a small improvement to the user experience to have a message telling them to contact support if they're caught in the infinite loading screen. But, I think it would be a much more elegant solution to actually escape from the infinite loading, and give them the option to try again with their cornell email, or contact support if they need to be added to our system.

Do you think it's worth to provide a way to escape from the infinite loading (logout probably)? Chances are (at least I'd imagine) they're logging in with their Cornell email anyways and just not added to IDOL. If that's the case they can't actually do anything about it.

JacksonStaniec:
> > I do think its a small improvement to the user experience to have a message telling them to contact support if they're caught in the infinite loading screen. But, I think it would be a much more elegant solution to actually escape from the infinite loading, and give them the option to try again with their cornell email, or contact support if they need to be added to our system.
> 
> Do you think it's worth it to provide a way to escape from the infinite loading (logout probably)? Chances are (at least I'd imagine) they're logging in with their Cornell email anyways and just not added to IDOL. If that's the case they can't actually do anything about it.

I think not having an infinite loading screen would be worth it, not only for ease of use but also for presentation. If they use their cornell email and aren't added, then the logout screen versus the loading screen is just cosmetic. But, if they accidentally click the wrong account (which I have done many times), you are basically stuck until the login cookie times out, unless we go in the backend and manually delete their non-cornell email from the authentication. Also, I think it is good presentation to account for non-DTI people who would attempt to login, so that we are more clear that this site is just for DTI members, and not the public facing cornelldti.org. This might leak into the other ticket dealing with cleaning up the login process, so if you want to just have this warning message for now, I can approve this PR.

henry-li-06:
> > > I do think its a small improvement to the user experience to have a message telling them to contact support if they're caught in the infinite loading screen. But, I think it would be a much more elegant solution to actually escape from the infinite loading, and give them the option to try again with their cornell email, or contact support if they need to be added to our system.
> > 
> > 
> > Do you think it's worth it to provide a way to escape from the infinite loading (logout probably)? Chances are (at least I'd imagine) they're logging in with their Cornell email anyways and just not added to IDOL. If that's the case they can't actually do anything about it.
> 
> I think not having an infinite loading screen would be worth it, not only for ease of use but also for presentation. If they use their cornell email and aren't added, then the logout screen versus the loading screen is just cosmetic. But, if they accidentally click the wrong account (which I have done many times), you are basically stuck until the login cookie times out, unless we go in the backend and manually delete their non-cornell email from the authentication. Also, I think it is good presentation to account for non-DTI people who would attempt to login, so that we are more clear that this site is just for DTI members, and not the public facing cornelldti.org. This might leak into the other ticket dealing with cleaning up the login process, so if you want to just have this warning message for now, I can approve this PR.

Sounds good, in that case I'll just close this PR for now and add those additional things.

PR #369, opened 2022-11-12
Show dev portfolio validation results in non-admin view.

JacksonStaniec:
### Summary 

This PR allows users to see the results of validation on their submission. This was previously blocked until we implemented the ability for admins to override the results of the validation:
- #354 

Changes:
- [x] Show validation results in both admin and non-admin view of details page
- [x] Add instructions to user-side details page to clarify grading policy and instruct them to reach out to dev leads if they have grading concerns
- [x] Show late deadline on details page

### Notion/Figma Link 

[Notion](https://www.notion.so/cornelldti/Idol-FA22-5a158866a8a1472c98b1f3c042b480f8?p=de7e03efa5b44f1d81e82b641aa0322e&pm=s)

### Test Plan 

See that validation results show up in non-admin view, but cannot override the results.
image

PR #380, opened 2022-11-18
Add info message about dev portfolios re-submissions

henry-li-06:
### Summary 
Add info message on dev portfolio form to indicate that re-submissions replace previous submissions.

### Notion/Figma Link 

[Notion link](https://www.notion.so/cornelldti/DP-Feedback-Make-it-more-clear-that-additional-submissions-are-resubmissions-32d065f5ceae49f095f2bb88e624907b)
### Test Plan 

👀


JacksonStaniec (approved):
For this ticket I had added a message about submissions replacing assignments in #369 for the details view, but I like this one better because its before they submit instead of after.

PR #382, opened 2022-12-07
Team Page Dark Half

andrew032011:
### Summary 

Implemented dark half of the team page

### Notion/Figma Link 

Figma: https://www.figma.com/file/7nWJN8aUvoN9B6T3StFJDq/FA22-Website-Designs?node-id=328%3A4437&t=xa74Et0q3BTFMlbV-0

Notion: https://www.notion.so/cornelldti/Implement-Team-Page-Dark-half-1f3116587cea44fabb719c6c0ad21967

### Notes 
- Added Donut graph, which is currently just an svg (not interactive). Thinking about it some more, we'll need to find a component that allows us to customize the text and its color, as well as the colors of the different sections of the pie chart. A survey of the possible options online doesn't really show this level of customizability. I haven't seen a component that allows for customizability of the color of the text, and some even allow for putting text on the pie chart itself, let alone customizing the font of the text. We might need to create our own component or extend off of an existing one. If so, that would go in a separate PR. For now, I've included the SVG from the Figma design.
- I think we could change the navbar in a separate PR


airaagrawal1 (commented):
Can you add the figma and notion task links, so it's easier to see what the page should look like?

airaagrawal1:
Screenshot 2022-12-07 at 2 22 41 PM
Is the page supposed to adjust if you zoom in/out? This is what happens if I make my window size bigger.

andrew032011:
> Can you add the figma and notion task links, so it's easier to see what the page should look like?

Added those, sorry totally forgot!

andrew032011:
> Screenshot 2022-12-07 at 2 22 41 PM
> 
> Is the page supposed to adjust if you zoom in/out? This is what happens if I make my window size bigger.

Just want to double check with @JacksonStaniec if that responsiveness is in the scope of this task or if we'll deal with it later?

JacksonStaniec:
> > Screenshot 2022-12-07 at 2 22 41 PM
> > Is the page supposed to adjust if you zoom in/out? This is what happens if I make my window size bigger.
> 
> Just want to double check with @JacksonStaniec if that responsiveness is in the scope of this task or if we'll deal with it later?

This was discussed offline, but just putting it here again since its been a while and so any newcomers can be caught up to this conversation:

> responsiveness is not in scope for this ticket, but i do find it strange that the title and paragraph can slide away from each other like that.  I think the title and paragraph should have the same position relative to each other no matter the screen size
> ...
> your "working together" paragraph keeps the relative positions the same

andrew032011:
Should be aligned now! @airaagrawal1 @JacksonStaniec

andrew032011:
Also, we discussed offline to save the donut graph for a separate ticket. For now we use an SVG.

JacksonStaniec (approved):

airaagrawal1 (approved):
LGTM!

PR #385, opened 2022-12-28
[WIP] Change styling for shoutouts display 

henry-li-06:
### Summary 
Changes the styling for the admin shoutouts display:
- ~~Add pagination for viewing shoutouts~~
image

- In present mode, move the `(From: <member>)` to be next to the shoutout recipient 
image

### Notion/Figma Link 
N/A

### Test Plan 
👀

### Notes 
These are mostly exploratory changes to improve the UI/UX for admin shoutouts. The changes are just proposed changes so please leave feedback.

### DIFF
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -168,3 +168,13 @@ interface DevPortfolioInfo {
 interface DevPortfolio extends DevPortfolioInfo {
   submissions: DevPortfolioSubmission[];
 }
+
+interface Shoutout {
+  readonly giver: IdolMember;
+  readonly receiver: string;
+  readonly message: string;
+  readonly isAnon: boolean;
+  readonly timestamp: number;
+  readonly hidden: boolean;
+  readonly uuid: string;
+}
--- a/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.module.css
+++ b/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.module.css
@@ -56,7 +56,8 @@
 }
 
 .presentShoutoutFrom {
-  margin-top: 0 !important;
+  font-size: 18px;
+  color: grey;
 }
 
 .shoutoutHide {
--- a/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
+++ b/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
@@ -120,25 +120,25 @@ const AdminShoutouts: React.FC = () => {
   };
 
   const DisplayList = (): JSX.Element => {
+    const displayedShoutouts = displayShoutouts;
+
     if (displayShoutouts.length === 0)
       return (
         <Card className={styles.noShoutoutsContainer}>
           <Card.Content>No shoutouts in this date range.</Card.Content>
         </Card>
       );
+
     if (view === 'PRESENT')
       return (
         <Item.Group divided>
-          {displayShoutouts.map((shoutout, i) => (
+          {displayedShoutouts.map((shoutout, i) => (
             <Item key={i}>
               <Item.Content>
-                <Item.Header
-                  className={styles.presentShoutoutTo}
-                >{`${shoutout.receiver}`}</Item.Header>
-                <Item.Meta
-                  className={styles.presentShoutoutFrom}
-                  content={` ${fromString(shoutout)}`}
-                />
+                <Item.Header className={styles.presentShoutoutTo}>
+                  {`${shoutout.receiver}`}{' '}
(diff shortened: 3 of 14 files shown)


andrew032011:
Maybe adding a customizable "number of shoutouts per page" setting would be good. Then the leads can control how much they want to scroll for when presenting these.

airaagrawal1:
> Maybe adding a customizable "number of shoutouts per page" setting would be good. Then the leads can control how much they want to scroll for when presenting these.

Can we make the default for this 20 or 25? Leads probably want more than 10 when presenting. Also, the number of shoutouts per page changes back to default when a user selects a different view - do we want this?

henry-li-06:
> > Maybe adding a customizable "number of shoutouts per page" setting would be good. Then the leads can control how much they want to scroll for when presenting these.
> 
> Can we make the default for this 20 or 25? Leads probably want more than 10 when presenting. Also, the number of shoutouts per page changes back to default when a user selects a different view - do we want this?

@JacksonStaniec

JacksonStaniec:
> > > Maybe adding a customizable "number of shoutouts per page" setting would be good. Then the leads can control how much they want to scroll for when presenting these.
> > 
> > 
> > Can we make the default for this 20 or 25? Leads probably want more than 10 when presenting. Also, the number of shoutouts per page changes back to default when a user selects a different view - do we want this?
> 
> @JacksonStaniec

I'm actually not too sure I like this pagination idea. Aira and I discussed something like this when trying to figure out present mode, and I believe we eventually decided that it would be easier to have a continuous scroll rather than a "slideshow" clicking next to display X amount of shoutouts at a time. Right now it seems like its just adding extra steps to going through the shoutouts, especially since they read all of the shoutouts at once and I can't imagine a scenario where being able to skip to "page 4" of the shoutouts would be convenient.

andrew032011:
I can see the pagination being a good workflow and very similar to the current format with slides if you set the number of shoutouts per page to be ~6 in present mode and then just clicking the next button every 5 shoutouts. Easier to not get lost in the pool of shoutouts/continuously scroll the entire way through.

henry-li-06:
@JacksonStaniec we can discuss pagination during work session today?

henry-li-06:
@JacksonStaniec there doesn't seem to be an easy way to add a carousel with semantic, did we decide to use highlighting over the current shoutout instead?

airaagrawal1:
> @JacksonStaniec there doesn't seem to be an easy way to add a carousel with semantic, did we decide to use highlighting over the current shoutout instead?

Was there a conclusion on this discussion? What if we add a small number to each shoutout card? This shows the shoutout count as well as helps mitigate the user getting lost in a long list.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.module.css
```diff
@@ -84,3 +85,8 @@
   margin-right: 2rem;
   padding-top: 2rem;
 }
+
+.paginationControlsContainer {
+  display: flex;
+  gap: 15px;
+}
```
JacksonStaniec, line 92: I think this can be removed since there's no pagination for now

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -120,25 +120,25 @@ const AdminShoutouts: React.FC = () => {
   };
 
   const DisplayList = (): JSX.Element => {
+    const displayedShoutouts = displayShoutouts;
+
```
JacksonStaniec, line 124: What is the purpose of assigning this state to a const?

common-types/index.d.ts
```diff
@@ -168,3 +168,13 @@ interface DevPortfolioInfo {
...
+
+interface Shoutout {
+  readonly giver: IdolMember;
+  readonly receiver: string;
+  readonly message: string;
+  readonly isAnon: boolean;
+  readonly timestamp: number;
+  readonly hidden: boolean;
+  readonly uuid: string;
+}
```
JacksonStaniec, line 180: I agree withing moving `Shoutouts` to its proper place, but can we document this change in the PR since there's no associated ticket with it ?

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
henry-li-06: Oops this was probably left on accident, probably from trying to add pagination

henry-li-06:
Alright there's a bunch of random changes in this PR that shouldn't be there... I'll make this PR WIP again and try to get rid of all those random changes... ;~;

PR #386, opened 2023-01-01
Add next-env.d.ts to gitignore

henry-li-06:
### Summary 

Title 

https://nextjs.org/docs/basic-features/typescript

### Notion/Figma Link 

https://cornelldti.slack.com/archives/C01BYPLV0KY/p1672199464929079
### Test Plan 

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ tsconfig.tsbuildinfo
 .env.test.local
 .env.production.local
 
+# next.js 
+next-env.d.ts
+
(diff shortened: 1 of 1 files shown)


andrew032011 (approved):
I believe we also want to run
```
git rm -rf --cached .
git add .
```
to remove the next-env files from the repository. But then we should manually add them back so that we still have them, just untracked. Right now if you try changing the files, they're still being tracked.

PR #391, opened 2023-02-13
Temporarily disable community events for TEC

JacksonStaniec:
### Summary 

The leads have decided not to have a separate community event tally this semester. However, this may be released in future semesters, so for now we should hide the ability to make Team Events "community" events, and hide the tally for community events from users.

### Notion/Figma Link 

[Notion](https://www.notion.so/cornelldti/Idol-SP23-e65533742dca4f898581cc5c496e821e?p=125511cf68374e268cc4edf42e88ec9b&pm=s)

### Test Plan 

Manually test to see that community events can no longer be created on the admin side, and that community event tallies are not visible on the member side.

### Notes 

I used a variable so we can toggle the community event visibility, instead of removing it or commenting it out. It is still spread out across several files which is not ideal, so im open to suggestions on how to accomplish this better.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -11,6 +11,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
+        with:
+          node-version-file: .node-version
       - name: Use Yarn Cache
         uses: actions/cache@v1
         with:
@@ -36,6 +38,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
+        with:
+          node-version-file: .node-version
       - name: Use Yarn Cache
         uses: actions/cache@v1
         with:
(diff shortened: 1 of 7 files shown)


andrew032011 (approved):
This seems like the right way to do it, after thinking about it and doing some research online. Issue here is we don't know when/how long it'll be until we use this code again.

I think however it might be slightly easier to manage if we define the `COMMUNITY_EVENTS` constant in one place. Could `environment.ts` be a potential spot? I know that's more for testing. 

Or define it in one of these files and export it, so you don't have 4 different flags to track.

JacksonStaniec:
> This seems like the right way to do it, after thinking about it and doing some research online. Issue here is we don't know when/how long it'll be until we use this code again.
> 
> I think however it might be slightly easier to manage if we define the `COMMUNITY_EVENTS` constant in one place. Could `environment.ts` be a potential spot? I know that's more for testing.
> 
> Or define it in one of these files and export it, so you don't have 4 different flags to track.

I do like that idea because it keeps it defined in one place, but we still run into the issue where we have to delete it everywhere its used anyways. I'll just make sure im the one tasked with undoing this when we release the community events feature eventually, but it shouldn't be too annoying.

PR #396, opened 2023-02-21
add name to contributors list

kembayeb:
### **Created Spring 2023 contributors list **
- Added my name to it


common-types/package.json
```diff
@@ -8,4 +8,4 @@
     "build": "exit 0"
   },
   "dependencies": {}
-}
+}
```
henry-li-06, line 11: Can you get rid of the extra line?

kembayeb: Its weird that that's showing, but on my end there is no extra line, just 11

henry-li-06: I think if you just do `git checkout main common-types/package.json` it should revert the change

andrew032011 (approved):

PR #404, opened 2023-03-06
update time labels for portfolio

mly32:
### Summary 

The backend create dev portfolio endpoint now sets deadlines to end of the day in the `America/New_York` time zone (and also the start of the day for earliest submission). Previously, it was using the local time zone of the server (which was not the expected behavior).

Also updated the portfolio submission UI's text, table, and late icon. Previously, only the date was shown; now, it is clear when something is past the deadline. The Table header was also bugged.

image

Ticket: <https://www.notion.so/cornelldti/DP-Bug-Deadlines-are-not-in-EST-8de856d9a9d14261accb5afbf3c91669?pvs=4>

### Test Plan 

Instead of doing logic in the test files, updated tests with hard-coded values.

### Notes

For some reason, when closing the sign-in popup (instead of signing in), the promise returns an error for popup cancelled, which was not being caught anywhere. Currently catching it to avoid warnings being shown. It's makes sense firebase throws an error since the user cancelled, but catching it wouldn't require any additional logic.

Sign-in still works fine though.

### Breaking Changes 

Switched to luxon date time library and removed moment (the project is dead).

Changed tests because the original logic was incorrect.

### DIFF
--- a/backend/src/API/devPortfolioAPI.ts
+++ b/backend/src/API/devPortfolioAPI.ts
@@ -4,9 +4,8 @@ import PermissionsManager from '../utils/permissionsManager';
 import { PermissionError, BadRequestError } from '../utils/errors';
 import { validateSubmission, isWithinDates } from '../utils/githubUtil';
 
-function zonedTime(timestamp: number, ianatz = 'America/New_York') {
-  return DateTime.fromMillis(timestamp, { zone: ianatz });
-}
+const zonedTime = (timestamp: number, ianatz = 'America/New_York') =>
+  DateTime.fromMillis(timestamp, { zone: ianatz });
 
 export const getAllDevPortfolios = async (user: IdolMember): Promise<DevPortfolio[]> => {
   const isLeadOrAdmin = await PermissionsManager.isLeadOrAdmin(user);
--- a/backend/src/API/devPortfolioAPI.ts
+++ b/backend/src/API/devPortfolioAPI.ts
@@ -1,7 +1,12 @@
+import { DateTime } from 'luxon';
 import DevPortfolioDao from '../dao/DevPortfolioDao';
 import PermissionsManager from '../utils/permissionsManager';
 import { PermissionError, BadRequestError } from '../utils/errors';
 import { validateSubmission, isWithinDates } from '../utils/githubUtil';
+
+function zonedTime(timestamp: number, ianatz = 'America/New_York') {
+  return DateTime.fromMillis(timestamp, { zone: ianatz });
+}
 
 export const getAllDevPortfolios = async (user: IdolMember): Promise<DevPortfolio[]> => {
   const isLeadOrAdmin = await PermissionsManager.isLeadOrAdmin(user);
@@ -47,21 +52,25 @@ export const createNewDevPortfolio = async (
     instance.name.length === 0 ||
     instance.deadline < instance.earliestValidDate ||
     (instance.lateDeadline && instance.lateDeadline < instance.deadline)
-  )
+  ) {
     throw new BadRequestError(
       `Unable to create the new dev portfolio instance: The provided dev portfolio is invalid.`
     );
-  const updatedDeadline = new Date(instance.deadline);
-  const updatedEarliestValidDate = new Date(instance.earliestValidDate);
-  const updatedLateDeadline = instance.lateDeadline ? new Date(instance.lateDeadline) : null;
-  const modifiedInstance = {
-    ...instance,
-    deadline: updatedDeadline.setHours(23, 59, 59),
-    earliestValidDate: updatedEarliestValidDate.setHours(0, 0, 0)
-  };
-  if (updatedLateDeadline) {
-    modifiedInstance.lateDeadline = updatedLateDeadline.setHours(23, 59, 59);
   }
+
+  const updatedDeadline = zonedTime(instance.deadline).endOf('day');
+  const updatedEarliestValidDate = zonedTime(instance.earliestValidDate).startOf('day');
+  const updatedLateDeadline = instance.lateDeadline
+    ? zonedTime(instance.lateDeadline).endOf('day')
+    : null;
+
+  const modifiedInstance: DevPortfolio = {
+    ...instance,
+    deadline: updatedDeadline.valueOf(),
+    earliestValidDate: updatedEarliestValidDate.valueOf(),
+    lateDeadline: updatedLateDeadline ? updatedDeadline.valueOf() : null
+  };
+
   return DevPortfolioDao.createNewInstance(modifiedInstance);
 };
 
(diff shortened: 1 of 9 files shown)


backend/src/API/devPortfolioAPI.ts
```diff
@@ -1,8 +1,13 @@
+import { DateTime } from 'luxon';
 import DevPortfolioDao from '../dao/DevPortfolioDao';
 import PermissionsManager from '../utils/permissionsManager';
 import { PermissionError, BadRequestError } from '../utils/errors';
 import { validateSubmission, isWithinDates } from '../utils/githubUtil';
 
+function zonedTime(timestamp: number, ianatz = 'America/New_York') {
```
andrew032011, line 7: Can we keep the functions consistent here? The other functions don't use the `function` keyword so maybe rewrite this one also as an anonymous function too.

frontend/src/components/Common/SignIn/SignIn.tsx
```diff
@@ -8,7 +8,9 @@ import { auth, provider } from '../../../firebase';
 
 const SignIn: React.FC = () => {
   const onGoogleSignIn = () => {
-    signInWithPopup(auth, provider);
+    signInWithPopup(auth, provider).catch((err) => {
+      console.error(err);
```
andrew032011, line 12: Did you mean to put this change in here? If so, it's probably small enough that it's not a big deal that it's unrelated. But worth noting in the PR description.

frontend/src/components/Admin/DevPortfolio/DevPortfolioDetails.tsx
```diff
@@ -89,6 +89,9 @@ const DevPortfolioDetails: React.FC<Props> = ({ uuid, isAdminView }) => {
     csvExporter.generateCsv(csvData);
   };
 
+  // console.log(new Date(portfolio?.lateDeadline));
```
andrew032011, line 92: Let's remove this commented out line here.

frontend/src/components/Admin/DevPortfolio/DevPortfolioDetails.tsx
```diff
@@ -239,9 +242,10 @@ const SubmissionDetails: React.FC<SubmissionDetailsProps> = ({
         negative={submissionStatus === 'invalid'}
         warning={submissionStatus === 'pending'}
       >
-        <Table.Cell rowSpan={`${numRows}`}>{`${submission.member.firstName} ${
-          submission.member.lastName
-        } (${submission.member.netid})${submission.isLate ? '*' : ''}`}</Table.Cell>
+        <Table.Cell rowSpan={`${numRows}`}>
+          {`${submission.member.firstName} ${submission.member.lastName} (${submission.member.netid})`}
+          {submission.isLate && <Icon name="exclamation circle" />}`
```
andrew032011, line 247: I like this for the UI much better than the * we used to have!

backend/tests/DevPortfolioAPI.test.ts
```diff
@@ -108,19 +113,11 @@ describe('User is lead or admin', () => {
   });
 
   test('createDevPortfolio should be successful', async () => {
-    const expectedDeadline = new Date(devPortfolio.deadline).setHours(23, 59, 59);
-    const expectedEarliestValidDate = new Date(devPortfolio.earliestValidDate).setHours(0, 0, 0);
-    const expectedLateDeadline = new Date(devPortfolio.lateDeadline).setHours(23, 59, 59);
-    await createNewDevPortfolio(devPortfolio, user);
+    const [input, output] = fakeCreateDevPortfolio();
```
andrew032011, line 116: 👍

andrew032011:
In general, please include the ticket link in the PR description! I went and found it on Notion, but it's definitely more helpful to have it attached to the PR.

frontend/package.json
```diff
@@ -18,6 +18,7 @@
     "csvtojson": "^2.0.10",
     "export-to-csv": "^0.2.1",
     "firebase": "^9.1.1",
+    "luxon": "^3.3.0",
     "moment": "^2.29.1",
```
henry-li-06, line 22: Let's get rid of the moment dependency if we're not using it anymore?

henry-li-06 (commented):
~~In `makeDevPortfolioSubmission`, the validation logic is~~
```
 if (!isWithinDates(Date.now(), devPortfolio.earliestValidDate, latestDeadline)) {
```
~~and I assume since it's still `Date.now()` in the local server time, it might cause some issues?~~

JacksonStaniec:
> I see it in the design, but I think you forgot to add the "exclamation circle" UI for the key down below (image attached). Also, if you're making this change on the admin side, it would make sense to also make the same updates on the member side too.
> 
> Now, this does start forcing you down a bit of a rabbit hole since the original ticket just said to fix the time zone issue. Feel free to pop that UI change entirely to a separate ticket too. @JacksonStaniec feel free to weigh in on this. I'm fine if we also make another ticket about this and just make this PR resolve both tickets (UI + timezone bug), just to keep things well documented.
> 
> Screen Shot 2023-03-10 at 1 03 20 AM

The [original ticket](https://www.notion.so/cornelldti/Idol-SP23-e65533742dca4f898581cc5c496e821e?p=8de856d9a9d14261accb5afbf3c91669&pm=s) just said to determine the root cause and make a ticket for the fix. @mly32 I think these changes are oke to have in one ticket since they are all related to how we handle deadlines, but I agree with @andrew032011 we should document it in a ticket.

backend/src/API/devPortfolioAPI.ts
mly32: makes sense :)

frontend/src/components/Common/SignIn/SignIn.tsx
mly32: updated PR description. TLDR: there was an error that shows up when you close the sign-in pop-up without signing in, however, there shouldn't be any logic change just because that error is thrown.

frontend/src/components/Admin/DevPortfolio/DevPortfolioDetails.tsx
mly32: 👍

mly32:
> > In `makeDevPortfolioSubmission`, the validation logic is
> > ```
> >  if (!isWithinDates(Date.now(), devPortfolio.earliestValidDate, latestDeadline)) {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and I assume since it's still `Date.now()` in the local server time, it might cause some issues?
> 
> Jk I think it's fine? Pls confirm though

Since the times are in UNIX timestamp which is timezone independent, the logic should hold.

frontend/src/components/Common/SignIn/SignIn.tsx
mly32: * I changed it so theres no console logging for it. I think the catch will still be necessary. Since the user shouldn't see the log that they canceled a popup.

andrew032011 (approved):

andrew032011:
Looks great! You say you don't like frontend, but you're super good at it.

mly32:
> Looks great! You say you don't like frontend, but you're super good at it.

LOL, ty. It's bc I do too much of it.

PR #409, opened 2023-03-11
updateShoutouts only fetches when allShoutouts is empty

oscarwang20:
### Summary 

- broke up the `if` block in `updateShoutouts` into three blocks:
  - first one checks if the date range is valid
  - second one updates `allShoutouts` by calling `getAllShoutouts()` if `allShoutouts` is empty
  - third one filters `allShoutouts` to display only shoutouts that are in the date range and/or hidden if it is not empty
- added `allShoutouts` to the dependency array for `updateShoutouts`

### Notion/Figma Link 
https://www.notion.so/cornelldti/Shoutouts-Fix-excess-reads-c9c10c482b644ae3a05e07f46f1b3b05?pvs=4

### Test Plan 

Screen Recording 2023-03-11 at 1 03 43 AM

### Notes 

- reloading the page still makes a call to the API, which allows users to fetch any additional shoutouts if they were added after the page loaded

### DIFF
--- a/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
+++ b/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
@@ -17,30 +17,31 @@ const AdminShoutouts: React.FC = () => {
   const [view, setView] = useState<ViewMode>('ALL');
 
   const updateShoutouts = useCallback(() => {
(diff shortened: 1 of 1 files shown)


henry-li-06 (approved):
Nice work! 
```
                __/___            
          _____/______|           
  _______/_____\_______\_____     
  \              < < <       |    
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #416, opened 2023-03-13
Fix profile update notifications with lodash deep equality 

henry-li-06:
### Summary 

Fix issue with profile update notification job by using lodash deep equality function to check for diffs instead of JSON.stringify. 

### Notion/Figma Link 

N/A
### Test Plan 

👀


andrew032011 (approved):
Out of curiosity, why didn’t stringifying the JSONs work to detect differences? Was there an edge case?

PR #422, opened 2023-03-14
Implemented Create, Update, and delete for TeamEventAttendanceDao

kembayeb:
### Summary 

This pull request is the first step towards implementing feature TeamEventAttendanceDao 

- implemented `createteamEventAttendance`
- implemented `updateteamEventAttendance`
- implemented `deleteteamEventAttendance`

### Notion/Figma Link 
https://www.notion.so/cornelldti/Implement-create-update-and-deleteTeamEventAttendance-methods-in-TeamEventAttendanceDao-6fbf330dc7b442029e85c91ce0d14163

### Test Plan 

Screen Shot 2023-03-14 at 5 14 24 PM
"
testing creating with and without the `hoursAttended` field"
tested updating Teameventattendance doc 
testing deleting

### DIFF
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -44,8 +44,8 @@ export default class TeamEventAttendanceDao {
   }
 
   /**
-   * Updates a TEC Attendance
-   * @param teamEventAttendance - updated TeamEventAttendance object
+   * Deletes a TEC Attendance
+   * @param uuid - DB uuid of TeamEventAttendance
    */
   static async deleteTeamEventAttendance(uuid: string): Promise<void> {
     const docRef = teamEventAttendanceCollection.doc(uuid);
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -1,1 +1,44 @@
-export default class TeamEventAttendanceDao {}
+import { v4 as uuidv4 } from 'uuid';
+import { memberCollection, teamEventAttendanceCollection } from '../firebase';
+import { DBTeamEventAttendance } from '../types/DataTypes';
+
+export default class TeamEventAttendanceDao {
+  static async createTeamEventAttendance(
+    teamEventAttendance: TeamEventAttendance
+  ): Promise<TeamEventAttendance> {
+    const teamAttendanceRef: DBTeamEventAttendance = {
+      member: memberCollection.doc(teamEventAttendance.member.email),
+      ...(typeof teamEventAttendance.hoursAttended === 'number'
+        ? { hoursAttended: teamEventAttendance.hoursAttended }
+        : {}),
+      image: teamEventAttendance.image,
+      eventUuid: teamEventAttendance.eventUuid,
+      pending: true,
+      uuid: teamEventAttendance.uuid ? teamEventAttendance.uuid : uuidv4()
+    };
+    await teamEventAttendanceCollection.doc(teamAttendanceRef.uuid).set(teamAttendanceRef);
+    return teamEventAttendance;
+  }
+
+  static async updateTeamEventAttendance(
+    attendance: TeamEventAttendance
+  ): Promise<TeamEventAttendance> {
+    const teamAttendanceRef: DBTeamEventAttendance = {
+      member: memberCollection.doc(attendance.member.email),
+      ...(typeof attendance.hoursAttended === 'number'
+        ? { hoursAttended: attendance.hoursAttended }
+        : {}),
+      image: attendance.image,
+      eventUuid: attendance.eventUuid,
+      pending: attendance.pending,
+      uuid: attendance.uuid
+    };
+    await teamEventAttendanceCollection.doc(teamAttendanceRef.uuid).update(teamAttendanceRef);
+    return attendance;
+  }
+
+  static async deleteTeamEventAttendance(uuid: string): Promise<void> {
+    const docRef = teamEventAttendanceCollection.doc(uuid);
+    docRef.delete();
+  }
+}
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -2,13 +2,13 @@ import { v4 as uuidv4 } from 'uuid';
 import { memberCollection, teamEventAttendanceCollection } from '../firebase';
 import { DBTeamEventAttendance } from '../types/DataTypes';
 
-/**
- * Creates a new TEC Attendance for member for a given team event
- * @param teamEventAttendance - Newly created TeamEventAttendance object.
- * If provided, the object uuid will be used. If not, a new one will be generated.
- * The pending field will be set to true by default.
- */
 export default class TeamEventAttendanceDao {
+  /**
+   * Creates a new TEC Attendance for member for a given team event
+   * @param teamEventAttendance - Newly created TeamEventAttendance object.
+   * If provided, the object uuid will be used. If not, a new one will be generated.
+   * The pending field will be set to true by default.
+   */
   static async createTeamEventAttendance(
     teamEventAttendance: TeamEventAttendance
   ): Promise<TeamEventAttendance> {
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -1,3 +1,4 @@
+import * as admin from 'firebase-admin';
 import { v4 as uuidv4 } from 'uuid';
 import { memberCollection, teamEventAttendanceCollection } from '../firebase';
 import { DBTeamEventAttendance } from '../types/DataTypes';
@@ -7,12 +8,11 @@ export default class TeamEventAttendanceDao {
     teamEventAttendance: TeamEventAttendance
   ): Promise<TeamEventAttendance> {
     const teamAttendanceRef: DBTeamEventAttendance = {
+      ...teamEventAttendance,
       member: memberCollection.doc(teamEventAttendance.member.email),
       ...(typeof teamEventAttendance.hoursAttended === 'number'
         ? { hoursAttended: teamEventAttendance.hoursAttended }
         : {}),
-      image: teamEventAttendance.image,
-      eventUuid: teamEventAttendance.eventUuid,
       pending: true,
       uuid: teamEventAttendance.uuid ? teamEventAttendance.uuid : uuidv4()
     };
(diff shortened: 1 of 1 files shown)


backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,44 @@
...
+import { DBTeamEventAttendance } from '../types/DataTypes';
+
+export default class TeamEventAttendanceDao {
+  static async createTeamEventAttendance(
+    teamEventAttendance: TeamEventAttendance
+  ): Promise<TeamEventAttendance> {
+    const teamAttendanceRef: DBTeamEventAttendance = {
+      member: memberCollection.doc(teamEventAttendance.member.email),
+      ...(typeof teamEventAttendance.hoursAttended === 'number'
+        ? { hoursAttended: teamEventAttendance.hoursAttended }
```
andrew032011, line 14: I would not write the logic this way. `hoursAttended` is an optional field, so it can either be `number` or `undefined`. But that doesn't matter here, since you can just pass `teamEventAttendance.hoursAttended` directly to the database, whether it's undefined or it's `number`. I think by default that field, if you pass in `undefined` will just default to 0 in the DB, which is fine.

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,44 @@
...
+        ? { hoursAttended: attendance.hoursAttended }
+        : {}),
+      image: attendance.image,
+      eventUuid: attendance.eventUuid,
+      pending: attendance.pending,
+      uuid: attendance.uuid
+    };
+    await teamEventAttendanceCollection.doc(teamAttendanceRef.uuid).update(teamAttendanceRef);
+    return attendance;
+  }
```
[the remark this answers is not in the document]
kembayeb: got it

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,44 @@
...
+    await teamEventAttendanceCollection.doc(teamAttendanceRef.uuid).set(teamAttendanceRef);
+    return teamEventAttendance;
+  }
+
+  static async updateTeamEventAttendance(
+    attendance: TeamEventAttendance
+  ): Promise<TeamEventAttendance> {
+    const teamAttendanceRef: DBTeamEventAttendance = {
+      member: memberCollection.doc(attendance.member.email),
+      ...(typeof attendance.hoursAttended === 'number'
```
andrew032011: Same thing here. I would just pass everything in except member.

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,44 @@
...
+  ): Promise<TeamEventAttendance> {
+    const teamAttendanceRef: DBTeamEventAttendance = {
+      member: memberCollection.doc(teamEventAttendance.member.email),
+      ...(typeof teamEventAttendance.hoursAttended === 'number'
+        ? { hoursAttended: teamEventAttendance.hoursAttended }
+        : {}),
+      image: teamEventAttendance.image,
+      eventUuid: teamEventAttendance.eventUuid,
+      pending: true,
+      uuid: teamEventAttendance.uuid ? teamEventAttendance.uuid : uuidv4()
```
andrew032011, line 17: +1 to Henry's comment, applied here too. You can just use the spread operator here, and then any fields that don't apply you can just assign them afterwards.

Look at `TeamEventsDao.createTeamEvent` for an example of that syntax. Fields that you specify explictly after the spread are "overwritten".

```
      requests: event.requests.map((req) => ({
        ...req,
        member: memberCollection.doc(req.member.email)
      })),
      attendees: event.attendees.map((att) => ({
        ...att,
        member: memberCollection.doc(att.member.email)
      }))
```

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,44 @@
...
+    };
+    await teamEventAttendanceCollection.doc(teamAttendanceRef.uuid).set(teamAttendanceRef);
+    return teamEventAttendance;
+  }
+
+  static async updateTeamEventAttendance(
+    attendance: TeamEventAttendance
```
andrew032011, line 24: This is my fault for writing the specs this way. I called the parameter here `attendance` but for `createTeamEventAttendance` I called it `teamEventAttendance`. We should have consistency everywhere this is used (and we haven't done a great job in some of our other dao classes, TBH). Let's call it `teamEventAttendance` everywhere. I just updated this in the ticket and the Design Doc. Again, this one's my fault.

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,53 @@
-export default class TeamEventAttendanceDao {}
+import * as admin from 'firebase-admin';
```
andrew032011, line 1: This can be simply written just as 

`import admin from 'firebase-admin';

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,53 @@
...
+      ...teamEventAttendance,
+      member: memberCollection.doc(teamEventAttendance.member.email),
+      ...(typeof teamEventAttendance.hoursAttended === 'number'
+        ? { hoursAttended: teamEventAttendance.hoursAttended }
+        : {}),
+      pending: true,
+      uuid: teamEventAttendance.uuid ? teamEventAttendance.uuid : uuidv4()
+    };
+    await teamEventAttendanceCollection.doc(teamAttendanceRef.uuid).set(teamAttendanceRef);
+    return teamEventAttendance;
```
andrew032011, line 20: I would return `teamEventAttendance` with the updated fields, not the original object. So `uuid` could potentially change to be the autogenerated one if it was an empty string originally, and `pending` will have been set to true no matter what.

andrew032011 (changes requested):

backend/src/dao/TeamEventAttendanceDao.ts
kembayeb: i tried this originally, but when u pass in undefined throws an error saying cant use 'undefined' in firebase. But i can set it to zero if it is undefined

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -1 +1,53 @@
...
+      ...attendance,
+      member: memberCollection.doc(attendance.member.email)
+    };
+
+    const docRef = await teamEventAttendanceCollection.doc(teamAttendanceRef.uuid);
+    const currentDoc = await docRef.get();
+    const currentAttendance = currentDoc.data() as DBTeamEventAttendance;
+
+    if (
+      teamAttendanceRef.hoursAttended === undefined &&
```
andrew032011, line 36: Not sure I understand the logic here. Are we only replacing the `hoursAttended` field data if it's undefined? Wouldn't that be covered just with the line below (`await docRef.update(teamAttendanceRef);`)? Does it matter if the original field was undefined or not?

andrew032011: This is probably related to what you did above with the `hoursAttended` field. Maybe there was some design issue discussed during work session that I'm not aware of that you could let me know about. I believe originally passing in undefined for hoursAttended field was fine, since we wouldn't use the field anyways if the event was not hour-based.

andrew032011 (changes requested):

backend/src/dao/TeamEventAttendanceDao.ts
andrew032011: Got it, makes sense. Personally I'd just pass in 0, or maybe `null` if that's allowed (we'd just have to change the types then since right now it's `number? = number | undefined`). And I'd use the nil-coalescing operator operator (`??`) for that to keep the logic short and simple to look at.

I want to hear from @JacksonStaniec @henry-li-06 though. Could you guys weigh in please?

backend/src/dao/TeamEventAttendanceDao.ts
andrew032011: Yeah, it's related. Ignore my comment for now while we wait for a response from @JacksonStaniec @henry-li-06.

backend/src/dao/TeamEventAttendanceDao.ts
andrew032011: Oh, you know what? I see the issue. You can pass in `null`, not `undefined` into firebase. But I set the type to be `number | undefined`.

andrew032011: The issue here is that Firebase currently does accept `undefined` values. But that's something we can change in a firebase config file. Once that's done, passing undefined to the DB will no longer lead to an error being thrown.

backend/src/dao/TeamEventAttendanceDao.ts
andrew032011: See above ^^ (`undefined` values will be enabled soon)

backend/src/dao/TeamEventAttendanceDao.ts
kembayeb: got it, thanks for all the help!

henry-li-06: @kembayeb after [this PR](https://github.com/cornell-dti/idol/pull/426) gets merged in, rebase your branch on top of staging and then you can get rid of this for this method and for the update method

andrew032011: Sure thing.

Jackson just got that fix in. Could you merge main into this branch and try it again, this time with `undefined` values?

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -24,21 +24,30 @@ export default class TeamEventAttendanceDao {
...
+      currentAttendance.hoursAttended !== undefined
+    ) {
+      await docRef.update(
+        new admin.firestore.FieldPath('hoursAttended'),
+        admin.firestore.FieldValue.delete()
```
kembayeb, line 41: @henry-li-06 @andrew032011 so here, after allowing undefined in firestore, if u originally `createteamevent attendance` and the `updateTeamEventAttendance` then the old `hoursAttended` value from the create remains, and doesnt  set to undefined. Instead of doing what i did here, is it ok, if I just set the value to zero if its undefined upon an update?

andrew032011: +1 to Henry’s comment. I’d probably insert the specification (which I provided in the ticket) here too, but I’d modify it to say if the hoursAttended field is originally undefined, it cannot be overwritten afterwards.

kembayeb: Can I include spec for the other 2

kembayeb: > If I'm understanding your question correctly -- realistically speaking I don't think there would ever be a case where the TeamEventAttendance object has `hoursAttended` populated and then is later updated to be `undefined`.
> 
> 
> 
> I think in practice, the update dao method will mostly be used to approve TEC requests where only the `pending` field gets set to `true`.

Ahh I see

henry-li-06 (approved):
This LGTM but this PR won't be able to get merged in until the PR that fixes CI into staging gets merged in. Hopefully that'll get done within the next day or two tho

andrew032011 (approved):

andrew032011:
Great work! And great to see you reducing code duplication throughout your code too.

PR #423, opened 2023-03-14
Change `src/*` -> `src/**/*`

JacksonStaniec:
### Summary 

This PR updates `tsconfig.json` in the backend directory to fix an issue with unresolved types that appears when some of us use VS Code.

### Test Plan 

`src/**/*` should match the same as `src/*`, so this technically shouldn't change anything.


andrew032011 (approved):
Someone's in the running for that "smallest PR" award!

PR #424, opened 2023-03-14
Use `.env` in frontend

JacksonStaniec:
### Summary 

This PR refactors configuration in the frontend directory to read from a `.env` file instead of boolean variables. Github CI tasks and the test suite had to be updated to support this change as well.

### Notion/Figma Link 

[Notion](https://www.notion.so/cornelldti/Idol-SP23-e65533742dca4f898581cc5c496e821e?p=5f49ffe553584f74a30b7d8c46e087e1&pm=s)

### Test Plan 

Should pass tests even in staging environment.


.github/workflows/ci-build-staging.yml
JacksonStaniec, line 1: This is the same as the `ci-build.yml` workflow except for the environment variables.

.github/workflows/ci-build-staging.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Type Check
+        run: yarn type-check
+      - name: Format
+        run: yarn format:check
+      - name: Test
+        env:
+          FIREBASE_PRIVATE_KEY: ${{ secrets.FIREBASE_PRIVATE_KEY }}
+          FIREBASE_PRIVATE_KEY_ID: ${{ secrets.FIREBASE_PRIVATE_KEY_ID }}
+          USE_PROD_DB: false
+          NEXT_PUBLIC_IS_STAGING: true
```
henry-li-06, line 35: Not sure if it's possible but oculd base the values off of `github.base_ref` 
https://docs.github.com/en/actions/learn-github-actions/contexts.

And then don't need two separate files.

.github/workflows/ci-build-staging.yml
```diff
@@ -0,0 +1,52 @@
+name: Build-Staging
+on:
+  push:
+    branches:
+      - staging
+  pull_request:
```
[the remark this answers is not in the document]
JacksonStaniec: When i combined the files I moved both `main` and `staging` to PRs since there shouldnt be any other possibilities to add to those branches. Would it be worth it add it to push as well just in case?

.github/workflows/ci-build-staging.yml
JacksonStaniec: I used [this hack](https://github.com/orgs/community/discussions/25725) to try to set the env vars with ternary operators, but we'll see if it works

.github/workflows/ci-build.yml
```diff
@@ -27,11 +27,24 @@ jobs:
...
+
+      - name: Set env var if base is staging
+        run: |
+          echo "USE_PROD_DB=false" >> $GITHUB_ENV
+          echo "NEXT_PUBLIC_IS_STAGING=true" >> $GITHUB_ENV
+        if: startsWith(github.base_ref, 'refs/heads/staging')
       - name: Test
         env:
           FIREBASE_PRIVATE_KEY: ${{ secrets.FIREBASE_PRIVATE_KEY }}
           FIREBASE_PRIVATE_KEY_ID: ${{ secrets.FIREBASE_PRIVATE_KEY_ID }}
```
henry-li-06, line 44: I think these need to be configured to use the dev firebase credentials when for CI tests into staging

henry-li-06:
@JacksonStaniec the value of `github.base_ref` is `refs/heads/<branch-name>`

JacksonStaniec:
> @JacksonStaniec the value of `github.base_ref` is `refs/heads/<branch-name>`

yea thats what the docs say but im trying to figure out why its setting the env vars to nothing. I might just go back to two files and figure this out later because we need the staging workflows to succeed.

.github/workflows/ci-build-staging.yml
JacksonStaniec: Currently trying [this one](https://dev.to/ku6ryo/how-to-set-env-parameters-conditionally-in-github-actions-1g6a)

.github/workflows/ci-build-staging.yml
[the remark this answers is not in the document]
henry-li-06: Yeah might be weorth adding push to `main` since that's what it originally had

.github/workflows/ci-build.yml
```diff
@@ -27,11 +27,26 @@ jobs:
...
+        run: |
+          echo "USE_PROD_DB=true" >> $GITHUB_ENV
+          echo "NEXT_PUBLIC_IS_STAGING=false" >> $GITHUB_ENV
+        if: startsWith(github.base_ref, 'refs/heads/main')
+
+      - name: Set env var if base is staging
+        run: |
+          echo "USE_PROD_DB=false" >> $GITHUB_ENV
+          echo "NEXT_PUBLIC_IS_STAGING=true" >> $GITHUB_ENV
+        if: startsWith(github.base_ref, 'refs/heads/staging')
```
henry-li-06: Does the `if` need to go before the `run`?

JacksonStaniec:
> > @JacksonStaniec the value of `github.base_ref` is `refs/heads/<branch-name>`
> 
> yea thats what the docs say but im trying to figure out why its setting the env vars to nothing. I might just go back to two files and figure this out later because we need the staging workflows to succeed.

The docs lied :( its apparently just the name instead of being preceded by "refs...". The environment variables are being set correctly by the conditionals, its just that I dont know how to set them for use by the call to `yarn test`.
image
I officially give up for now on keeping this one file, and im just gonna do a 2 file solution.

henry-li-06:
So I made some basic changes to fix the CI and then had some ideas for slightly more significant changes that I added in the subsequent commits that we should discuss. @andrew032011 please weigh in on this too. 

**Fixing CI:**
I think the CI changes you made were pretty much all good. It was just that the CI checks were configured to run on **all** branches which would break CI for this branch since `github.base_ref` doesn't exist for that since it isn't on a PR. I changed that to only run on pushes to `main`, which fixed our CI for staging. 

**Other Changes**
I thought it might be easier to use an environment variable `ENV` that can be used across the frontend and backend. This variable should really only be important in staging and production and doesn't have to be set locally. Since next.js requires frontend env to have the prefix `NEXT_PUBLIC`, I just added a command in building the frontend to set `NEXT_PUBLIC_ENV` to the value of `ENV`. I thought that the benefit of this is that we don't need to separately define the frontend and backend environments. 

The other changes I made were to disable some of the backend/frontend configurations in production settings. 

For the backend, when `ENV=staging || prod`, the value of `USE_PROD_DB` isn't read anymore. This lets us get rid of this configuration value in Github Actions and in Netlify since we'll just set the `ENV` value.

For the frontend:
- If `NEXT_PUBLIC_ENV` is `staging` or `prod`, the value of `USE_PROD_DB` isn't used anymore. We can also get rid of the rest for `useProdDb` since the FIrebase configuration that's used now is based on the environment. 
- If `isProduction === true` (`process.env.NODE_ENV==='production')`, then the `allowAdmin` flag gets basically disabled. This essentially disables this flag for staging and prod since the node env will be production for both.

andrew032011:
@henry-li-06 The changes look good to me. Should we figure out a way to get those API keys out and into env variables too while we're here?

henry-li-06:
> @henry-li-06 The changes look good to me. Should we figure out a way to get those API keys out and into env variables too while we're here?

Apparently firebase API keys don't need to be treated as secrets 🤷🏻‍♂️

PR #428, opened 2023-03-19
Add BaseDao class and refactor daos 

henry-li-06:
### Summary 

This PR adds a new `BaseDao` class that is intended to serve as a parent class for the other Dao classes by implementing some of the shared CRUD logic. 

Dao classes that inherit from `BaseDao` need to supply the data type, the db data type, a Firestore collection, a function for materializing data from a document, and a function for serializing data into a document. 

```
export default abstract class BaseDao<E, D> {
  readonly collection: firestore.CollectionReference<D>;

  readonly materializeData: (d: D) => Promise<E>;

  readonly serializeData: (e: E) => Promise<D>;

  constructor(
    collection: firestore.CollectionReference<D>,
    materializeData: (d: D) => Promise<E>,
    serializeData: (e: E) => Promise<D>
  ) {
    this.collection = collection;
    this.materializeData = materializeData;
    this.serializeData = serializeData;
  }
```

### Notion/Figma Link 

N/A

### Test Plan 

Did manually testing on IDOL, fixed existing backend test suite and all tests pass. 

### Notes 

This PR doesn't change the `SignInFormDao` since the methods' params are different than the other Dao classes so changing `SignInFormDao` to be consistent with the other Daos would require some bigger changes.

### DIFF
--- a/backend/src/dao/BaseDao.ts
+++ b/backend/src/dao/BaseDao.ts
@@ -1,6 +1,6 @@
 import { firestore } from 'firebase-admin';
 
-export default class BaseDao<E, D> {
+export default abstract class BaseDao<E, D> {
   readonly collection: firestore.CollectionReference<D>;
 
   readonly materializeData: (d: D) => Promise<E>;
--- a/backend/src/dao/BaseDao.ts
+++ b/backend/src/dao/BaseDao.ts
@@ -17,6 +17,11 @@ export default abstract class BaseDao<E, D> {
     this.serializeData = serializeData;
   }
 
+  /**
+   * Gets a document with docID
+   * @param docID
+   * @returns The document with id: docID
+   */
   protected async getDocument(docID: string): Promise<E | null> {
     const doc = await this.collection.doc(docID).get();
     const data = doc.data();
@@ -24,21 +29,40 @@ export default abstract class BaseDao<E, D> {
     return this.materializeData(data);
   }
 
+  /**
+   * @returns All documents in the collection
+   */
   protected async getAllDocuments(): Promise<E[]> {
     const docRefs = await this.collection.get();
     return Promise.all(docRefs.docs.map((doc) => this.materializeData(doc.data())));
   }
 
+  /**
+   * Creates a document with id: docID and data: data
+   * @param docID - The id of the document
+   * @param data - The data in the document
+   * @returns The new document
+   */
   protected async createDocument(docID: string, data: E): Promise<E> {
     const doc = await this.serializeData(data);
     await this.collection.doc(docID).set(doc);
     return data;
   }
 
+  /**
+   * Deletes document with id docID
+   * @param docID - The ide of the document to delete
+   */
   protected async deleteDocument(docID: string): Promise<void> {
     await this.collection.doc(docID).delete();
   }
 
+  /**
+   * Updates a document
+   * @param docID - The id of the document to update
+   * @param data - The updated data for the document
+   * @returns - The newly updated document
+   */
   protected async updateDocument(docID: string, data: E): Promise<E> {
     const doc = await this.serializeData(data);
     await this.collection.doc(docID).update(doc as firestore.UpdateData);
--- a/backend/src/dao/BaseDao.ts
+++ b/backend/src/dao/BaseDao.ts
@@ -1,0 +1,47 @@
+import { firestore } from 'firebase-admin';
+
+export default class BaseDao<E, D> {
+  readonly collection: firestore.CollectionReference<D>;
+
+  readonly materializeData: (d: D) => Promise<E>;
+
+  readonly serializeData: (e: E) => Promise<D>;
+
+  constructor(
+    collection: firestore.CollectionReference<D>,
+    materializeData: (d: D) => Promise<E>,
+    serializeData: (e: E) => Promise<D>
+  ) {
+    this.collection = collection;
+    this.materializeData = materializeData;
+    this.serializeData = serializeData;
+  }
+
+  protected async getDocument(docID: string): Promise<E | null> {
+    const doc = await this.collection.doc(docID).get();
+    const data = doc.data();
+    if (!data) return null;
+    return this.materializeData(data);
+  }
+
+  protected async getAllDocuments(): Promise<E[]> {
+    const docRefs = await this.collection.get();
+    return Promise.all(docRefs.docs.map((doc) => this.materializeData(doc.data())));
+  }
+
+  protected async createDocument(docID: string, data: E): Promise<E> {
+    const doc = await this.serializeData(data);
+    await this.collection.doc(docID).set(doc);
+    return data;
+  }
+
+  protected async deleteDocument(docID: string): Promise<void> {
+    await this.collection.doc(docID).delete();
+  }
+
+  protected async updateDocument(docID: string, data: E): Promise<E> {
+    const doc = await this.serializeData(data);
+    await this.collection.doc(docID).update(doc as firestore.UpdateData);
+    return data;
+  }
+}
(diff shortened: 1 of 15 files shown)


backend/src/dao/DevPortfolioDao.ts
```diff
@@ -13,40 +14,39 @@ export function devPortfolioSubmissionToDBDevPortfolioSubmission(
...
+      return fromDb;
+    })
+  );
+
+  return { ...data, submissions };
+}
 
-  public static async getDevPortfolio(uuid: string): Promise<DevPortfolio> {
-    const doc = await devPortfolioCollection.doc(uuid).get();
-    return this.DBDevPortfolioToDevPortfolio(doc.data() as DBDevPortfolio);
```
henry-li-06, line 49: Got rid of this method since `getInstance` does the same thing.

backend/src/dao/BaseDao.ts
andrew032011, line 1: It would be nice to add specifications to this base class so that it's easier to build off of.

backend/src/dao/CandidateDeciderDao.ts
```diff
@@ -2,8 +2,70 @@ import { v4 as uuidv4 } from 'uuid';
...
+      })),
+      comments: candidate.comments.map((comment) => ({
+        ...comment,
+        reviewer: memberCollection.doc(comment.reviewer.email)
+      }))
+    }))
+  };
+}
+
+async function materializeCandidateDeciderInstance(
```
andrew032011, line 29: Let's agree on a naming convention here. Here we call it `materialize....` and `serialize....`, while in DevPortfolio we called this `DBDevPortfolioToDevPortfolio` and `DevPortfolioToDBDevPortfolio`.

andrew032011 (commented):
Love this change overall! It's great that we're unionizing the daos and reducing code duplication. We should get some more eyes on this since it's such a big change.

andrew032011:
Another thought: it might be cleaner if we wait for [Implement deleteAllTeamEventAttendance and get dao methods for TeamEventAttendanceDao ](https://www.notion.so/cornelldti/Implement-deleteAllTeamEventAttendance-and-get-dao-methods-for-TeamEventAttendanceDao-6e9407c1e1834c28af665951c975a3ca?pvs=4) to go in first before this one, so you can conform your changes to the new `TeamEventAttendanceDao`.

henry-li-06:
> Have you considered using a singleton class for the dao classes moving forward? Seems like we've instantiated the same stateless MemberDao object 5 times here.
> 
> Also, it might be a good time to reconcile the naming for the dao methods here, for example CandidateDeciderDao uses `createNewInstance`, `getInstance`, `updateInstance`, `deleteInstance`. But `MembersDao` uses `deleteMember` `setMember` `updateMember`. But because of how huge of a change this is, it might be good to separate that out into a different ticket so we don't risk breaking too much at once.

1. I think the singleton pattern is generally considered an antipattern.
2. I think the reason that the naming is different is that the objects are different. For example, for the `MembersDao`, we're acting on **members** so we have "getMember", "deleteMember", etc. For `CandidateDeciderDao`, we're acting on candidate decider **instances** so we have "getInstance", "deleteInstance", etc. For `ShoutoutsDao`, we're acting on **shoutouts**... In essence, I think the way that the dao methods are named is fine since we're acting on different objects.

henry-li-06:
> Another thought: it might be cleaner if we wait for [Implement deleteAllTeamEventAttendance and get dao methods for TeamEventAttendanceDao ](https://www.notion.so/cornelldti/Implement-deleteAllTeamEventAttendance-and-get-dao-methods-for-TeamEventAttendanceDao-6e9407c1e1834c28af665951c975a3ca?pvs=4) to go in first before this one, so you can conform `TeamEventAttendanceDao` to your new changes.

I don't think the order that the PRs are merged is particularly important. We can just have a separate PR for those changes. I need to make a separate PR for SignInDao anyways.

andrew032011 (approved):
Looks good here! But again would love to get a few more pairs of eyes on this if possible. I'll bump it in #idol-dev.

backend/src/dao/ShoutoutsDao.ts
```diff
@@ -1,86 +1,59 @@
...
-    await shoutoutCollection.doc(shoutout.uuid).update(shoutoutRef);
-    return shoutout;
+  async updateShoutout(shoutout: Shoutout): Promise<Shoutout> {
+    return this.updateDocument(shoutout.uuid, shoutout);
   }
 
-  static async deleteShoutout(uuid: string): Promise<void> {
-    await shoutoutCollection.doc(uuid).delete();
+  async deleteShoutout(uuid: string): Promise<void> {
+    await this.deleteDocument(uuid);
```
JacksonStaniec, line 57: I see the conversation you and @andrew032011 had about these naming conventions. I also see you made the design decision to make the crud methods from the base dao protected, but I feel like these public wrapper methods can be redundant sometimes. I think the DAO layer as a whole would be easier to use and extend if naming was uniform for its basic operations `getDocument`, `deleteDocument`, etc. If logic needs to be slightly different before the call to the inherited method, there could be an override instead of a wrapper.

backend/src/dao/ShoutoutsDao.ts
```diff
@@ -1,86 +1,59 @@
...
-    const ref = await shoutoutDoc.get();
-    if (!ref.exists) throw new NotFoundError(`No shoutout '${shoutout.uuid}' exists.`);
-    const shoutoutRef: DBShoutout = {
-      ...shoutout,
-      giver: memberCollection.doc(shoutout.giver.email)
-    };
-    await shoutoutCollection.doc(shoutout.uuid).update(shoutoutRef);
-    return shoutout;
+  async updateShoutout(shoutout: Shoutout): Promise<Shoutout> {
+    return this.updateDocument(shoutout.uuid, shoutout);
```
JacksonStaniec, line 53: Along with the concrete and db types we parameterize the base dao on, could we pass information about which field should be used as the doc id (email for members, uuid for most other things)? Or even just an assumption that it will be `uuid`, and we can override it in case of exceptions.

JacksonStaniec (approved):
This is awesome. I didn't even know Typescript allowed generics, but this solution is clean. I think there is room for possible refinements and further unification, but this PR is a good start and I dont really want it getting bigger than it already is.

backend/src/dao/ShoutoutsDao.ts
JacksonStaniec: This would also allow us to implement the common `uuid ? uuid : uuidv4()` check by default, which could remove forgetting this as a potential source for bugs in the future

henry-li-06: I think it's hard to be able to make modifications to the data in `BaseDao` since we're working with generic types. TypeScript doesn't know that the type `E` or `D` has a field `uuid`. If we really wanted to make it so that we assume that there is a `uuid` field or to implement the common `uuid ? uuid : uuidv4()` logic, we'd have to use an interface like 
```
interface ObjectWithUUID {
  uuid: string;
}
```
that all of the types with uuid implement and we use this type instead of the generic types in `BaseDao`. But then the issue with this is that we can't use `BaseDao` anymore for members or sign-in since neither of those types have a uuid field. 

Similarly we also can't really pass information about which field should be used as the doc id since TypeScript doesn't know that that type exists on some generic type.

backend/src/dao/ShoutoutsDao.ts
henry-li-06: Somewhat related to what's below, but I think the dao class for each of the types still needs to implement its own create and update methods bc we're using generics and TypeScript isn't able to infer what the document id should just from generic types. We could definitely just remove the `delete` and `get` wrappers for each of the classes since they're really just wrappers, but then we're left with a class that sometimes uses `Document` and sometimes uses the type that we're acting on.

PR #431, opened 2023-03-21
implemented dao methods

oscarwang20:
### Summary 

Implemented dao methods to:
- delete all team event attendance
- get all team event attendance by user
- get all team event attendance by event
- get all team event attendance

### Notion/Figma Link 

https://www.notion.so/cornelldti/Implement-deleteAllTeamEventAttendance-and-get-dao-methods-for-TeamEventAttendanceDao-6e9407c1e1834c28af665951c975a3ca?pvs=4

### Test Plan 

Tests are from the following file:
```
import { fakeIdolMember } from '../../tests/data/createData';
import { teamEventAttendanceCollection } from '../firebase';
import TeamEventAttendanceDao from './TeamEventAttendanceDao';

const member1 = {
  netid: 'ow39',
  email: 'ow39@cornell.edu',
  firstName: 'Oscar',
  lastName: 'Wang',
  pronouns: '',
  graduation: '',
  major: '',
  hometown: '',
  about: '',
  subteams: '',
  role: 'developer',
  roleDescription: 'Developer'
};
const member2 = fakeIdolMember();

const testAttendance1: TeamEventAttendance = {
  member: member1,
  hoursAttended: 1,
  image: '',
  eventUuid: 'event1',
  pending: true,
  uuid: 'test1'
};

const testAttendance2: TeamEventAttendance = {
  member: member2,
  hoursAttended: 1,
  image: '',
  eventUuid: 'event1',
  pending: true,
  uuid: 'test2'
};

const testAttendance3: TeamEventAttendance = {
  member: member1,
  hoursAttended: 1,
  image: '',
  eventUuid: 'event2',
  pending: true,
  uuid: 'test3'
};

TeamEventAttendanceDao.createTeamEventAttendance(testAttendance1).then((res) => console.log(res));
TeamEventAttendanceDao.createTeamEventAttendance(testAttendance2).then((res) => console.log(res));
TeamEventAttendanceDao.createTeamEventAttendance(testAttendance3).then((res) => console.log(res));
TeamEventAttendanceDao.getAllTeamEventAttendance().then((res) => console.log(res));
TeamEventAttendanceDao.getTeamEventAttendance('event1').then((res) => console.log(res));
TeamEventAttendanceDao.getAllUsersTeamEventAttendance(member1).then((res) => console.log(res));
TeamEventAttendanceDao.deleteAllTeamEventAttendance().then((res) => console.log(res));
```

Get all team event attendance
image

Get team event attendance by event
image

Get team event attendance by user
image

### DIFF
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -1,6 +1,7 @@
 import { v4 as uuidv4 } from 'uuid';
-import { memberCollection, teamEventAttendanceCollection } from '../firebase';
+import { db, memberCollection, teamEventAttendanceCollection } from '../firebase';
 import { DBTeamEventAttendance } from '../types/DataTypes';
+import { deleteCollection } from '../utils/firebase-utils';
 import { getMemberFromDocumentReference } from '../utils/memberUtil';
 
 export default class TeamEventAttendanceDao {
@@ -58,24 +59,10 @@ export default class TeamEventAttendanceDao {
    */
   static async deleteAllTeamEventAttendance(): Promise<void> {
     const snapshot = await teamEventAttendanceCollection.get();
-    await this.deleteAllTeamEventAttendanceHelper();
+    await deleteCollection(db, 'team-event-attendance', 500);
     if (snapshot.size !== 0) {
       await this.deleteAllTeamEventAttendance();
     }
-  }
-
-  /**
-   * Helper function for deleting all TEC Attendance
-   * Deletes 500 documents at a time
-   */
-  private static async deleteAllTeamEventAttendanceHelper(): Promise<void> {
-    const snapshot = await teamEventAttendanceCollection.limit(500).get();
-    const batch = teamEventAttendanceCollection.firestore.batch();
-
-    snapshot.docs.forEach((doc) => {
-      batch.delete(doc.ref);
-    });
-    await batch.commit();
   }
 
   /**
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -102,10 +102,11 @@ export default class TeamEventAttendanceDao {
    */
   static async getAllTeamEventAttendance(): Promise<TeamEventAttendance[]> {
     const attendanceRefs = await teamEventAttendanceCollection.get();
+    const attendance = attendanceRefs.docs.map((doc) => doc.data() as DBTeamEventAttendance);
     return Promise.all(
-      attendanceRefs.docs.map(async (doc) => ({
-        ...doc.data(),
-        member: await getMemberFromDocumentReference(doc.get('member'))
+      attendance.map(async (att) => ({
+        ...att,
+        member: await getMemberFromDocumentReference(att.member)
       }))
     );
   }
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -57,10 +57,24 @@ export default class TeamEventAttendanceDao {
    * Deletes all TEC Attendance
    */
   static async deleteAllTeamEventAttendance(): Promise<void> {
-    const batch = teamEventAttendanceCollection.firestore.batch();
-    const attendanceRefs = await teamEventAttendanceCollection.get();
+    const snapshot = await teamEventAttendanceCollection.get();
+    await this.deleteAllTeamEventAttendanceHelper();
+    if (snapshot.size !== 0) {
+      await this.deleteAllTeamEventAttendance();
+    }
+  }
 
-    attendanceRefs.docs.forEach((doc) => batch.delete(doc.ref));
+  /**
+   * Helper function for deleting all TEC Attendance
+   * Deletes 500 documents at a time
+   */
+  private static async deleteAllTeamEventAttendanceHelper(): Promise<void> {
+    const snapshot = await teamEventAttendanceCollection.limit(500).get();
+    const batch = teamEventAttendanceCollection.firestore.batch();
+
+    snapshot.docs.forEach((doc) => {
+      batch.delete(doc.ref);
+    });
     await batch.commit();
   }
 
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -1,6 +1,7 @@
 import { v4 as uuidv4 } from 'uuid';
 import { memberCollection, teamEventAttendanceCollection } from '../firebase';
 import { DBTeamEventAttendance } from '../types/DataTypes';
+import { getMemberFromDocumentReference } from '../utils/memberUtil';
 
 export default class TeamEventAttendanceDao {
   /**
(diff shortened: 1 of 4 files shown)


backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -51,4 +52,67 @@ export default class TeamEventAttendanceDao {
...
+      attendanceRefs.docs.map(async (doc) => {
+        const { eventUuid, hoursAttended, image, member, pending, uuid } = doc.data();
+        return {
+          eventUuid,
+          hoursAttended,
+          image,
+          member: await getMemberFromDocumentReference(member),
+          pending,
+          uuid
+        };
```
henry-li-06, line 114: You can just use the spread operator on `doc.data()` for most of the values.

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -51,4 +52,67 @@ export default class TeamEventAttendanceDao {
...
+    coll.docs.forEach((doc) => batch.delete(doc.ref));
+    await batch.commit;
+  }
+
+  /**
+   * Updates a TEC Attendance
+   * @param teamEventAttendance - updated TeamEventAttendance object
+   *  If hoursAttended is originally defined, it cannot be overwritten afterwards
+   */
+  static async getAllUsersTeamEventAttendance(user: IdolMember): Promise<TeamEventAttendance[]> {
```
henry-li-06, line 72: I think it might be better to name this function `getTeamEventAttendanceByUser`

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -51,4 +52,67 @@ export default class TeamEventAttendanceDao {
...
+        member: await getMemberFromDocumentReference(att.member)
+      }))
+    );
+  }
+
+  /**
+   * Deletes a TEC Attendance
+   * @param uuid - DB uuid of TeamEventAttendance
+   */
+  static async getTeamEventAttendance(uuid: string): Promise<TeamEventAttendance[]> {
```
henry-li-06, line 88: Might be better to name this function `getTeamEventAttendanceByEventID`

henry-li-06 (commented):
Some of the method specs don't seem to match the actual method.

oscarwang20:
> Some of the method specs don't seem to match the actual method.

Just pushed the changes!

henry-li-06 (approved):
LGTM
```
                 |~
           |/    w
          / (   (|   \
         /( (/   |)  |\
  ____  ( (/    (|   | )  ,
 |----\ (/ |    /|   |'\ /^;
\---*---Y--+-----+---+--/(
 \------*---*--*---*--/
  '~~ ~~~~~~~~~~~~~~~
```

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -51,4 +52,59 @@ export default class TeamEventAttendanceDao {
...
+
+  /**
+   * Deletes all TEC Attendance
+   */
+  static async deleteAllTeamEventAttendance(): Promise<void> {
+    const batch = teamEventAttendanceCollection.firestore.batch();
+    const coll = await teamEventAttendanceCollection.get();
+
+    coll.docs.forEach((doc) => batch.delete(doc.ref));
+    await batch.commit();
```
andrew032011, line 64: This logic looks mostly correct. Just want to double check with @henry-li-06 @JacksonStaniec though. Does the batch get all the documents within the collection, or just a selection of them? I wonder if we tried this on a larger amount of data (i.e. 500+ documents), if this would fail. 

Reading up [here](https://cloud.google.com/firestore/quotas#writes_and_transactions). 

> 

Maximum number of writes that can be passed to a Commit operation or performed in a transaction | 500
-- | --




If the batch ends up going over 500, this delete would fail, since we can't make more than 500 writes at once. Also, the code assumes that the doc is within the batch. What happens when we call `batch.delete(...)` and the document is not in the batch? We don't have a fallback for that.

This wouldn't really be a problem for a bit, especially since we're not going to use this method for awhile, but I can see us getting up to 500 fast. If you don't end up making any changes here, I would at the very least mention in the specification that this could fail if there's 500+ documents, and we'd file a ticket to come back and fix this.

I found some recipes here for deleting all documents in a collection, one of them takes into account the 500 limit and splits up the batches into max 500 documents each: 

https://firebase.google.com/docs/firestore/manage-data/delete-data#collections

https://stackoverflow.com/questions/48175235/how-to-delete-multiple-documents-from-cloud-firestore

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -51,4 +52,59 @@ export default class TeamEventAttendanceDao {
...
+
+  /**
+   * Gets all TEC Attendance for all events
+   */
+  static async getAllTeamEventAttendance(): Promise<TeamEventAttendance[]> {
+    const attendanceRefs = await teamEventAttendanceCollection.get();
+    return Promise.all(
+      attendanceRefs.docs.map(async (doc) => ({
+        ...doc.data(),
+        member: await getMemberFromDocumentReference(doc.get('member'))
```
andrew032011, line 106: I don't think we need to call `get` from the doc every time. `doc.data()` should have the member information. I believe your logic here would double the amount of reads for each member (which we've been working to reduce these days). Since it's not too complicated to optimize here I would add that: save doc.data() in a variable (of type `DBTeamEventAttendance`) and then grab the `member` field off of that.

andrew032011 (changes requested):
Nice work! See my comments.

I see you also changed the names of some of the dao methods. That's fine with me. I'm curious, did you work with Jackson on renaming those? If we agree on some name changes, it would be good to update it in the ticket as well, so it's well documented. And I can make updates in the design doc too (since it might affect other tickets).

henry-li-06:
> Nice work! See my comments.
> 
> I see you also changed the names of some of the dao methods. That's fine with me. I'm curious, did you work with Jackson on renaming those? If we agree on some name changes, it would be good to update it in the ticket as well, so it's well documented. And I can make updates in the design doc too (since it might affect other tickets).

I asked for the method names to be changed.

backend/src/dao/TeamEventAttendanceDao.ts
andrew032011: Ah, great @oscarwang20 ignore the solutions I posted then! Looks like we already have a solution (which happens to be taken directly off the Firebase solution I linked).

andrew032011:
> > Nice work! See my comments.
> > I see you also changed the names of some of the dao methods. That's fine with me. I'm curious, did you work with Jackson on renaming those? If we agree on some name changes, it would be good to update it in the ticket as well, so it's well documented. And I can make updates in the design doc too (since it might affect other tickets).
> 
> I asked for the method names to be changed, the original names were very ambiguous.

Oh right, I see your comments above. I agree with those changes then. Once we merge this in I can update those names in the ticket + design doc + other tickets.

andrew032011:
> > > > Nice work! See my comments.
> > > > I see you also changed the names of some of the dao methods. That's fine with me. I'm curious, did you work with Jackson on renaming those? If we agree on some name changes, it would be good to update it in the ticket as well, so it's well documented. And I can make updates in the design doc too (since it might affect other tickets).
> > > 
> > > 
> > > I asked for the method names to be changed, the original names were very ambiguous.
> > 
> > 
> > Oh right, I see your comments above. I agree with those changes then. Once we merge this in I can update those names in the ticket + design doc + other tickets.
> 
> @andrew032011 I also like the new names. Will we update the names of the corresponding API endpoints as well?

I’ll revisit those tonight! If there’s updates I will let you know!

andrew032011:
Could you add the solution @henry-li-06 posted to delete the collection? I meant ignore the solutions I posted and use this code instead: https://github.com/cornell-dti/idol/blob/main/backend/scripts/update-dev-db.ts#L17-L56, since we already have a solution implemented for delete a collection correctly.

andrew032011:
> > > > Nice work! See my comments.
> > > > I see you also changed the names of some of the dao methods. That's fine with me. I'm curious, did you work with Jackson on renaming those? If we agree on some name changes, it would be good to update it in the ticket as well, so it's well documented. And I can make updates in the design doc too (since it might affect other tickets).
> > > 
> > > 
> > > I asked for the method names to be changed, the original names were very ambiguous.
> > 
> > 
> > Oh right, I see your comments above. I agree with those changes then. Once we merge this in I can update those names in the ticket + design doc + other tickets.
> 
> @andrew032011 I also like the new names. Will we update the names of the corresponding API endpoints as well?

Endpoint `/getAllUsersTeamEventAttendance` has been renamed to `/getTeamEventAttendanceByUser`. Updated in design doc and your ticket.

Other dao method name updates have also propagated through the tickets and design doc, so everything should be up to date now.

oscarwang20:
- Implemented `deleteAllTeamEventAttendanceHelper` to delete 500 documents at a time
- refactored `deleteAllTeamEventAttendance` to recursively call `deleteAllTeamEventAttendanceHelper` when there are still documents to delete after each batch deletion

andrew032011:
Feel free to message me on Slack if it’s not clear what I mean.

oscarwang20:
- moved `deleteCollection` method to `firebase-utils.ts`
- `update-dev-db.ts` and `teamEventAttendanceDao.ts` now both import `deleteCollection` from `firebase-utils.ts`

backend/src/utils/firebase-utils.ts
```diff
@@ -16,4 +17,45 @@ const configureAccount = (sa: any, isProd: boolean) => {
...
+  await batch.commit();
+
+  // Recurse on the next process tick, to avoid
+  // exploding the stack.
+  process.nextTick(() => {
+    deleteQueryBatch(db, query, resolve);
+  });
+}
+
 export default configureAccount;
```
andrew032011, line 61: I'd remove this line and update the imports that are affected.

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -58,26 +59,12 @@ export default class TeamEventAttendanceDao {
    */
   static async deleteAllTeamEventAttendance(): Promise<void> {
     const snapshot = await teamEventAttendanceCollection.get();
-    await this.deleteAllTeamEventAttendanceHelper();
+    await deleteCollection(db, 'team-event-attendance', 500);
     if (snapshot.size !== 0) {
       await this.deleteAllTeamEventAttendance();
     }
```
henry-li-06: I don't think this is needed anymore

andrew032011: +1. The logic in `deleteCollection` already takes care of ensuring that all the documents are deleted. You specify the batch size for the number in each batch that we delete (500 is the limit), and the function will keep deleting in batches of size 500. You don't need to recurse further.

andrew032011 (approved):

PR #443, opened 2023-04-11
refactored dropdown

mly32:
### Summary 

Created a shared component to handle dropdowns with labeled badges.

image
image

Due to the larger vertical spacing labels take up, the clear/select icon on the right isn't completely centered. I tried changing it but could not get it to work (perhaps due to my unfamiliarity with semantic UI). I don't think it's too ugly though.

### Notion/Figma Link 

<https://www.notion.so/cornelldti/DP-Form-Display-information-badges-in-dropdown-b92008ab24c5450cbd10ea0ed2eee5a6?pvs=4>

### Notes 

Removed search option for the drop-down since it was breaking some UI. +It won't really be needed considering the dropdown contents wouldn't go past a few dozen items.

### DIFF
--- a/frontend/src/components/Common/LabeledDropdown/index.tsx
+++ b/frontend/src/components/Common/LabeledDropdown/index.tsx
@@ -12,12 +12,12 @@ export interface LabeledDropdownProps extends DropdownProps {
 }
 
 const Item = ({ label, info = [] }: Data): JSX.Element => (
-  <div style={{ display: 'flex', alignItems: 'center', minHeight: 26 }}>
-    <div style={{ flex: 1 }}>{label}</div>
+  <div className={styles.itemContainer}>
+    <div className={styles.flex}>{label}</div>
     {info
-      .filter((x) => x !== undefined)
-      .map((x, i) => (
-        <Label key={i} content={x} />
+      .filter((infoLabel) => infoLabel !== undefined)
+      .map((infoLabel, i) => (
+        <Label key={i} content={infoLabel} />
       ))}
   </div>
 );
--- a/frontend/src/components/Common/LabeledDropdown/index.tsx
+++ b/frontend/src/components/Common/LabeledDropdown/index.tsx
@@ -1,0 +1,55 @@
+import { DropdownProps, Label, Dropdown } from 'semantic-ui-react';
+import styles from './LabeledDropdown.module.css';
+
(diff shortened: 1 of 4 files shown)


andrew032011:
I'd argue the search feature is important for TEC credits, maybe not Dev Portfolio. It would be nice to have that when the team events really start to stack up. If you want to leave that out of this PR, that's fine, but we should file a ticket to try and add that back.

andrew032011:
Looks great overall. Some minor comments. Love the idea of factoring this custom component out!

frontend/src/components/Common/LabeledDropdown/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <div style={{ flex: 1 }}>{label}</div>
+    {info
+      .filter((x) => x !== undefined)
+      .map((x, i) => (
+        <Label key={i} content={x} />
+      ))}
+  </div>
+);
+
+const LabeledDropdown = ({
```
kembayeb, line 25: i really do like this work that you've done with the dropdowns! @team I was wondering if we've ever considered changing the way we account for extensions, instead of being in the dropdown, like keep the dropdown as 1,2,3,4,5  and then do like a extensions check box in the side?

mly32:
> I'd argue the search feature is important for TEC credits, maybe not Dev Portfolio. It would be nice to have that when the team events really start to stack up. If you want to leave that out of this PR, that's fine, but we should file a ticket to try and add that back.

I can add search back in a different PR. I think I would be able to do it with Mantine select components instead.

frontend/src/components/Common/LabeledDropdown/index.tsx
mly32: We can ask designers about this. It might also involve a backend change because "extensions" are just a name we gave the assignments (e.g., "portfolio #1" and "portfolio #1 extension").

airaagrawal1: I like the idea of having an extensions checkbox on the side, but the leads don't always offer extensions for every portfolio, so this may not be necessary.

andrew032011:
Closing as this PR is getting stale, and we'd lose search as a result of the refactor. If we want to refactor more safely later on, we can always reference this PR.

PR #444, opened 2023-04-11
create app layout with mantine

mly32:
### Summary 

Test out Mantine UI, a different UI library, in replacement of Semantic UI.

I implemented two different kinds of app layouts. (Toggleable in code with the `NEW_STYLE` flag in `_app.tsx`).

Depending on feedback, I'll get rid of the other one.

The old sidebar did not properly register the nav items as "links" (no way to right-click or open in new tab). Both new designs fix that.

There does not seem to be any conflicts with Semantic UI (from what I can tell).

#### Layout 1 (new one)

This layout is more responsive (different view at a smaller screenwidth).

It also shows an example with a main menu and a submenu. In case there's a want for the navbar to show sub links such as `forms/devPortfolio`.

image
image
image

#### Layout 2 (similar to existing side bar)

The active link (shown in red in image 3) does not register if the current path is a sublink such as `/forms/devPortfolio` instead of `/forms`. Depending on the preferred design I can make the active color also show for sublinks, not sure. 

image
image
image

### Notion/Figma Link 

<https://www.notion.so/cornelldti/Modernize-IDOL-homepage-c84a5381155a46b5a8d9eeec8b14ca04?pvs=4>


airaagrawal1 (commented):
I personally prefer the first layout with the nav bar at the top. I think it's faster to navigate the website this way by requiring fewer clicks.

kembayeb:
agreed, I like layout one the best, everything just seems much more accessible

oscarwang20:
I also like Layout 1, but I'm wondering what the expanding "IDOL" acronym animation would do to the nav links. Would it push everything to the right or would it cover the text?

andrew032011:
Closing this for now as the migration won't be happening this semester, and because we are more likely to do a migration to TailwindCSS/ShadCN to achieve parity with the website's frontend stack. We can revisit a potential migration to MantineUI/some other library in the future. Can always come back to the git history in this PR too.

PR #437, opened 2023-03-26
Update TEC member and admin forms to integrate with new TeamEventAPI endpoints [WIP]

airaagrawal1:
### Summary 

- [x] delete endpoint `getAllTeamEventsForMember`
- [x] change parameters of `createTeamEvent` and `updateTeamEvent` to `teamEventInfo: TeamEventInfo`
- [x] add getTeamEventAttendanceByUser, updateTeamEventAttendance, and deleteTeamEventAttendance
- [x] update `TeamEventAPI` calls in member and admin TEC forms 
- [ ] return mock data in `TeamEventAPI.tsx`

### Notion/Figma Link 
https://www.notion.so/cornelldti/Idol-SP23-e65533742dca4f898581cc5c496e821e?p=ed86c780582f42299b2630ce6f00ec24&pm=s

### Test Plan 
Ensure that mock data in `TeamEventAPI.tsx` is being returned.

PR #445, opened 2023-04-13
Add admin TEC dashboard 

henry-li-06:
### Summary 

Add admin TEC dashboard to show all attendance for all events and all members. 
image

The name column on the left is sticky if the number of columns overflows: 
image

If a member has completed their TEC requirement, their name and total credit cells are highlighted green. 

### Notion/Figma Link 

[Notion link](https://www.notion.so/cornelldti/TEC-Admin-dashboard-to-glance-at-users-credits-49105c8ae1d64b179266e41d52357dbe?pvs=4)
### Test Plan 
👀

### DIFF
--- a/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
+++ b/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
@@ -17,14 +17,16 @@ const calculateMemberCreditsForEvent = (
   event: TeamEvent,
   isCommunity: boolean
 ): number =>
-  event.attendees.reduce((val: number, attendee) => {
-    if (attendee.member.email !== member.email || (isCommunity && !event.isCommunity)) {
-      return val;
-    }
-    if (event.hasHours && attendee.hoursAttended)
-      return val + Number(event.numCredits) * attendee.hoursAttended;
-    return val + Number(event.numCredits);
-  }, 0);
+  isCommunity && !event.isCommunity
+    ? 0
+    : event.attendees.reduce((val: number, attendee) => {
+        if (attendee.member.email !== member.email) {
+          return val;
+        }
+        if (event.hasHours && attendee.hoursAttended)
+          return val + Number(event.numCredits) * attendee.hoursAttended;
+        return val + Number(event.numCredits);
+      }, 0);
 
 const calculateTotalCreditsForEvent = (member: IdolMember, event: TeamEvent): number =>
   calculateMemberCreditsForEvent(member, event, false);
--- a/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
+++ b/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
@@ -1,0 +1,77 @@
+import React, { useState, useEffect } from 'react';
+import { Table, Header, Loader } from 'semantic-ui-react';
+import { useMembers } from '../../Common/FirestoreDataProvider';
+import { TeamEventsAPI } from '../../../API/TeamEventsAPI';
+import { REQUIRED_LEAD_TEC_CREDITS, REQUIRED_MEMBER_TEC_CREDITS } from '../../../consts';
+import styles from './TeamEventDashboard.module.css';
+
+const calculateMemberCreditsForEvent = (member: IdolMember, event: TeamEvent): number =>
+  event.attendees.reduce((val: number, attendee) => {
+    if (attendee.member.email !== member.email) {
+      return val;
+    }
+    if (event.hasHours && attendee.hoursAttended)
+      return val + Number(event.numCredits) * attendee.hoursAttended;
+    return val + Number(event.numCredits);
+  }, 0);
+
+const TeamEventDashboard: React.FC = () => {
+  const [teamEvents, setTeamEvents] = useState<TeamEvent[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+
+  const allMembers = useMembers();
+
+  useEffect(() => {
+    TeamEventsAPI.getAllTeamEvents().then((events) => {
+      setTeamEvents(events);
+      setIsLoading(false);
+    });
+  }, []);
+
+  if (isLoading) return <Loader active>Fetching team event data...</Loader>;
+
+  return (
+    <div className={styles.dashboardContainer}>
+      <Header as="h1" textAlign="center">
+        Team Event Dashboard
+      </Header>
+      <div className={styles.tableContainer}>
+        <Table celled selectable striped classname={styles.dashboardTable}>
+          <Table.Header>
+            <Table.HeaderCell className={styles.nameCell}>Name</Table.HeaderCell>
+            <Table.HeaderCell>Total</Table.HeaderCell>
+            {teamEvents.map((event) => (
+              <Table.HeaderCell>{event.name}</Table.HeaderCell>
+            ))}
+          </Table.Header>
+          <Table.Body>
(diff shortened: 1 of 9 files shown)


frontend/src/consts.ts
```diff
@@ -0,0 +1,4 @@
+// Team Event constants
```
andrew032011, line 1: This was long overdue. But don't we already have a `constants.d.ts`/`constants.js`? What's up with those? Can we combine?

Maybe another ticket to investigate and factor out more constants.

frontend/src/components/Admin/TeamEvent/TeamEvents.tsx
```diff
@@ -82,9 +82,13 @@ const TeamEvents: React.FC = () => {
...
           <TeamEventsDisplay isLoading={isLoading} teamEvents={teamEvents} />
         </div>
-        <div className={styles.resetButtonContainer}>
+        <div className={styles.buttonContainer}>
+          <Button>
+            <Link href="/admin/team-events/dashboard">View Team Events Dashboard</Link>
+          </Button>
           <ClearTeamEventsModal setTeamEvents={setTeamEvents} />
         </div>
+        <div></div>
```
andrew032011, line 91: Should we remove this empty `<div>` tag?

frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { TeamEventsAPI } from '../../../API/TeamEventsAPI';
+import { REQUIRED_LEAD_TEC_CREDITS, REQUIRED_MEMBER_TEC_CREDITS } from '../../../consts';
+import styles from './TeamEventDashboard.module.css';
+
+const calculateMemberCreditsForEvent = (member: IdolMember, event: TeamEvent): number =>
+  event.attendees.reduce((val: number, attendee) => {
+    if (attendee.member.email !== member.email) {
+      return val;
+    }
+    if (event.hasHours && attendee.hoursAttended)
```
andrew032011, line 13: This is OK, but I wonder if it makes it sense for this calculation to go into a `TeamEventsUtils` file. Aira's got something really similar here: https://github.com/cornell-dti/idol/pull/437.

andrew032011:
Woah! A lot of names that should not be there. Most of them are alumni. Do we need a ticket to clean this DB up?

frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <div className={styles.dashboardContainer}>
+      <Header as="h1" textAlign="center">
+        Team Event Dashboard
+      </Header>
+      <div className={styles.tableContainer}>
+        <Table celled selectable striped classname={styles.dashboardTable}>
+          <Table.Header>
+            <Table.HeaderCell className={styles.nameCell}>Name</Table.HeaderCell>
+            <Table.HeaderCell>Total</Table.HeaderCell>
```
andrew032011, line 42: I would add community total too, but figure out a way to hide it for now.

andrew032011:
Not that the leads could possibly forget who the rest of the leads are, but it might be nice to have some indication of who in the table is a lead. And it might be good to also put the TEC requirements for what Leads need and what members need. 

I think this will give some clarification/transparency to the leads for how we are calculating everything.

andrew032011:
Another thought, not in scope of this PR, it would be super cool if we had a warning for when the number of credits being requested is ridiculous (like mine in the screenshot, haha). Or also if someone has a duplication submission to a team event, it would be good to warn the admin approving it, just so they are aware.

This will require some separate design discussions. I can file a ticket for this.

andrew032011:
This table is alphabetically sorted by NetID. It might be easier to read if ordered by first or last name. 

Also we should take into account input grades, if there's a specific order. Or if they'd like a "export to CSV" feature with this to upload to Canvas (or wherever they calculate grades).

frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
henry-li-06: Good point but let's deal with that in separate PR after both of these are done.

frontend/src/components/Admin/TeamEvent/TeamEvents.tsx
henry-li-06: yup, definitely just forgot to remove it

frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
henry-li-06: sure, didn't notice that in the ticket

frontend/src/consts.ts
henry-li-06: So that file is in `common-types` which is intended to hold shared types/data/info/etc across more than one project. We can definitely look into investigating constants that are shared across projects but in this case, this is a frontend only constant so I think leaving it here should be fine.

henry-li-06:
> Woah! A lot of names that should not be there. Most of them are alumni. Do we need a ticket to clean this DB up? I forget...is the member collection synced up with the "Returning DTI Members" form responses?

I'm not sure what the exact state of the `members` collection is right now. I just checked though and it seems like there are 136 documents in the collection so there are definitely too many. I don't know if old members have been removed at all or only partly removed. That's probably a question for @JacksonStaniec. 

The whole IDOL members archive is definitely not updated though.

henry-li-06:
> Another thought, not in scope of this PR, it would be super cool if we had a warning for when the number of credits being requested is ridiculous (like mine in the screenshot, haha). Or also if someone has a duplication submission to a team event, it would be good to warn the admin approving it, just so they are aware. This will require some separate design discussions. I can file a ticket for this.

Yeah those are great ideas, we can discuss later today.

henry-li-06:
> This table is alphabetically sorted by NetID. It might be easier to read if ordered by first or last name.

It's probably better off sorted by netid since that's the source of truth in terms of identifying people. Especially in the context of grades.

henry-li-06:
> Also we should take into account input grades, if there's a specific order. Or if they'd like a "export to CSV" feature with this to upload to Canvas (or wherever they calculate grades). This feature would be a separate ticket, of course.

Yup there's already a [ticket](
https://www.notion.so/cornelldti/TEC-Export-to-csv-c0abddb6cdde4801aa3457438cc1c4a6?pvs=4) for exporting TEC info to CSV.

JacksonStaniec:
> > Woah! A lot of names that should not be there. Most of them are alumni. Do we need a ticket to clean this DB up? I forget...is the member collection synced up with the "Returning DTI Members" form responses?
> 
> I'm not sure what the exact state of the `members` collection is right now. I just checked though and it seems like there are 136 documents in the collection so there are definitely too many. I don't know if old members have been removed at all or only partly removed. That's probably a question for @JacksonStaniec.
> 
> The whole IDOL members archive is definitely not updated though.

The members collection is definitely not up-to-date with the current roster, since all changes need to be done manually. All I've done this semester is add newbies.  This is a great motivation to get an automatic way to sync member data with returning members surveys like @andrew032011 suggested. However, as @henry-li-06 and I have discussed previously, it is not easy to update the info from something like a csv because the schemas change every year, and its a lot of painful work to get it to a format we're looking for. We should definitely discuss this further offline tho.

andrew032011 (approved):

PR #446, opened 2023-04-16
Clear dev portfolio form inputs upon submission

henry-li-06:
### Summary 

~~This PR fixes a couple of small issues with the dev portfolio form regarding submission:
- ~~Clears the dropdown for selecting dev portfolio after submission~~ 
- ~~Clears the text area after submission~~ 
- ~~Clears form state only after the submission request is successful~~

This PR removes clearing the dev portfolio form after submission. 

### Notion/Figma Link 

N/A - From #idol-support (https://cornelldti.slack.com/archives/C044T59JD34/p1681666314088999)
### Test Plan 
The dev portfolio form should not be cleared after submission. It should be possible to re-submit the form by just clicking on the submit button a second time.

### DIFF
--- a/frontend/src/components/Forms/DevPortfolioForm/DevPortfolioForm.tsx
+++ b/frontend/src/components/Forms/DevPortfolioForm/DevPortfolioForm.tsx
@@ -117,12 +117,7 @@ const DevPortfolioForm: React.FC = () => {
         status: 'pending',
         ...(text && { text })
       };
-      sendSubmissionRequest(newDevPortfolioSubmission, devPortfolio).then(() => {
-        setDevPortfolio(undefined);
-        setOpenPRs(['']);
-        setReviewedPRs(['']);
-        setText('');
-      });
+      sendSubmissionRequest(newDevPortfolioSubmission, devPortfolio);
     }
   };
 
@@ -149,7 +144,7 @@ const DevPortfolioForm: React.FC = () => {
                 fluid
                 search
                 selection
-                value={devPortfolio?.name}
+                value={devPortfolio?.uuid}
                 options={devPortfolios
                   .sort((a, b) => a.deadline - b.deadline)
                   .map((assignment) => ({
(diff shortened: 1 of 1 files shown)


frontend/src/components/Forms/DevPortfolioForm/DevPortfolioForm.tsx
```diff
@@ -117,11 +117,12 @@ const DevPortfolioForm: React.FC = () => {
...
-      sendSubmissionRequest(newDevPortfolioSubmission, devPortfolio);
-      setDevPortfolio(undefined);
-      setOpenPRs(['']);
-      setReviewedPRs(['']);
-      setText(null);
+      sendSubmissionRequest(newDevPortfolioSubmission, devPortfolio).then(() => {
+        setDevPortfolio(undefined);
+        setOpenPRs(['']);
+        setReviewedPRs(['']);
+        setText('');
```
henry-li-06, line 124: For some reason setting text to `undefined` or `null` doesn't clear the `TextArea` 🤷🏻‍♂️

JacksonStaniec:
I was just writing a ticket for this. Do you think it would be more user-friendly if instead of clearing the form, we preserve the same information they just submitted? It seems like a pretty common use case where people want to resubmit immediately after submitting because they accidentally clicked too early or forgot something. I can't see a single case where clearing the form would be more convenient, since no one would want to submit to two different portfolios in succession.

henry-li-06:
> I was just writing a ticket for this. Do you think it would be more user-friendly if instead of clearing the form, we preserve the same information they just submitted? It seems like a pretty common use case where people want to resubmit immediately after submitting because they accidentally clicked too early or forgot something. I can't see a single case where clearing the form would be more convenient, since no one would want to submit to two different portfolios in succession.

Are you saying just the selected dev portfolio or everything (i.e. prs, text area) doesn't clear?

JacksonStaniec:
> > I was just writing a ticket for this. Do you think it would be more user-friendly if instead of clearing the form, we preserve the same information they just submitted? It seems like a pretty common use case where people want to resubmit immediately after submitting because they accidentally clicked too early or forgot something. I can't see a single case where clearing the form would be more convenient, since no one would want to submit to two different portfolios in succession.
> 
> Are you saying just the selected dev portfolio or everything (i.e. prs, text area) doesn't clear?

Everything doesn't clear. We also have had some user feedback in the past that its annoying to have to reinput everything just to resubmit. I think I have a better idea for how to solve that, but I think not clearing for now would accomplish basically the same goal.

andrew032011 (approved):

PR #468, opened 2023-05-09
firebase listener in shoutouts

kembayeb:
added firebase listener on frontend so that we can update allshoutouts lists, 

allows for updating shououts when doc is added and removed

right now, we are updating and removing in real time successfully, **however** it is just showing all shoutouts regardless if hidden or not,

I tried to setup an additional hook  so that once `allshououts` list has changed, then we should change the `displayList` as well to update its content based on the additions to `allshoutouts`, but that approach didn't seem to work. 
Also tried added all the functionality to the same useeffect,and creating a setdisplaylist hook

Any suggestions on a way that i can update  the displayShoutouts list once allShoutouts changes?

https://www.notion.so/cornelldti/Idol-SP23-e65533742dca4f898581cc5c496e821e?p=83f4ec1786ad4cbc8310691000dbe088&pm=s

Tested, and updates are happening in realtime minus issues stated above

### DIFF
--- a/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
+++ b/frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { collection, onSnapshot } from 'firebase/firestore';
-import { Button, Form, Item, Card, Modal, Header } from 'semantic-ui-react';
+import { Button, Form, Item, Card, Modal, Header, Loader } from 'semantic-ui-react';
 import DatePicker from 'react-datepicker';
 import { db } from '../../../firebase';
 import 'react-datepicker/dist/react-datepicker.css';
@@ -33,7 +33,7 @@ const getListTitle = (view: ViewMode): string => {
 
 const HideModal = (props: {
   shoutout: Shoutout;
-  setAllShoutouts: React.Dispatch<React.SetStateAction<Shoutout[]>>;
+  setAllShoutouts: React.Dispatch<React.SetStateAction<Shoutout[] | undefined>>;
 }): JSX.Element => {
   const { shoutout, setAllShoutouts } = props;
 
@@ -41,7 +41,7 @@ const HideModal = (props: {
     const oppHide = !shoutout.hidden;
     ShoutoutsAPI.hideShoutout(shoutout.uuid, oppHide).then(() => {
       setAllShoutouts((shoutouts) =>
-        shoutouts.map((s) => (s.uuid === shoutout.uuid ? { ...s, hidden: oppHide } : s))
+        shoutouts?.map((s) => (s.uuid === shoutout.uuid ? { ...s, hidden: oppHide } : s))
       );
       if (oppHide) {
         Emitters.generalSuccess.emit({
@@ -97,11 +97,14 @@ const DisplayList = ({
   view,
   setAllShoutouts
 }: {
-  displayShoutouts: Shoutout[];
+  displayShoutouts: Shoutout[] | undefined;
   view: ViewMode;
-  setAllShoutouts: React.Dispatch<React.SetStateAction<Shoutout[]>>;
+  setAllShoutouts: React.Dispatch<React.SetStateAction<Shoutout[] | undefined>>;
 }): JSX.Element => {
-  if (displayShoutouts.length === 0)
+  if (displayShoutouts === undefined) {
+    return <Loader active size="big" content="Loading..." />;
+  }
+  if (displayShoutouts?.length === 0)
     return (
       <Card className={styles.noShoutoutsContainer}>
         <Card.Content>No shoutouts in this date range.</Card.Content>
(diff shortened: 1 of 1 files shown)


frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -182,18 +203,18 @@ const AdminShoutouts: React.FC = () => {
...
+  //   const shoutoutDate = new Date(shoutout.timestamp);
+  //   return (
+  //     (view === 'ALL' ||
+  //       (view === 'PRESENT' && !shoutout.hidden) ||
+  //       (view === 'HIDDEN' && shoutout.hidden)) &&
+  //     shoutoutDate >= earlyDate &&
+  //     shoutoutDate <= lastDate
+  //   );
+  // })
+  // .sort((a, b) => a.timestamp - b.timestamp);
```
henry-li-06, line 217: This is what should be updating `displayShoutouts`. Since `allShoutouts` is part of the component state, whenever `setAllShoutouts` is called to update the state, it causes a re-render of the component. This code that's commented should run every render of the component to derive `displayShoutouts` from `allShoutouts`

kembayeb: ahh i fixed it, was missing the async keyword💔

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -173,7 +175,26 @@ const AdminShoutouts: React.FC = () => {
...
+    const unsubscribe = onSnapshot(shoutoutCollection, (snapshot) => {
+      snapshot.docChanges().forEach((document) => {
+        if (document.type === 'added') {
+          const newShoutout = document.doc.data() as Shoutout;
+          setAllShoutouts((oldShoutouts = []) => [...oldShoutouts, newShoutout]);
+        }
+        if (document.type === 'removed') {
+          const removedShoutout = document.doc.data() as Shoutout;
+          setAllShoutouts((updatedList = []) =>
+            updatedList.filter((x) => x.uuid !== removedShoutout.uuid)
```
andrew032011, line 189: As a general comment, I would name your variables to be more specific than `x`.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -218,9 +239,13 @@ const AdminShoutouts: React.FC = () => {
...
         <DisplayList
-          displayShoutouts={displayShoutouts}
+          displayShoutouts={allShoutouts?.length ? allShoutouts : []}
           view={view}
-          setAllShoutouts={setAllShoutouts}
+          setAllShoutouts={(shoutouts) => {
+            if (shoutouts) {
+              setAllShoutouts((shoutouts = []) => shoutouts);
+            }
+          }}
```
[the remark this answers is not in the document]
andrew032011: +1

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
kembayeb: i made these changes, however, the displayShoutouts list contents still doesnt update upon rerender when allShououts changes

henry-li-06: Hmm good catch... I think this is likely an issue with time zones on the server. We had a similar issue with deadlines for dev portfolio before. I'm not sure if that fix was also made everywhere else timestamps are used.

Either way I think this problem isn't within the scope of this PR. @andrew032011 should be something to look into in the future.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -173,7 +175,18 @@ const AdminShoutouts: React.FC = () => {
...
+    const unsubscribe = onSnapshot(shoutoutCollection, (snapshot) => {
+      const newShoutouts = snapshot.docs.map((docSnapshot) => {
+        const data = docSnapshot.data() as Shoutout;
+        return data;
+      });
+
+      setAllShoutouts(newShoutouts);
+    });
+    return unsubscribe;
+  }, [setAllShoutouts]);
```
henry-li-06, line 189: I'm not sure if we want to setup the snapshot listener for admin shoutouts for all view modes (i.e. not just presentation mode). Might be something that @andrew032011 and @JacksonStaniec can weigh in on.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -102,7 +104,7 @@ const DisplayList = ({
   if (displayShoutouts === undefined) {
     return <Loader active size="big" content="Loading..." />;
   }
-  if (displayShoutouts.length === 0)
+  if (displayShoutouts?.length === 0)
```
henry-li-06, line 107: I think we don't need this bc we check if `displayShoutouts` is `undefined` before.

PR #471, opened 2023-05-14
Permissions revamp

henry-li-06:
### Summary 

This pull request is the first step towards implementing feature Foo

- [x] implemented X
- [ ] fixed Y

- [ ] Updated backend api specification

### Breaking Changes 

### DIFF
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -177,6 +177,11 @@ const loginCheckedDelete = (
   handler: (req: Request, user: IdolMember) => Promise<Record<string, unknown>>
 ) => router.delete(path, loginCheckedHandler(handler));
 
+const loginCheckedPut = (
+  path: string,
+  handler: (req: Request, user: IdolMember) => Promise<Record<string, unknown>>
+) => router.put(path, loginCheckedHandler(handler));
+
 // Members
 router.get('/allMembers', async (_, res) => {
   const members = await allMembers();
--- a/backend/src/api.ts
+++ b/backend/src/api.ts
@@ -183,16 +183,20 @@ const loginCheckedPut = (
 ) => router.put(path, loginCheckedHandler(handler));
 
 // Members
-router.get('/allMembers', async (_, res) => {
-  const members = await allMembers();
+router.get('/member', async (req, res) => {
+  const type = req.query.type as string | undefined;
+  let members;
+  switch (type) {
+    case 'all-semesters':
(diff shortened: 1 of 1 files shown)


backend/src/utils/auth.ts
```diff
@@ -0,0 +1,122 @@
...
+  resource: string,
+  user: IdolMember,
+  rbacConfig: any
+): Promise<boolean> => {
+  const userRole = await getUserRole(user);
+  const resourceRBACConfig = rbacConfig.resources[resource];
+
+  if (userRole === 'admin') return true;
+
+  if (req.method === 'GET') {
```
andrew032011, line 49: I see you have another PR out to make our endpoints more RESTful. Hopefully all our endpoints follow the pattern of "GET" is for read-only, and everything else involves some kind of write operation.

backend/src/utils/auth.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (!user) {
+      res.status(401).send({ error: `No user with email: ${userEmail}` });
+      return;
+    }
+    if (env === 'staging' && !(await PermissionsManager.isAdmin(user))) {
+      res.status(401).json({ error: 'Only admins users have permismsions to the staging API!' });
+    }
+    // RBAC
+    if (
+      resource &&
```
andrew032011, line 84: I think it would be safer to reject the request if the resource is somehow not found in the config, rather than returning 200. It would alert us if we as the developer made a mistake/typo somewhere, or somehow messed up the config file. We could handle this check in `isAuthorized`.

backend/src/utils/auth.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  if (req.method === 'GET') {
+    if (resourceRBACConfig.attributes.includes('meta_only') && req.query.meta_only) return true;
+    const canReadRoles = resourceRBACConfig.read_only.push(...resourceRBACConfig.read_and_write);
+    if (canReadRoles.includes(userRole)) return true;
+  }
+
+  const canWriteRoles = resourceRBACConfig.read_and_write;
+  if (resourceRBACConfig.attributes.includes('email') && req.params.email) {
+    if (req.params.email === user.email) return true;
```
andrew032011, line 57: Cool, I understand this as if someone is requesting data specific to a person and they are that person, then this request overrides any role-based read/read-write permissions in the config. I personally think it reads better to put this check right under the admin userRole check.

PR #473, opened 2023-08-16
Rebrand to "Cornell Digital Technology & Innovation"

andrew032011:
### Summary 
As part of the rebranding, update README and two spots on the new website that contain the old team name with "Cornell Digital Technology & Innovation".

### Test Plan 

Manual. Did a Git grep to check for any instances of "Cornell Design & Tech Initiative", "Design & Tech", "Design and Tech", and "Initiative" and only found these three. Devs, feel free to put up a PR if you find any other ones.

### Notes 
We need to change the image on the new website. But that project is in progress, not public, and won't be released during Fall 2023.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ All the common types used by the packages are defined [here](./common-types/inde
 
 ## DTI
 
-We are a team within the **Cornell Design & Tech Initiative**. For more information, see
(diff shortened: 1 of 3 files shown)


michelleli01 (approved):
Looks good!

PR #475, opened 2023-09-04
Archive Spring 2023 Members

andrew032011:
### Summary 

- Store Spring 2023 Member collection as a JSON archive
- Modify archive script to pull Fa21 and Sp23 jsons

### Notes 

- Ideally, we would have an official script that does this. Historically we just scrapped one together each semester and ran it to generate the JSON, which, considering the frequency, isn't the worst thing.
- The purpose of saving this JSON's is to have a potential "Alumni" section in the future on the new DTI website and to be able to add back users which have been deleted (i.e. users who decided to take a semester off but then return in a different semester)

### Breaking Changes


oscarwang20 (approved):
Agreed with the scripting bullet point. We might also want to make an announcement for members to update their info. I see that several members who joined last semester don't have as much information as others. Also something to keep in mind for incoming members too just to keep the data more consistent.

PR #472, opened 2023-05-21
Change endpoints to be RESTful

henry-li-06:
### Summary 
This PR is the first of three PRs to revamp the permissions system within IDOL. 
- [x] Make endpoints RESTful  
- [ ] Add (express.js) routers to handle each resource (and sub-resource) API (e.g. `devPortfolioRouter` and `devPortfolioSubmissionRouter`)
- [ ] Modify permissions system:
  - [ ] Create RBAC rules for each resource and sub-resource (that needs permission checks) 
  - [ ]  Setup new auth roles 
  - [ ] Add auth middleware on each router based on the router's resource 

This PR includes endpoint level refactoring mostly to make endpoints RESTful: 
- Removed verbs from endpoints 
  - (e.g. `GET /member` instead of `GET /getMember`
- Added nested resource paths 
  - (e.g. `GET /dev-portfolio/:uuid/submission/:email` instead of `GET /getUsersDevPortfolioSubmissions/:uuid` 
- Unified some endpoints:
  - (e.g. `GET /team-event` gets all `TeamEvent` objects and `GET /team-event?meta_only` gets all `TeamEventInfo` objects 

### Notion/Figma Link 

N/A
### Test Plan 
Check that frontend endpoints match with the new backend endpoints (paths and query params).

PR #483, opened 2023-09-16
Update References to API

andrew032011:
### Summary 

Fixed failing job, which depends on an API endpoint that was changed in: https://github.com/cornell-dti/idol/pull/472, but I forgot to update the endpoint in the job here.


oscarwang20 (approved):
LGTM!

PR #485, opened 2023-09-17
Members/csv upload

oscarwang20:
### Summary 

This pr implements a feature in the admin edit users page to upload a csv file containing a user's email and any other updated information that can be updated through the edit users page. The csv file must have the user's email as a field, and the field names must be in `camelCase` (`"first name"` -> `"firstName"`)

### Notion/Figma Link 

https://www.notion.so/cornelldti/Import-and-Update-Member-Info-By-CSV-62243888a1484a22bde912ca167fd31f?pvs=4

### Test Plan 

`members.csv`
image
File selected:
image
File uploaded:
image
Changes:
image
Changes (cont):
image


frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +128,75 @@ export default function AddUser(): JSX.Element {
...
+      const csv = await csvFile.text();
+      const json = await csvtojson().fromString(csv);
+      const allNetIds = allMembers.map((m) => m.netid);
+      const members: IdolMember[] = [];
+      if (json[0].email) {
+        json.forEach((m) => {
+          const currMember = allMembers.find((mem) => mem.netid === getNetIDFromEmail(m.email));
+          if (currMember) {
+            members.push({
+              netid: getNetIDFromEmail(m.email),
```
andrew032011, line 142: Since you already calculated `getNetIDFromEmail(m.email)` in `currMember`, and you do it for every member, and you use it again here, how about you store that value in a variable first so we don't have to repeatedly calculate it?

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -182,6 +254,27 @@ export default function AddUser(): JSX.Element {
                 </Button>
               </div>
             </Card.Content>
+            <Card.Content>
+              <div className={`ui one buttons ${styles.fullWidth}`}>
```
andrew032011, line 258: UX improvement ideas: how about we only show the upload button if a file has been chosen already? And let's add some text here that says "Add/update users through CSV". And maybe let's even include a template file that the admin can download?

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +128,75 @@ export default function AddUser(): JSX.Element {
...
+              doubleMajor: m.doubleMajor || '',
+              minor: m.minor || '',
+              website: m.website || '',
+              linkedin: m.linkedin || '',
+              github: m.github || '',
+              hometown: m.hometown || '',
+              about: m.about || '',
+              subteams: [m.subteam] || [],
+              formerSubteams: m.formerSubteams || [],
+              role: m.role || ('' as Role),
```
andrew032011, line 179: Getting the role right is important. We should add an error if they don't include a "role" column. And I think we should also throw an error if not all the roles specified are one of the roles available "lead, tpm, pm, developer, designer, business". See if you can add this requirement without code duplication (i.e. redefining the role list again).

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +128,75 @@ export default function AddUser(): JSX.Element {
...
+              graduation: m.graduation || '',
+              major: m.major || '',
+              doubleMajor: m.doubleMajor || '',
+              minor: m.minor || '',
+              website: m.website || '',
+              linkedin: m.linkedin || '',
+              github: m.github || '',
+              hometown: m.hometown || '',
+              about: m.about || '',
+              subteams: [m.subteam] || [],
```
andrew032011, line 177: Can we do a validation check to make sure that the subteam actually exists?

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +128,75 @@ export default function AddUser(): JSX.Element {
...
+              formerSubteams: m.formerSubteams || [],
```
andrew032011, line 178: `formerSubteams` should have the same validation check as the `subteam` one.

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +128,75 @@ export default function AddUser(): JSX.Element {
...
+              minor: m.minor || currMember.minor,
+              website: m.website || currMember.website,
+              linkedin: m.linkedin || currMember.linkedin,
+              github: m.github || currMember.github,
+              hometown: m.hometown || currMember.hometown,
+              about: m.about || currMember.about,
+              subteams: [m.subteam] || currMember.subteams,
+              formerSubteams: m.formerSubteams || currMember.formerSubteams,
+              role: m.role || currMember.role,
+              roleDescription: getRoleDescriptionFromRoleID(m.role)
```
andrew032011, line 159: Can you do a null check on `m.role` here too, like you did below?

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +128,75 @@ export default function AddUser(): JSX.Element {
...
+    if (csvFile) {
+      const csv = await csvFile.text();
+      const json = await csvtojson().fromString(csv);
+      const allNetIds = allMembers.map((m) => m.netid);
+      const members: IdolMember[] = [];
+      if (json[0].email) {
+        json.forEach((m) => {
+          const currMember = allMembers.find((mem) => mem.netid === getNetIDFromEmail(m.email));
+          if (currMember) {
+            members.push({
```
andrew032011, line 141: I like to avoid code duplication if possible, but not while sacrificing readability. But I think we can do a little better here. Thoughts on this?

Create a generic "EMPTY_MEMBER" object and place it up top as a constant. Line 84 in this file can use this object to set the state.

```
const EMPTY_MEMBER = {
        firstName: '',
        lastName: '',
        pronouns: '',
        email: '',
        role: '' as Role,
        graduation: '',
        major: '',
        doubleMajor: '',
        minor: '',
        website: '',
        linkedin: '',
        github: '',
        hometown: '',
        about: '',
        subteams: [],
        formerSubteams: []
}
```

Then we use this same object to loop through the keys of the object:

```
const member = {};
Object.keys(EMPTY_MEMBER).forEach((field) => { 
 member[field] = m[field] ?? currMember[field]
}
members.push(member)
```
And then down below we can do 

```
const member = {...EMPTY_MEMBER};
Object.keys(EMPTY_MEMBER).forEach((field) => { 
 if(m[field])
   member[field] = m[field]
}
members.push(member)
```

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -182,6 +254,27 @@ export default function AddUser(): JSX.Element {
...
+                  {csvFile ? `Upload ${csvFile.name}` : 'Choose a CSV file'}
+                </Button>
+              </div>
+              <input
+                className={styles.fileUpload}
+                type="file"
+                accept=".csv"
+                onChange={(e) => setCsvFile(e.target.files?.[0])}
+              />
+              {uploadStatus ? <p>{uploadStatus}</p> : undefined}
```
andrew032011, line 276: UX suggestion: What if we made this red, so it shows up easier?

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -74,6 +75,8 @@ export default function AddUser(): JSX.Element {
     currentSelectedMember: allMembers[0],
     isCreatingUser: false
   });
+  const [csvFile, setCsvFile] = useState<File | undefined>(undefined);
+  const [uploadStatus, setUploadStatus] = useState<string>('');
```
andrew032011, line 79: More like, `uploadError`, at least based on how it's being used.

andrew032011 (changes requested):
Very nice work! This looks great. I think the leads are going to love this when the newbies get here.

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +146,46 @@ export default function AddUser(): JSX.Element {
     });
   }
 
+  async function uploadUsersCsv(csvFile: File | undefined): Promise<void> {
+    if (csvFile) {
+      const csv = await csvFile.text();
+      const json = await csvtojson().fromString(csv);
+      const allNetIds = allMembers.map((m) => m.netid);
+      const members: IdolMember[] = [];
+      if (json[0].email && json[0].role) {
```
andrew032011, line 155: The requirement for `role` should only be enforced for new members. For existing members, we should just keep the role the same if it's unspecified.

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +146,46 @@ export default function AddUser(): JSX.Element {
...
+            members.push(member);
+          } else {
+            const member = { ...EMPTY_MEMBER };
+            Object.keys(EMPTY_MEMBER).forEach((field) => {
+              if (m[field]) member[field] = m[field];
+            });
+            members.push(member);
+          }
+        });
+        members.forEach((m) => {
```
andrew032011, line 173: Hm, can we upload each member one by one by calling `MembersAPI.updateMember`/`setMember` in the `json.forEach` function? Instead of first populating a list, and then going through the list again to call the API for each one?

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +146,46 @@ export default function AddUser(): JSX.Element {
...
+            Object.keys(EMPTY_MEMBER).forEach((field) => {
+              member[field] = m[field] ?? currMember[field];
+            });
+            members.push(member);
+          } else {
+            const member = { ...EMPTY_MEMBER };
+            Object.keys(EMPTY_MEMBER).forEach((field) => {
+              if (m[field]) member[field] = m[field];
+            });
+            members.push(member);
```
andrew032011, line 170: I think these member objects are missing the `netid` and `roleDescription` fields.

andrew032011: Yep, those two fields are missing.

Screenshot 2023-09-19 at 12 33 00 PM

frontend/src/components/Admin/AddUser/AddUser.tsx
oscarwang20: It's also used to display a success message indicating how many members were uploaded, so `uploadError` is a little imprecise.
image

I've changed the hook to include a status and a message which is rendered as either a green success message or a red error message.
image
image

andrew032011: I agree. I didn't see that you had the `Uploaded` message originally. I like what you have now.

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,13 +146,64 @@ export default function AddUser(): JSX.Element {
...
+  }
+
   function setCurrentlySelectedMember(setter: (m: CurrentSelectedMember) => CurrentSelectedMember) {
     setState((s) => {
       if (!s.currentSelectedMember) return s;
       return { ...s, currentSelectedMember: setter(s.currentSelectedMember) };
     });
   }
 
+  function RenderUploadStatus(): JSX.Element {
```
andrew032011, line 194: nit: `UploadStatus`, since it is just a component. No need for `Render`

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,13 +146,64 @@ export default function AddUser(): JSX.Element {
...
+    if (uploadStatus) {
+      switch (uploadStatus.status) {
```
andrew032011, line 196: nit: How about just 
```
<p className={uploadStatus.status === 'error' ? styles.errorMessage : styles.successMessage}>{`${uploadStatus.msg}`}</p>;
```
and no switch statement?

frontend/src/components/Admin/AddUser/AddUser.tsx
andrew032011: You can insert these fields right before you push the `member` to `members` array (or if you make the revision I said above, just add the two fields right before you call `MembersAPI`.

andrew032011:
UI bug? I wouldn't block releasing this change from this but just something to note

Screenshot 2023-09-19 at 12 08 55 PM

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +129,75 @@ export default function AddUser(): JSX.Element {
...
+          graduation: m.graduation || currMember.graduation,
+          major: m.major || currMember.major,
+          doubleMajor: m.doubleMajor || currMember.doubleMajor,
+          minor: m.minor || currMember.minor,
+          website: m.website || currMember.website,
+          linkedin: m.linkedin || currMember.linkedin,
+          github: m.github || currMember.github,
+          hometown: m.hometown || currMember.hometown,
+          about: m.about || currMember.about,
+          subteams: [m.subteam] || currMember.subteams,
```
andrew032011, line 152: Empty arrays are truthy in JavaScript! If you test this out with a csv and you don't specify the subteam, the existing subteams will actually be deleted. Do `m.subteam ? [m.subteam] : currMember.subteams`

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +129,75 @@ export default function AddUser(): JSX.Element {
...
+          graduation: m.graduation || '',
+          major: m.major || '',
+          doubleMajor: m.doubleMajor || '',
+          minor: m.minor || '',
+          website: m.website || '',
+          linkedin: m.linkedin || '',
+          github: m.github || '',
+          hometown: m.hometown || '',
+          about: m.about || '',
+          subteams: [m.subteam] || [],
```
andrew032011, line 174: See above. rewrite as `m.subteam ? [m.subteam] : []`

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +129,75 @@ export default function AddUser(): JSX.Element {
...
+        MembersAPI.setMember(updatedMember);
+      }
+    });
+  }
+
+  async function uploadUsersCsv(csvFile: File | undefined): Promise<void> {
+    if (csvFile) {
+      const csv = await csvFile.text();
+      const json = await csvtojson().fromString(csv);
+      if (json[0].email) {
```
andrew032011, line 188: This logic is a little confusing to follow and would make refactoring/adding on difficult. I'd rewrite it like this:
```
if(!json[0].email) {
  setUploadStatus({ status: 'error', msg: 'Error: No email field!' });
}
else if(!json[0].role) {
  setUploadStatus({ status: 'error', msg: 'Error: No role field!' });
}
else {
  processJson(json);
  setUploadStatus({ status: 'success', msg: `Uploaded ${json.length} members` });
}
```

frontend/src/components/Admin/AddUser/AddUser.tsx
andrew032011: Screenshot 2023-09-19 at 10 20 56 PM

andrew032011 (changes requested):
How about some validation checks for role and subteam? I'd ignore an entry entirely if the role/subteam was wrong. And I'd list that out in my error message all the entries that were wrong.

oscarwang20:
Instead of checking for missing fields using `json[0]`, `uploadUsersCsv` now filters out the invalid fields. I'll try to make the error messages more descriptive so it's easier to identify the invalid rows.

I had to change it because the state updates are asynchronous and aren't always caught as `uploadUsersCsv` runs.

Alternatively, I could change the filtering to filter out the well-formatted members and submit those while displaying error messages for the improperly formatted users. Wondering what would be less annoying to deal with...

andrew032011:
> Alternatively, I could change the filtering to filter out the well-formatted members and submit those while displaying error messages for the improperly formatted users. Wondering what would be less annoying to deal with...

I vote for failing the entire process, but tell us specifically rows are wrong/off. I think it could get more annoying if you had some of them uploaded and some of them not because it can get hard to keep track.

andrew032011:
If you're missing the email or role column altogether though, that should be an overall error that you raise, not specific to any row (unless I guess they left that box empty).

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -125,6 +135,84 @@ export default function AddUser(): JSX.Element {
...
+  async function uploadUsersCsv(csvFile: File | undefined): Promise<void> {
+    if (csvFile) {
+      const csv = await csvFile.text();
+      const json = await csvtojson().fromString(csv);
+      const errors = json.filter(
+        (m) =>
+          !m.email ||
+          !m.role ||
+          !ALL_ROLES.includes(m.role as Role) ||
+          !validSubteams.includes(m.subteam)
```
andrew032011, line 199: You need to check if we've even specified a subteam in the first place (i.e. `m.subteam && !validSubteams.includes(m.subteam)`), since the `subteam` field is optional.

oscarwang20:
- Added two new error messages for missing email and role fields
- Error messages now list each invalid row and their invalid fields

Felt might also be worthwhile to check if a member's subteam is in both `subteam` and `formerSubteams`. Excel and Google Sheets also format cells with comma separated values as strings, so I changed the way that's handled as well.

andrew032011:
> Also, can we include a template .csv here? One that contains all the columns? At the moment, there's no intuitive way someone would know our field names unless they looked at the code.

And some helper text about what's required would be nice. Similar to what I'm adding for Candidate Decider here: https://github.com/cornell-dti/idol/pull/496

andrew032011 (changes requested):
Very nice work, and excellent error handling! The messages are very helpful. A few nits.

I manually tested your changes using the following CSVs. Please add an example CSV that includes ALL the possible columns and add some helper text that explains the format we expect the input in, and that `email` and `role` are required columns.

[Missing_one_email_role_misspell_subteam_Member_CSV_Upload_Example.csv](https://github.com/cornell-dti/idol/files/12711261/Missing_one_email_role_misspell_subteam_Member_CSV_Upload_Example.csv)
[Missing_email_Member_CSV_Upload_Example.csv](https://github.com/cornell-dti/idol/files/12711262/Missing_email_Member_CSV_Upload_Example.csv)
[Member_CSV_Upload_Example.csv](https://github.com/cornell-dti/idol/files/12711263/Member_CSV_Upload_Example.csv)

andrew032011 (approved):

PR #487, opened 2023-09-18
Add Workaround for Loader Showing up Properly inside Modal

andrew032011:
This PR adds a workaround to `App.css` to allow for SemanticUI `Loader` component to show up properly inside `Modal`s.

This is a known issue, and since SemanticUI is on longer being maintained, will not be fixed: https://github.com/Semantic-Org/Semantic-UI/issues/4014.

We use a workaround recommended here: https://github.com/Semantic-Org/Semantic-UI-React/issues/3133#
issuecomment-418766748

Before (example of attempting to place a loader in a modal, which doesn't show up):

Screenshot 2023-09-17 at 9 08 33 PM

After:
Screenshot 2023-09-17 at 9 07 07 PM

### DIFF
--- a/frontend/src/pages/App.css
+++ b/frontend/src/pages/App.css
@@ -22,3 +22,14 @@
   margin: 0;
   padding: 0;
 }
+
+/* Workaround for an unfixed bug with Loaders not showing up properly in Modals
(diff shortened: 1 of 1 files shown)


alyssayzhang (approved):
Ahh I see - thanks for looking into this!

PR #489, opened 2023-09-19
Fix REST Method Typo in CandidateDeciderAPI

andrew032011:
### Summary 

The `/candidate-decider/${uuid}/rating` endpoint was updated to PUT instead of POST, but I missed this when getting this change in https://github.com/cornell-dti/idol/pull/472.


oscarwang20 (approved):
LGTM!

PR #493, opened 2023-09-23
Remove Extraneous } in ShoutoutsAPI call

andrew032011:
Remove typo in ShoutoutsAPI call, an extraneous "}" that was accidentally introduced.

This should make "given" shoutouts visible again. Thanks @oscarwang20 for catching that!

(This PR might win the smallest PR award)


oscarwang20 (approved):
LGTM!

PR #495, opened 2023-09-24
Candidate Decider - Hide Latency from Listeners through Frontend State Updates

andrew032011:
### Summary 
Previously, there was a latency issue on the candidate decider related to the listeners needing about a 2-3 second delay to update the responses real-time. 

I have "hidden" the latency by updating frontend state instead of solely relying on Firestone listeners to wait for DB updates. 

When the listeners receives the new snapshot, the local state will be properly synced. Updates to ratings/comments now show up instantaneously on the candidate decider UI.

Also:
- Add save button to prevent multi-threading issues (when users spam the ratings really hard the UI glitches out. The save button lets them switch between ratings but then commit the rating they want)
- For some reason, we were setting the comment to be `''` every time we hit the save button, so if you hit the save button twice, you would clear out your comment

### Notion/Figma Link 

https://www.notion.so/cornelldti/Candidate-Decider-Hide-Latency-from-Listeners-through-Frontend-State-Updates-d020a49d69b3439b93380add5833d6e0?pvs=4

### Test Plan 
Recorded a demo video for the candidate decider which includes this instantaneous update functionality (skip to 1:50).

https://github.com/cornell-dti/idol/assets/59291082/13307952-fc5e-4882-8ea3-a8e5d777d008

### Breaking Changes

### DIFF
--- a/backend/src/API/candidateDeciderAPI.ts
+++ b/backend/src/API/candidateDeciderAPI.ts
@@ -14,26 +14,27 @@ export const createNewCandidateDeciderInstance = async (
 ): Promise<CandidateDeciderInfo> => {
   if (!(await PermissionsManager.isAdmin(user)))
     throw new PermissionError(
-      'User does not have permission to create new Candidate Decider instance'
+      'User does not have permission to create new Candidate Decider instances.'
     );
   return candidateDeciderDao.createNewInstance(instance);
 };
 
-export const toggleCandidateDeciderInstance = async (
-  uuid: string,
+export const updateCandidateDeciderInstance = async (
+  instanceEdit: CandidateDeciderEdit,
   user: IdolMember
-): Promise<void> => {
+): Promise<CandidateDeciderInstance> => {
   if (!(await PermissionsManager.isAdmin(user)))
-    throw new PermissionError(
-      'User does not have permission to create new Candidate Decider instance'
-    );
-  const instance = await candidateDeciderDao.getInstance(uuid);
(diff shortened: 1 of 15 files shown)


oscarwang20 (changes requested):
Can we add a check to see if the form inputs changed before we call the api in `handleRatingChange` and `handleCommentChange`? It would just use `getRating()` and `getComment()` to check if the current rating and comment are different from what they were originally.

oscarwang20:
UI suggestion: maybe we could also just have one save button to handle both the rating change and comment change. And to make it clear that we are saving both, we could put the button above the form field:
image

andrew032011:
> UI suggestion: maybe we could also just have one save button to handle both the rating change and comment change. And to make it clear that we are saving both, we could put the button above the form field:

Agreed, done.
Screenshot 2023-09-26 at 5 37 09 PM

andrew032011:
> Can we add a check to see if the form inputs changed before we call the api in `handleRatingChange` and `handleCommentChange`? It would just use `getRating()` and `getComment()` to check if the current rating and comment are different from what they were originally.

I agree with this to prevent excessive API calls, but I like the added UX of disabling the save button until at least one change (rating or comment) has been made.  So I've done that instead. This behavior should also remind them to save too, since the button will light up once they've made a change.

https://github.com/cornell-dti/idol/assets/59291082/60d88386-8569-4938-ae92-d0ae4d3da09f

oscarwang20 (approved):
Looks great!

PR #515, opened 2023-10-01
Use Transactions for Candidate Decider Rating and Comment Updates

andrew032011:
This is a band-aid fix that will hopefully eliminate concurrency issues between users using the candidate decider at the same time, AND from requests from a single user that happen near-simultaneously. 

It uses firestore transactions in order keep the read/write operations atomic so that only they only happen one at a time, instead of in parallel.

### DIFF
--- a/backend/src/dao/CandidateDeciderDao.ts
+++ b/backend/src/dao/CandidateDeciderDao.ts
@@ -1,5 +1,5 @@
 import { v4 as uuidv4 } from 'uuid';
-import { candidateDeciderCollection, memberCollection } from '../firebase';
+import { candidateDeciderCollection, db, memberCollection } from '../firebase';
 import { DBCandidateDeciderInstance } from '../types/DataTypes';
 import { getMemberFromDocumentReference } from '../utils/memberUtil';
 import BaseDao from './BaseDao';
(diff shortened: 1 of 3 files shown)


backend/src/dao/CandidateDeciderDao.ts
```diff
@@ -94,4 +94,55 @@ export default class CandidateDeciderDao extends BaseDao<
...
+              ...comment,
+              reviewer: memberCollection.doc(comment.reviewer.email)
+            }))
+          }))
+      });
+    });
+
+    const newInstance = await this.getInstance(instance.uuid);
+    if (newInstance == null) throw new Error();
+    return newInstance;
```
henry-li-06, line 146: https://firebase.google.com/docs/firestore/manage-data/transactions#passing_information_out_of_transactions

I think you can just put the return inside the transaction callback

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -129,9 +129,9 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
           <Button
             className="ui blue button"
             disabled={currentComment === getComment() && currentRating === getRating()}
-            onClick={() => {
-              handleRatingAndCommentChange(currentCandidate, currentRating, currentComment);
-            }}
+            onClick={() =>
+              handleRatingAndCommentChange(currentCandidate, currentRating, currentComment)
+            }
```
henry-li-06, line 134: i'd get rid of this for git history reasons

henry-li-06 (approved):

backend/src/dao/CandidateDeciderDao.ts
henry-li-06: don't have to change btw if you don't want to re-test it

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
andrew032011: agreed

backend/src/dao/CandidateDeciderDao.ts
andrew032011: changed

PR #521, opened 2023-10-02
Tec/rejected bucket

oscarwang20:
### Summary 

- Replaces `pending` field with `status`, which can be either `approved`, `rejected`, or `pending`
- Added `reason` field for the reason why the tec attendance request was rejected
- Added documentation for `teamEventAPI.ts`
- Frontend displays and changes for everything mentioned above

### Notion/Figma Link 
https://www.notion.so/cornelldti/TEC-Add-Rejected-Bucket-Instead-of-Deleting-Rejected-TEC-Requests-e5604f13856042b7ad35b9f35879401f?pvs=4

### Test Plan 

Rejection and approval in firestore with new `status` and `reason` fields:
image

Member side: 
new display for rejected events
image

Admin side:
reason field disables reject button if empty
image

image

### Breaking Changes 

- Removed `pending` field for `team-event-attendance` collection
- Added `status` and `reason` fields for `team-event-attendance` collection


oscarwang20:
Regarding the members pending / members approved columns: what are people's thoughts on having three columns (pending, rejected, accepted)? I can imagine it might hinder readability when there are a lot of requests approved / rejected, so I'm wondering if anyone else might have thoughts on this.

As reference, here's what it is right now:
image

andrew032011:
How about pending and accepted from left to right?

And then below, we include rejected? That way most of the screen bandwidth is dedicated to pending + accepted and we can scroll down to see the others. Or, we can hide rejected and have a "Show Rejected" button to reveal them. I'd imagine the rejected bucket won't be something the reviewer is referencing often, but we should still make it available if they do want to see it.

backend/src/API/teamEventsAPI.ts
```diff
@@ -135,3 +204,20 @@ export const deleteTeamEventAttendance = async (uuid: string, user: IdolMember):
   }
   await teamEventAttendanceDao.deleteTeamEventAttendance(uuid);
 };
+
```
andrew032011, line 207: delete

backend/src/dao/deletePending.ts
```diff
@@ -0,0 +1,25 @@
+import * as firebase from 'firebase-admin';
+import { teamEventAttendanceCollection } from '../firebase';
+// import { createStatusFromPending } from '../API/teamEventsAPI';
```
andrew032011, line 3: delete comment

andrew032011 (commented):
Thanks for adding docstrings!

How are we going to convert the old data to conform to the new schema? We talked offline about a database migration. It would be good if you put it here, or if we chatted offline on how you're going to do this, so we can make this transition seamless.

backend/src/dao/deletePending.ts
oscarwang20: I was executing this file to modify the dev database. I'll remove it entirely once the changes are up on prod.

backend/src/API/teamEventsAPI.ts
oscarwang20: Shouldn't we keep the delete endpoint to allow users to cancel their requests? We can just keep the same code, but just remove the admin check.

alyssayzhang:
> How about pending and accepted from left to right?
> 
> And then below, we include rejected? That way most of the screen bandwidth is dedicated to pending + accepted and we can scroll down to see the others. Or, we can hide rejected and have a "Show Rejected" button to reveal them. I'd imagine the rejected bucket won't be something the reviewer is referencing often, but we should still make it available if they do want to see it.

I agree with this ➕ I think that mainly from the admin side, the admin wants to see the pending and accepted. We can have rejected below just in case the admin users needs to refer back to it. 

I'm not sure how many are rejected so if we did 3 columns (pending, rejected, accepted) maybe the middle column would not have a lot of TEC cards while the left and rightmost have a lot?

oscarwang20:
- Added a "Members Rejected" panel in `TeamEventDetails`
- Added `pending?: boolean` to `TeamEventAttendance` in `index.d.ts` and `DataTypes.d.ts` in preparation for prod db migration
- Keeping a local copy of `deletePending.ts` to run after merging to main
image

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -10,6 +10,7 @@ async function materializeTeamEventAttendance(
 ): Promise<TeamEventAttendance> {
   return {
     ...dbTeamEventAttendance,
+    status: dbTeamEventAttendance.status as Status,
```
andrew032011, line 13: Try changing `DBTeamEventAttendance` `status` field to be type `Status`. Then you can get rid of this line entirely, and we can use `Status` type everywhere instead of `string`.

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -19,6 +20,7 @@ async function serializeTeamEventAttendance(
 ): Promise<DBTeamEventAttendance> {
   return {
     ...teamEventAttendance,
+    status: teamEventAttendance.status as string,
```
andrew032011, line 23: This line is not needed.

andrew032011 (approved):
Nice! Do you have your script that you can send me? Maybe over Slack, and I can test this out locally before we run it on the production DB. And actually, before we even go about running the script on the production DB, it might be a good idea to get a backup stored somewhere. (Sigh, unfortunately we don't currently support DB backups, now might be a good reason to get them going).

frontend/src/components/Admin/TeamEvent/TeamEventCreditReview.tsx
```diff
@@ -24,7 +26,8 @@ const TeamEventCreditReview = (props: {
   const approveCreditRequest = (teamEventAttendance: TeamEventAttendance) => {
     const updatedTeamEventAttendance = {
       ...teamEventAttendance,
-      pending: false
```
andrew032011, line 27: Why don't we keep `pending` for now still? My reasoning is, if some of these changes break something on IDOL, we can always rollback the changes, and our DB will still be compatible with the old backend.

backend/src/API/teamEventsAPI.ts
```diff
@@ -75,13 +107,20 @@ export const requestTeamEventCredit = async (
       `User with email ${user.email} cannot request team event credit for another member, ${request.member.email}.`
     );
   }
-  const updatedteamEvent = { ...request, pending: true };
+  const updatedteamEvent = { ...request, status: 'pending' as Status };
```
andrew032011, line 110: See my comment regarding keeping `pending`.

backend/src/dao/TeamEventAttendanceDao.ts
```diff
@@ -46,7 +48,7 @@ export default class TeamEventAttendanceDao extends BaseDao<
   ): Promise<TeamEventAttendance> {
     const teamEventAttendanceWithUUID = {
       ...teamEventAttendance,
-      pending: true,
```
andrew032011, line 49: See my comment regarding keeping `pending`.

backend/src/types/DataTypes.d.ts
```diff
@@ -37,7 +37,9 @@ export type DBTeamEventAttendance = {
   hoursAttended?: number;
   image: string;
   eventUuid: string;
-  pending: boolean;
+  pending?: boolean;
```
andrew032011, line 40: See my comment regarding keeping `pending`.

common-types/index.d.ts
```diff
@@ -80,12 +80,16 @@ interface SignInForm {
...
 }
 
+type Status = 'pending' | 'approved' | 'rejected';
+
 interface TeamEventAttendance {
   member: IdolMember;
   hoursAttended?: number;
   image: string;
   readonly eventUuid: string;
-  readonly pending: boolean;
```
andrew032011, line 88: See my comment regarding keeping `pending`.

backend/src/types/DataTypes.d.ts
```diff
@@ -37,7 +37,9 @@ export type DBTeamEventAttendance = {
...
+  status: string;
```
andrew032011, line 41: Echoing my other comment, change this to `Status` type instead of `string`.

PR #524, opened 2023-10-12
Setup New Tailwind Website

andrew032011:
## Summary ##
Create new React/Next/TailwindCSS app with ShadCN support and added "new-dti-website" as a yarn workspace.

Just like in the IDOL repo:

`yarn run dev` to start the development server
`yarn run format` to format

A few noteworthy files: configure Tailwind in `tailwind.config.ts` and ShadCN in `components.json`.

## Notes ##
- There were typechecking issues with the new repo that stemmed from "typescript" being outdated. I upgraded typescript for the entire repo from ^4.4.3 to ^5.2.2
- Also upgraded a few other dependencies to their latest versions to prevent typechecking/build issues
- Upgraded backend "firebase-admin" dependency


JasonMun7 (approved):
This looks great! Thanks for adding that Shadcn button as a little demo and the extended color variations! I think maybe extending more screen breakpoints, especially mobile (sm) would be great but a very minor detail we can always add on later! Also we could add the @tailwindcss/aspect-ratio plugin to help with the aspect ratio of some of our elements. Other than that, this is a great starting point!

PR #526, opened 2023-10-21
Allow admin to edit dev portfolios

patriciaahuang:
### Summary

Notion ticket: https://www.notion.so/DP-Allow-Admins-to-Edit-Assignments-2307db13393c4f358e377e75f8bd6fea?pvs=4

Admins can now edit a dev portfolio without having to delete and replace an assignment with a new version (which was not feasible once submissions have been made). 

Frontend changes include adding an edit button for each portfolio that opens an edit modal, a success message if an edit is made, and error messages if conditions are violated. 

Backend changes include adding a new update endpoint, and documenting the Dev Portfolio API.

### Test Plan

I tested if a dev portfolio is successfully edited by looking at the firebase database, and checking that it updated correctly.

Each portfolio card now has an edit button, which opens a form to edit the portfolio.
Screen Shot 2023-10-21 at 4 24 22 PM

The form has pre-populated values depending on the original portfolio.
Screen Shot 2023-10-21 at 4 24 49 PM

If a portfolio is successfully edited, a success message will appear and the original page will automatically rerender in the background to show the updated portfolio.
Screen Shot 2023-10-21 at 4 27 14 PM

An error message will appear if certain conditions are violated.
Screen Shot 2023-10-21 at 4 28 33 PM

### DIFF
--- a/backend/tests/DevPortfolioAPI.test.ts
+++ b/backend/tests/DevPortfolioAPI.test.ts
@@ -46,7 +46,7 @@ describe('User is not lead or admin', () => {
     );
   });
 
-  test('deleteDevPortfolio shoudl throw permission error', async () => {
+  test('deleteDevPortfolio should throw permission error', async () => {
     await expect(deleteDevPortfolio('fake-uuid', user)).rejects.toThrow(
       new PermissionError(
         `User with email: ${user.email} does not have permission to delete dev portfolio!`
--- a/frontend/src/API/DevPortfolioAPI.ts
+++ b/frontend/src/API/DevPortfolioAPI.ts
@@ -23,7 +23,9 @@ export default class DevPortfolioAPI {
   }
 
   public static updateDevPortfolio(devPortfolio: DevPortfolio): Promise<DevPortfolio> {
-    return APIWrapper.put(`${backendURL}/dev-portfolio`, devPortfolio).then((res) => res.data.portfolio);
+    return APIWrapper.put(`${backendURL}/dev-portfolio`, devPortfolio).then(
+      (res) => res.data.portfolio
+    );
   }
 
   public static async makeDevPortfolioSubmission(
--- a/frontend/src/API/DevPortfolioAPI.ts
+++ b/frontend/src/API/DevPortfolioAPI.ts
@@ -22,6 +22,10 @@ export default class DevPortfolioAPI {
     );
   }
 
+  public static updateDevPortfolio(devPortfolio: DevPortfolio): Promise<DevPortfolio> {
+    return APIWrapper.put(`${backendURL}/dev-portfolio`, devPortfolio).then((res) => res.data.portfolio);
+  }
+
   public static async makeDevPortfolioSubmission(
     uuid: string,
     submission: DevPortfolioSubmission
--- a/frontend/src/components/Admin/DevPortfolio/AdminDevPortfolio.tsx
+++ b/frontend/src/components/Admin/DevPortfolio/AdminDevPortfolio.tsx
@@ -30,9 +30,9 @@ const AdminDevPortfolio: React.FC = () => {
   };
 
   useEffect(() => {
-    Emitters.devPortfolioUpdated.subscribe(() => fullReset());
+    Emitters.devPortfolioUpdated.subscribe(fullReset);
     return () => {
-      Emitters.devPortfolioUpdated.unsubscribe(() => fullReset());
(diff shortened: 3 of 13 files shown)


kevinmram (approved):
The new edit functionality significantly improves the users' interaction with the platform, streamlining the process of updating developer portfolios, rather than deleting them, saving time and reduce the possibility of errors or data inconsistencies!

backend/tests/DevPortfolioAPI.test.ts
```diff
@@ -46,7 +46,7 @@ describe('User is not lead or admin', () => {
     );
   });
 
-  test('deleteDevPortfolio shoudl throw permission error', async () => {
+  test('deleteDevPortfolio should throw permission error', async () => {
```
andrew032011, line 49: 👍

frontend/src/components/Admin/DevPortfolio/EditDevPortfolio.tsx
```diff
@@ -0,0 +1,52 @@
...
+    >
+      <Modal.Header>Edit Dev Portfolio</Modal.Header>
+      <Modal.Content>
+        <AdminDevPortfolioForm
+          formType={'edit'}
+          devPortfolio={devPortfolio}
+          editDevPortfolio={editDevPortfolio}
+          setOpen={setOpen}
+          setDevPortfolios={setDevPortfolios}
+        ></AdminDevPortfolioForm>
```
andrew032011, line 47: nit:
```
<AdminDevPortfolioForm>
</AdminDevPortfolioForm>
```
can be rewritten as
```
<AdminDevPortfolioForm/>
```

frontend/src/API/DevPortfolioAPI.ts
oscarwang20, line 1: We don't need `DevPortfolioResponseObj` since the error handling is already done by the API wrapper. Since `updateDevPortfolio` uses `PUT`, it should return the updated resource, so it should be `res.data.portfolio` like the other endpoints that modify a devPortfolio instance.

oscarwang20 (changes requested):

frontend/src/components/Admin/DevPortfolio/AdminDevPortfolio.tsx
```diff
@@ -37,10 +24,35 @@ const AdminDevPortfolio: React.FC = () => {
...
+  const fullReset = () => {
+    setIsLoading(true);
+    setDevPortfolios([]);
+  };
+
+  useEffect(() => {
+    const cb = () => {
+      fullReset();
+    };
+    Emitters.devPortfolioUpdated.subscribe(cb);
```
andrew032011, line 36: I think you can just pass `fullReset` directly in here. No need for the extra callback function.

frontend/src/API/DevPortfolioAPI.ts
andrew032011: To add on, it's excellent that you're doing error handling here, but we generally don't want to expose these error messages directly to the user on the UI. For now, since the only error we could really get is Internal Server Error or Permissions Error we don't really need to pass this info to the UI. 

I left a comment in `updateDevPortfolio` for how I would handle the error.

frontend/src/components/Admin/DevPortfolio/EditDevPortfolio.tsx
```diff
@@ -0,0 +1,52 @@
...
+const EditDevPortfolio = (props: {
+  devPortfolio: DevPortfolio;
+  setDevPortfolios: React.Dispatch<React.SetStateAction<DevPortfolio[]>>;
+}): JSX.Element => {
+  const { devPortfolio, setDevPortfolios } = props;
+  const [open, setOpen] = useState(false);
+
+  const editDevPortfolio = (portfolio: DevPortfolio) => {
+    DevPortfolioAPI.updateDevPortfolio(portfolio).then((val) => {
+      if (val.error) {
```
andrew032011, line 17: `val === undefined` is a sign that the DevPortfolio update did not go through and some error was thrown. As developers, we can inspect the network tab or look at the backend logs to see further what went wrong here. No need to expose the error message itself on the UI.

andrew032011 (commented):
This is really excellent work. You're picking things up fast and your prior backend experience shows!

See my/Oscar's comments about the UI. They are minor, just some style considerations.

andrew032011 (approved):

oscarwang20 (approved):

PR #532, opened 2023-10-23
Configure Backup DB

andrew032011:
### Summary 
Add script and corresponding and GitHub Actions cron job for backing up the Dev DB.

I'm deliberately not changing the `.env.example` file, since we should not be working with the backup DB ever, except for backing up the data periodically, which will be done through the job.

Factored these methods to be shared between `update-dev-db` and `backup-prod-db` scripts.

### Notion/Figma Link 

https://www.notion.so/Add-Firebase-backup-DB-9962ea595ead417bb40bd53fad94388a?pvs=4

### Test Plan 

Ran cron job, and the data matches the prod DB.

### DIFF
--- a/.github/workflows/backup-prod-db.yml
+++ b/.github/workflows/backup-prod-db.yml
@@ -1,0 +1,32 @@
+name: Back Up Prod Database
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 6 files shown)


oscarwang20 (approved):
Looks great, thanks for adding this!

PR #535, opened 2023-11-02
Fix `sendMail` API endpoint

andrew032011:
### Summary 

- Add permissions check in `sendMail` API endpoint to restrict to admins only
- Remove extraneous/outdated `isProd` check. The corresponding environment variable is not called `isProd` anymore.
- Remove `sendMemberUpdateNotifications` call for every Member information update. This endpoint was previously "dead code" anyway, and it doesn't make sense to keep (i.e. notify every admin every time someone updates a profile)
-  Replace usage of `EMAIL` env variable. This environment variable is not used in the codebase anymore. We use exclusively "dti.idol.github.bot@gmail.com" in our codebase.

### Test Plan 
Running this CURL command:
```
curl -X POST "localhost:9000/.netlify/functions/api/sendMail" -d '{"to": "<your email>", "subject": "test", "text": "Test test test"}' -H "auth-token:  <auth-token>" -H 'Content-Type: application/json'
```

### DIFF
--- a/backend/src/API/mailAPI.ts
+++ b/backend/src/API/mailAPI.ts
@@ -3,10 +3,20 @@ import { Request } from 'express';
 import getEmailTransporter from '../nodemailer';
 import { isProd } from '../api';
 import AdminsDao from '../dao/AdminsDao';
+import PermissionsManager from '../utils/permissionsManager';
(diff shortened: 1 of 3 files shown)


patriciaahuang (approved):
Looks good! Thanks, Andrew!

PR #541, opened 2023-11-09
Prevent mail sends in Non-Prod Environments

andrew032011:
### Summary 

Prevent mail from being sent in non-prod environments.

### Test Plan 

Verified:
Screenshot 2023-11-09 at 1 01 22 PM


patriciaahuang (approved):
Looks good!

PR #542, opened 2023-11-10
Send staging emails to stdout

andrew032011:
Small change: set it so that only "prod" environments will allow for the sending of emails. This will prevent us from accidentally sending out emails in staging and dev environments to lots of members.


patriciaahuang (approved):
This looks good to me! This will be very helpful when trying to debug email features in IDOL like the TEC reminder!

PR #543, opened 2023-11-11
Add documentation for team events image api

alyssayzhang:
### Summary 

- adds documentation to the methods in the` teamEventsImageAPI.ts` file


andrew032011 (approved):
Some nits, also

Maybe instead of "an IdolMember object", we can say "the member who made the request"

PR #551, opened 2023-11-13
Adjust Shoutouts Ending Time to Local Time Zone

andrew032011:
For the latest shoutout date, adjust the time to be 11:59pm on the same day, now adjusted to the current local time zone.


kevinmram (approved):
Everything's looking good to me- that's why you're the GOAT!

PR #552, opened 2023-11-14
Replace TeamEventsAPI references in MailAPI with Calls to Dao Instead

andrew032011:
### Summary 

Previously, we had TeamEventAPI references in `mailAPI`, which may reduce some code duplication, however, there are some permissions/authorization checks we do at the API level that we don't want to check "twice." 

In this case, using `TeamEventsAPI` introduced a bug that led to us not able to send reminder TEC emails to non-IDOL Admins/non-leads. 

Instead, reference `TeamEventAttendanceDao` and `TeamEventsDao` directly.

Screenshot 2023-11-14 at 4 12 52 PM

### Test Plan 

Sent the rest of the emails to all non-admins successfully!


patriciaahuang (approved):
Looks good!

PR #555, opened 2023-11-27
Improve error handling for edit users tool (require role to be selected, valid Cornell email)

andrew032011:
### Summary 
Add error alert if role is not selected and if valid Cornell email is not provided (or field is left blank)

### Test Plan 

Manual (see video attached).

https://github.com/cornell-dti/idol/assets/59291082/2a1493d5-3ed8-40b3-bfe0-f99c2c1dc3dc

### DIFF
--- a/frontend/src/components/Admin/AddUser/AddUser.tsx
+++ b/frontend/src/components/Admin/AddUser/AddUser.tsx
@@ -124,26 +124,26 @@ export default function AddUser(): JSX.Element {
   }
 
   async function setUser(member?: CurrentSelectedMember): Promise<void> {
-    if(!member) return;
+    if (!member) return;
 
     // Get role description (error-safe)
     let roleDescription;
     try {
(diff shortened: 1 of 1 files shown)


andrew032011:
> This looks great! Do you think we could write a helper function to validate each user's fields? It could reduce code duplication if we enforce the same rules for editing a single user as we do for batch CSV uploads.

Good point, I did not think about the other error handling we're doing for the CSV's, which is arguably much more comprehensive than I am doing here.

The format we're doing things in is a tiny bit different between the two, but I can think something along the lines of creating a `validateMember` function that returns an array of strings with all the issues from a given user might be a nice way to abstract out the logic.

oscarwang20 (approved):

andrew032011:
@oscarwang20 I'm going to modify the logic to just check to make sure we have a valid email inputed and role is selected in the dropdown.

PR #559, opened 2023-11-29
Clear TEC and Dev Portfolio Forms Upon Submission

andrew032011:
### Summary 

Context:

Just from taking a few passes on submissions this semester, I’ve noticed a lot of duplicate submissions for the TEC form and Dev Portfolios. This is most likely because for both the [TEC form](https://idol.cornelldti.org/forms/teamEventCredits) and [Dev Portfolio form](https://idol.cornelldti.org/forms/devPortfolio), when you submit, the form fields do not clear. So users might be confused as to whether or not their submissions went through, so they double/triple click and then we get a bunch of duplicate submissions.

This is by design, in case you wanted to make multiple submissions, or for Dev Portfolio, it was in case you wanted to replace your existing submission with a minor adjustment (so you don’t have to fill everything in the form again).
But I haven’t personally used this feature. It feels more like a bug, and I think we’re getting a lot of unintentional duplicate TEC/Dev Portfolio submissions from it.

Taking a poll in #idol-devs, survey says Dev's don't use it either.

### Test Plan 

https://github.com/cornell-dti/idol/assets/59291082/3a3edb3f-afa9-4436-b598-33c25a117ffe

### Breaking Changes

### DIFF
--- a/frontend/src/components/Forms/DevPortfolioForm/DevPortfolioForm.tsx
+++ b/frontend/src/components/Forms/DevPortfolioForm/DevPortfolioForm.tsx
@@ -47,8 +47,10 @@ const DevPortfolioForm: React.FC = () => {
           setDevPortfolio(undefined);
           setOpenPRs(['']);
           setReviewedPRs(['']);
-          setText(undefined);
+          setOtherPRs(['']);
+          setText('');
           setDocumentationText('');
+          setOpenOther(false);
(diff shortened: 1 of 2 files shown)


kevinmram (approved):
This adjustment will enhance user experience by reducing confusion and streamlining the submission process, making it more efficient and intuitive- that's awesome! 🐐

PR #560, opened 2023-11-29
Check for otherPRs instead of Member Role for Admin Dev Portfolio Display

andrew032011:
### Summary 

We avoid checking the role since otherPRs applies to Dev's only. If there are other PR's submitted, then the submission is a Dev Portfolio submission, not a TPM portfolio. 

Consider the case where a member is promoted to TPM mid-semester. That should not change the way this Dev Portfolio is displayed. 

Consider also the case where a member switches from TPM to Dev mid-semester. (Yes, both of these cases have happened at some point!). We want to preserve the state of the portfolio.


kevinmram (approved):
This approach is more flexible and resilient, accommodating scenarios where roles may shift. It's a small, but smart, enhancement to our system's logic, ensuring accuracy and adaptability in portfolio management. 🐐

PR #567, opened 2023-12-03
added documentation, CandidateDeciderAPI

kevinmram:
added documentation to CandidateDeciderAPI- documenting the purpose and usage of each method within the class.

### DIFF
--- a/backend/src/API/candidateDeciderAPI.ts
+++ b/backend/src/API/candidateDeciderAPI.ts
@@ -4,6 +4,11 @@ import PermissionsManager from '../utils/permissionsManager';
 
 const candidateDeciderDao = new CandidateDeciderDao();
 
+/**
+ * Retrieves all Candidate Decider instances, filtering based on the user's role and permissions.
+ * @param {IdolMember} user - The user requesting the instances.
+ * @returns {Promise<CandidateDeciderInfo[]>} - A promise that resolves to an array of CandidateDeciderInfo objects.
+ */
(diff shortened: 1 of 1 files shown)


backend/src/API/candidateDeciderAPI.ts
```diff
@@ -89,6 +124,17 @@ export const getCandidateDeciderInstance = async (
...
 };
 
+/**
+ * Updates the rating and comment for a Candidate Decider instance.
+ * @param {IdolMember} user - The user updating the rating and comment.
+ * @param {string} uuid - The unique identifier of the instance.
+ * @param {number} id - The identifier of the specific candidate.
+ * @param {Rating} rating - The new rating.
+ * @param {string} comment - The new comment.
+ * @returns {Promise<void>} - A promise that resolves to void.
```
andrew032011, line 134: I don't think you put this in the docstrings for the other methods in this file that have `Promise<void>` return type. Maybe remove it?

backend/src/API/candidateDeciderAPI.ts
```diff
@@ -4,6 +4,11 @@ import PermissionsManager from '../utils/permissionsManager';
 
 const candidateDeciderDao = new CandidateDeciderDao();
 
+/**
+ * Retrieves all Candidate Decider instances, filtering based on the user's role and permissions.
```
henry-li-06, line 8: Only retrieves the metadata and not the data itself. Also doesn't hurt to mention in more detail how the permission work.

PR #569, opened 2023-12-03
Break Word in Documentation Cell in Dev Portfolio Table

andrew032011:
### Summary 

We have a UI bug in the Dev Portfolio Details table due to longer documentation responses that are not being properly text-wrapped, so much so that some columns are not even visible unless you zoom out on a standard browser viewport.

Set max width of documentation column to be 200px and wrap the text.

### Test Plan 

Before:
Screenshot 2023-12-03 at 2 29 39 PM

After:
Screenshot 2023-12-03 at 2 29 16 PM


kevinmram (approved):
LGTM + GOAT

PR #575, opened 2024-02-09
Remove pending field dependency for TEC

patriciaahuang:
### Summary

TEC's used to have both a "pending" and "status" field, which was redundant. The backend and frontend are updated to not use/depend on the pending field, and the pending field is made optional in data types. A script was run to remove the pending field from all Team Event Attendance collections in the dev database.

### Notion/Figma Link

https://www.notion.so/TEC-Remove-dependency-on-Pending-field-8d7d76c75e2a46758da27620e890573c?pvs=4

### Test Plan

To ensure removing the pending field in the dev database and updating the backend and frontend didn't mess with any functionalities, I tested each function that was modified locally (creating a team event attendance, submitting a team event attendance, etc.) and everything still works the same (for new and old data).


backend/src/types/DataTypes.d.ts
```diff
@@ -37,7 +37,7 @@ export type DBTeamEventAttendance = {
   hoursAttended?: number;
   image: string;
   eventUuid: string;
-  pending: boolean;
+  pending?: boolean;
```
andrew032011, line 40: I think we'll be ok to remove this field entirely, since we're not using it. Note that even if the data in the database has this field, you won't have a "type error" due to extra fields that the type doesn't have.

common-types/index.d.ts
```diff
@@ -87,7 +87,7 @@ interface TeamEventAttendance {
   hoursAttended?: number;
   image: string;
   readonly eventUuid: string;
-  readonly pending: boolean;
+  readonly pending?: boolean;
```
andrew032011, line 90: Same here as well, feel free to remove this field entirely, since we've fully migrated over to the `status` field.

andrew032011 (commented):
Are there plans to run a script to remove all `pending` fields? Super low priority and the risk probably outweighs the reward, so not really necessary. 

We can leave it and then starting Fall 2024 it will be officially retired when we clear the data.

andrew032011 (approved):
Great work! Good to see us killing off the field and cleaning up the status pending/approved/rejected work from last sem. Please see my comments above.

andrew032011 (approved):

alyssayzhang (approved):
this looks good to me 🙌!! nice work on testing this ticket and making sure that everything works the same :)

PR #577, opened 2024-02-12
Adds documentation to TeamEventAttendanceDao.ts

alyssayzhang:
### Summary 

This PR finishes up documenting all the classes and functions in `TeamEventAttendanceDao.ts`

### DIFF
--- a/backend/src/dao/TeamEventAttendanceDao.ts
+++ b/backend/src/dao/TeamEventAttendanceDao.ts
@@ -5,6 +5,9 @@ import { deleteCollection } from '../utils/firebase-utils';
 import { getMemberFromDocumentReference } from '../utils/memberUtil';
 import BaseDao from './BaseDao';
 
+/**
+ * Materializes team event attendance object from the serialized form in the database
+ */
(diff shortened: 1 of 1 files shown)


kevinmram (approved):
LGTM!

oscarwang20 (changes requested):
I would advise against doing this just for the documentation requirement. The names are pretty self explanatory here.

andrew032011 (changes requested):
I would have to agree with Oscar. Those are inherited functions so it's not telling too much than what BaseDao mentions. I think we could keep the docstring for the class, but also it does not add more info.

andrew032011:
@alyssayzhang Doing some cleanup here. Are we good to close this PR for the time being? If we want to add documentation to this file later on that's fine too. Just don't want this PR to get stale.

alyssayzhang:
> @alyssayzhang Doing some cleanup here. Are we good to close this PR for the time being? If we want to add documentation to this file later on that's fine too. Just don't want this PR to get stale.

Yeah I'll close it!

PR #578, opened 2024-02-12
[IDOL, shoutouts] - edit shoutouts

kevinmram:
added a button and functions for editing shoutouts , allowing users to edit their shoutouts after submission.

- [x] implemented the (Edit) button.
- [x] implemented the (editShoutout, updateShoutout) functions.
- [x] display the updated shoutout- without reversion after a refresh.

https://github.com/user-attachments/assets/3caa8374-84cd-4cb9-ba51-387a8b4385fe


frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
```diff
@@ -1,30 +1,86 @@
...
 import ShoutoutDeleteModal from '../../Modals/ShoutoutDeleteModal';
+import { Emitters } from '../../../utils';
 import styles from './ShoutoutCard.module.css';
 
-const ShoutoutCard = (props: {
-  shoutout: Shoutout;
-  setGivenShoutouts: React.Dispatch<React.SetStateAction<Shoutout[]>>;
-}): JSX.Element => {
-  const { shoutout, setGivenShoutouts } = props;
+const ShoutoutCard = ({ shoutout, setGivenShoutouts }) => {
```
patriciaahuang, line 8: Make sure to give a type to `shoutout` and `setGivenShoutouts` (like the original code) to pass the type check.

patriciaahuang (commented):
Good job! I left a note about how to pass the last check. Also, it might be helpful to link the Notion ticket and an image for what you see on your end (edit button, success/error messages, etc), so reviewers can quickly check that we're see the same thing.

backend/src/dao/ShoutoutsDao.ts
```diff
@@ -76,6 +76,20 @@ export default class ShoutoutsDao extends BaseDao<Shoutout, DBShoutout> {
...
+  /**
+   * Edits a shoutout
+   * @param uuid - uuid of the shoutout
+   * @param shoutout - shoutout object
+   */
+  async editShoutout(uuid: string, newMessage: string): Promise<Shoutout> {
+    await this.collection.doc(uuid).update({ message: newMessage });
+    const updatedDoc = await this.getDocument(uuid);
+    if (!updatedDoc) {
+      throw new Error('Failed to fetch updated shoutout...');
```
andrew032011, line 88: Remove the `...` in the string?

andrew032011: Also, could you use `updateShoutout` instead of creating a method just to edit the message?

backend/src/API/shoutoutAPI.ts
```diff
@@ -64,6 +64,18 @@ export const hideShoutout = async (
...
+export const editShoutout = async (
+  uuid: string,
+  newMessage: string,
+  user: IdolMember
+): Promise<Shoutout> => {
+  const shoutout = await shoutoutsDao.getShoutout(uuid);
+  if (!shoutout) {
+    throw new NotFoundError(`Shoutout with uuid: ${uuid} does not exist!`);
+  }
+  return shoutoutsDao.editShoutout(uuid, newMessage);
```
andrew032011, line 76: see if you can use `updateShoutout` here instead

frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
```diff
@@ -1,30 +1,80 @@
...
-  const fromString = shoutout.isAnon
-    ? 'From: Anonymous'
-    : `From: ${shoutout.giver?.firstName} ${shoutout.giver?.lastName}`;
-  const dateString = `${new Date(shoutout.timestamp).toDateString()}`;
+const ShoutoutCard: React.FC<ShoutoutCardProps> = ({ shoutout, setGivenShoutouts }) => {
+  const [isEditing, setIsEditing] = useState(false);
+  const [editedMessage, setEditedMessage] = useState(shoutout.message);
+
+  const handleEditShoutout = async () => {
+    try {
```
andrew032011, line 17: I think we can do without this `try/catch` and console statement, since it doesn't do much other than `console.log` the error (which we would get in the console anyways). It would be worth it to try/catch if we had some sort of fallback code we wanted to run or wanted to record metrics.

backend/src/API/shoutoutAPI.ts
```diff
@@ -64,6 +64,18 @@ export const hideShoutout = async (
   await shoutoutsDao.updateShoutout({ ...shoutout, hidden: hide });
 };
 
+export const editShoutout = async (
+  uuid: string,
+  newMessage: string,
+  user: IdolMember
+): Promise<Shoutout> => {
+  const shoutout = await shoutoutsDao.getShoutout(uuid);
```
andrew032011, line 72: Can you add permissions checking here to make sure that the shoutout belongs to the user who's trying to edit it OR it's a leadOrAdmin?

andrew032011:
Screenshot 2024-09-24 at 6 26 30 PM
Can you make it so that the spacing around the icon doesn't change as a result of adding this icon? The layout looks a little funky now since the icon is larger than the trash icon.

backend/src/dao/ShoutoutsDao.ts
patriciaahuang: This is more for future reference, but I think checking if the document exists in the DAO layer might be repetitive, since we already check if the uuid exists in the API layer. Not including the checks here also helps keep the DAO files consistent, since most of them do not have additional checks.

patriciaahuang (commented):
Overall functionality looks good! I agree with what Andrew said about using updateShoutouts instead of creating a new function. Also left a comment to help keep the DAO layer consistent.

backend/src/dao/ShoutoutsDao.ts
andrew032011: woah thanks for catching this @patriciaahuang, not sure how I didn't see it. We should not do any permissions checks at the DAO layer. That's reserved for the API layer!

frontend/src/API/ShoutoutsAPI.ts
```diff
@@ -48,6 +48,13 @@ export default class ShoutoutsAPI {
     return APIWrapper.put(`${backendURL}/shoutout`, { uuid, hide }).then((res) => res.data);
   }
 
+  public static updateShoutout(
+    uuid: string,
+    shoutout: Partial<Shoutout>
```
andrew032011, line 53: Why is this `Partial` type?

cchrischen (approved):
almost the one year anniversary of this PR!! once you fix this merge conflict, it's good to go

PR #579, opened 2024-02-12
added documentation, shoutoutAPI

kevinmram:
added documentation to shoutoutAPI- documenting the purpose and usage of each method within the class.


oscarwang20 (changes requested):

andrew032011 (changes requested):
I find it a bit redundant to mention permissions in the docstring. I think in general it is implied. If we want to include info about permissions here, we should be more specific as to what we're checking. Otherwise, I might remove it (you can keep the @throws).

Also soon we'll have an auth model that takes care of some of this logic.

andrew032011 (approved):

oscarwang20 (approved):

PR #584, opened 2024-02-21
update TEC email notifs to only send weekly on mondays

oscarwang20:
### Summary 

Kevin requested to receive TEC notifications less frequently. 

- Maybe we could allow C&I lead to change the frequency of these emails themselves?
- Could potentially look into for Slack integrations as well.


andrew032011 (approved):

andrew032011:
> Maybe we could allow C&I lead to change the frequency of these emails themselves?

Let's discuss offline the implications here!

PR #585, opened 2024-02-25
Add support for initiative TEC

cchrischen:
### Summary 
- Factor out `INITIATIVE_EVENTS` from several files relating to TEC into `consts.ts`
- Add a line for displaying remaining initiative TEC needed
- Add label for initiative events when submitting TEC

### Test Plan 
- Previewing each page 
image
image

### Notes 
- The codebase still uses community instead of the initiative (official terminology) in several files, mainly in the API and backend files, which can be really confusing.

### DIFF
--- a/frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
+++ b/frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
@@ -126,9 +126,7 @@ const TeamEventCreditForm: React.FC = () => {
                       <div className={styles.flex_space_center}>
                         <div className={styles.flex_start}>{event.name}</div>
                         <div className={styles.flex_end}>
-                          {INITIATIVE_EVENTS && event.isCommunity && (
-                            <Label content="initiative"/>
-                          )}
+                          {INITIATIVE_EVENTS && event.isCommunity && <Label content="initiative" />}
                           <Label
                             content={`${new Date(event.date).toLocaleDateString('en-us', {
                               month: 'short',
--- a/frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
+++ b/frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
@@ -6,6 +6,7 @@ import { TeamEventsAPI } from '../../../API/TeamEventsAPI';
 import TeamEventCreditDashboard from './TeamEventsCreditDashboard';
 import styles from './TeamEventCreditsForm.module.css';
 import ImagesAPI from '../../../API/ImagesAPI';
+import { INITIATIVE_EVENTS } from '../../../consts';
 
 const TeamEventCreditForm: React.FC = () => {
   // When the user is logged in, `useSelf` always return non-null data.
@@ -125,6 +126,9 @@ const TeamEventCreditForm: React.FC = () => {
                       <div className={styles.flex_space_center}>
                         <div className={styles.flex_start}>{event.name}</div>
                         <div className={styles.flex_end}>
+                          {INITIATIVE_EVENTS && event.isCommunity && (
+                            <Label content="initiative"></Label>
+                          )}
                           <Label
                             content={`${new Date(event.date).toLocaleDateString('en-us', {
                               month: 'short',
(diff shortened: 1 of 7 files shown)


frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
```diff
@@ -125,6 +126,9 @@ const TeamEventCreditForm: React.FC = () => {
                       <div className={styles.flex_space_center}>
                         <div className={styles.flex_start}>{event.name}</div>
                         <div className={styles.flex_end}>
+                          {INITIATIVE_EVENTS && event.isCommunity && (
+                            <Label content="initiative"></Label>
```
andrew032011, line 130: nit: can rewrite as `<Label content="initiative"/>`

andrew032011 (changes requested):
> The codebase still uses community instead of the initiative (official terminology) in several files, mainly in the API and backend files, which can be really confusing.

Agreed it is confusing. Can we get rid of all references to community on the frontend as a first complete step? This is already some great work, but I still see references in `frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx` and some other files. If it's a type issue, like `teamEvent.isCommunity`, you can leave it, but everywhere else I'd get rid of it so that other people working in the codebase don't get confused.

Since changing the type involves a schema change, I'd make a followup PR to change that field from `isCommunity` to `isInitativeEvent` (or something like that) so you derisk breaking anything on the backend from this change alone. And after that follow up PR is merged in, we should do a DB migration of existing TeamEvents (we can start a Slack thread with you, me, Oscar to discuss this since it's your first time).

But other than the types, everything else should change now.

henry-li-06:
> Agreed it is confusing. Can we get rid of all references to community on the frontend as a first complete step? This is already some great work, but I still see references in `frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx` and some other files. If it's a type issue, like `teamEvent.isCommunity`, you can leave it, but everywhere else I'd get rid of it so that other people working in the codebase don't get confused.

If this is a feature release then it's probably best to not make significant changes other than changing the feature flag and adding the extra label, new option, etc. If you really want to migrate off of referencing `community` (it's probably a bit confusing to leave it even tho it's effectively the same thing), the more "proper" way would be:
1. Run a db script to add the new field. I'd assume you'd want it to be like `eventType` or something like that so that it's extendable in the future to more than just initiatives. This script _would not_ remove the `isCommunity` field. You want to leave this field in for the time being so that it's backwards compatible with releases. 
2. Update the BE so that `isCommunity` and `eventType` are both updated together whenever either of them are updated.
3. Update the FE to fully reference `eventType` instead of `isCommunity`.
4. Update the BE to stop sending the `eventType` in the API response. This _does not_ mean removing the field from the db yet. 
5. Finally remove the `isCommunity` field from all the docs in the db. 

At every step you should of course verify that everything still works as expected. And also if a step that is a code change wasn't done correctly, only the code changes needs to be reverted while the db is still backwards compatible with the previous release. 

It's also better to do this in staging of course but that might be a lost cause with the state of the staging environment 🙃

andrew032011:
> > Agreed it is confusing. Can we get rid of all references to community on the frontend as a first complete step? This is already some great work, but I still see references in `frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx` and some other files. If it's a type issue, like `teamEvent.isCommunity`, you can leave it, but everywhere else I'd get rid of it so that other people working in the codebase don't get confused.
> 
> If this is a feature release then it's probably best to not make significant changes other than changing the feature flag and adding the extra label, new option, etc. If you really want to migrate off of referencing `community` (it's probably a bit confusing to leave it even tho it's effectively the same thing), the more "proper" way would be:
> 
> 1. Run a db script to add the new field. I'd assume you'd want it to be like `eventType` or something like that so that it's extendable in the future to more than just initiatives. This script _would not_ remove the `isCommunity` field. You want to leave this field in for the time being so that it's backwards compatible with releases.
> 2. Update the BE so that `isCommunity` and `eventType` are both updated together whenever either of them are updated.
> 3. Update the FE to fully reference `eventType` instead of `isCommunity`.
> 4. Update the BE to stop sending the `eventType` in the API response. This _does not_ mean removing the field from the db yet.
> 5. Finally remove the `isCommunity` field from all the docs in the db.
> 
> At every step you should of course verify that everything still works as expected. And also if a step that is a code change wasn't done correctly, only the code changes needs to be reverted while the db is still backwards compatible with the previous release.
> 
> It's also better to do this in staging of course but that might be a lost cause with the state of the staging environment 🙃

Oh yeah we're talking about a similar way of doing this migration (on Slack). Thinking about it more I'm ok with us just merging this in the way it is. @cchrischen let's just follow this up with a clean-up PR to get the rest of the changes!

andrew032011 (approved):

oscarwang20 (approved):

PR #589, opened 2024-03-04
Add documentation to Members DAO

patriciaahuang:
Added documentation to `getAllTeams` and `getTeam` in `MembersDao.ts`. It wasn't immediately clear why these functions are in the Members DAO, so there is now a short clarification (based on the explanation of the `Team` data type in #581).

### DIFF
--- a/backend/src/dao/MembersDao.ts
+++ b/backend/src/dao/MembersDao.ts
@@ -97,7 +97,7 @@ export default class MembersDao extends BaseDao<IdolMember, IdolMember> {
   }
 
   /**
-   * Gets all teams based on the subteam and formersubteam(s) each member is on, since teams do not have 
+   * Gets all teams based on the subteam and formersubteam(s) each member is on, since teams do not have
    * a collection in the database (teams are aggregated data from the members collection).
    * @returns A promise that resolves to an array of all Team objects.
    */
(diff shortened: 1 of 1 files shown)


backend/src/dao/MembersDao.ts
```diff
@@ -125,6 +130,11 @@ export default class MembersDao extends BaseDao<IdolMember, IdolMember> {
     return Array.from(teamsMap.values());
   }
 
+  /**
+   * Gets a specific team by searching through each IDOL member and saving each member of the team.
+   * @param id The name of the team
+   * @returns A promise that resolves to the Team object with the specified name.
```
oscarwang20, line 136: This method could also return null if the team has no members. Can you specify that here as well?

backend/src/dao/MembersDao.ts
```diff
@@ -96,6 +96,11 @@ export default class MembersDao extends BaseDao<IdolMember, IdolMember> {
     await batch.commit();
   }
 
+  /**
+   * Gets all teams based on the subteam and formersubteam(s) each member is on, since teams do not have
+   * a collection in the database (teams are aggregated data from the members collection).
+   * @returns A promise that resolves to an array of all Team objects.
```
oscarwang20, line 102: Saying "all Team objects" could be a little confusing here because it implies theres a set of team objects somewhere out there when we're really just constructing these teams from scratch.

oscarwang20 (changes requested):

oscarwang20 (approved):

PR #590, opened 2024-03-04
added documentation, CandidateDeciderAPI

kevinmram:
added documentation to CandidateDeciderAPI- documenting the purpose and usage of each method within the class.
- (didn't include @param to avoid redundancy)

### DIFF
--- a/backend/src/API/candidateDeciderAPI.ts
+++ b/backend/src/API/candidateDeciderAPI.ts
@@ -37,7 +37,6 @@ export const getAllCandidateDeciderInstances = async (
  * Creates a new CandidateDecider instance in the database.
  * This method checks if the user has admin permissions before allowing the creation of the instance.
  * @returns {Promise<CandidateDeciderInfo>} A promise that resolves with the created CandidateDeciderInfo object.
- * @throws {PermissionError} If the user does not have permission to create a new instance.
  */
 export const createNewCandidateDeciderInstance = async (
   instance: CandidateDeciderInstance,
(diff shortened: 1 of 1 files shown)


patriciaahuang (changes requested):
All the errors have specific messages on why it may be thrown within the body of the function, so all `@throws` are a bit redundant.

andrew032011:
@keverie Let's get this merged in (after comments addressed)!

backend/src/API/candidateDeciderAPI.ts
```diff
@@ -67,6 +88,11 @@ export const deleteCandidateDeciderInstance = async (
   await candidateDeciderDao.deleteInstance(uuid);
 };
 
+/**
+ * Retrieves a specific CandidateDecider instance from the database.
+ * This method checks if the user has permission to access the requested instance.
```
patriciaahuang, line 93: "Retrieves a specific CandidateDecider instance from the database, if it exists." Since you included that the method checks for permissions, briefly mentioning that it also checks if the instance exists will be helpful.

backend/src/API/candidateDeciderAPI.ts
```diff
@@ -83,6 +109,11 @@ export const getCandidateDeciderInstance = async (
   return instance;
 };
 
+/**
+ * Updates the rating and comment of a CandidateDecider instance in the database.
+ * This method ensures that only users with access permission can update the rating and comment.
```
patriciaahuang, line 114: Same here.

patriciaahuang (approved):
I left a small comment, but otherwise looks good!

PR #591, opened 2024-03-21
Add team page member display

cchrischen:
### Summary 
- Implemented the member display section of the team page using the fall 2023 archive json
- Add shadcn card component
- Add types file to use common-types

This pull request is the first step towards implementing the **team page**

- [x] implemented member display
- [x] add responsiveness

### Notion/Figma Link 
https://www.figma.com/file/jKbvR1jVr1k80GGWnCpiRT/Website-Designs-SP23%2FFA23?type=design&node-id=5464-9095&mode=design&t=5jZUX1RMV8eKUdRC-0

https://www.notion.so/Team-Page-Team-Member-Display-5d8845dc798f4145b541c4282c4a1417

### Test Plan 
see video!
https://github.com/cornell-dti/idol/assets/144409850/a4fe1d6a-0f0b-4dc1-b1cd-9fb3f81d7309

### Notes 
- martha.png is a placeholder


oscarwang20:
Thanks for all the hard work on this Chris! Looks awesome so far.

alyssayzhang (commented):
This looks really good 🙌 🙌!!  I'm wondering if we should display Technical PM as Technical Product Manager ?

new-dti-website/components/team/MemberGroup.tsx
```diff
@@ -0,0 +1,352 @@
...
+                    </a>
+                  )
+                );
+              })}
+            </div>
+          </div>
+          <div>
+            <button
+              onMouseEnter={() => setHover(true)}
+              onMouseLeave={() => setHover(false)}
```
henry-li-06, line 252: Ok so I've since learned that using inline arrow functions like this is kinda "frowned upon." It's supposedly less performant (although very marginally) because it needs to be redefined for each re-render but more importantly I think it's a little less readable (although also debatably marginally).  

It's theoretically better to have 
`const mouseHandler = () => setHover(prev => !prev)` and `onMouseEnter={mouseHandler}`

But feel free to ignore, this is just an FYI for the sake of sharing information.

new-dti-website/components/team/MemberGroup.tsx
```diff
@@ -0,0 +1,352 @@
...
+        <h2 className="font-semibold text-[32px]">{`${roleName} ${
+          roleName !== 'Leads' ? '' : 'Team'
+        }`}</h2>
+        <p className="mt-3 text-xl">{description}</p>
+        <div className="flex flex-row justify-between flex-wrap grid lg:grid-cols-4 md:grid-cols-3 gap-10 mt-10 mb-[100px]">
+          {members.map((member, index) => (
+            <>
+              <MemberCard
+                {...member}
+                image="martha.png"
```
henry-li-06, line 329: This is a fairly small thing but I'm guessing that `martha.png` was just pulled off Google since the image dimensions aren't very nicely rounded. It probably wouldn't hurt to at least crop the image to whatever the expected size of headshots should be according to the design. 

(I just say this because knowing how projects go, this will probably be forgotten about til the actual headshots get put in after this PR gets merged). 

This repo has an `image-cli` that should be able to crop (and also compress but that isn't particularly important rn) to the desired size. @oscarwang20 should be aware of it if not @andrew032011 definitely should be. 

Now if Martha would just fix her weather machine...

henry-li-06 (commented):
Wow this is exciting it looks nice :D 

Left some comments, some are very trivial feel free to ignore those

new-dti-website/components/team/MemberGroup.tsx
```diff
@@ -0,0 +1,352 @@
...
+        subteamName = 'IDOL';
+        break;
+      case 'cornellgo':
+        subteamName = 'Cornell GO';
+        break;
+      case 'leads':
+        subteamName = 'Lead';
+        break;
+      default:
+        subteamName = subteam.charAt(0).toUpperCase() + subteam.slice(1);
```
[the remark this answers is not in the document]
henry-li-06: Or maybe even just a data file tbh that maps the team id to the actual team data. And maybe don't want a default still. I think this should theoretically pull from the team API to make sure things stay up to date 🤔

new-dti-website/components/team/MemberDisplay.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const memberDetailsRef = useRef<HTMLInputElement>(null);
+
+  const allMembers = FA23Members.members as IdolMember[];
+  const roles: {
+    [key in Role]: {
+      roleName: string;
+      description: string;
+      members: IdolMember[];
+      order: string[];
```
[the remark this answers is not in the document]
cchrischen: Yeah, dividing the two member types would be good for privacy reasons. Members can then edit what they want to display on their profile via IDOL, but that's another feature we can talk about.

The team page has risen some concerns about the current database, which we should resolve first before deciding on using `NovaMember` (which I'm leaning towards). First, we need to keep track of the colleges of each member. Majors should be standardized (CS vs. Computer Science) -- these are important when determining the represented colleges and majors in the About portion of the page. Another issue is how to deal with multiple subroles within a main  role, akin to the advisor discussion on a different PR. That is, how to expand `RoleDescription` or some other field to include advisors, APM, PMM, different lead roles, etc.

cchrischen:
### Commit Changes
* Refactor several dictionaries and lists to `team.json`
* Refactor some utility functions to `utils.ts`
* Add Subteam type to common-types

henry-li-06:
> ### Commit Changes
> * Refactor several dictionaries and lists to `team.json`
> * Refactor some utility functions to `utils.ts`
> * Add Subteam type to common-types

Just for informative purposes, if you split up the single _refactoring and fixing nits_ commit into 3 commits with these exact commit messages, you wouldn't need this comment :D. 

That's kinda the purpose of using commits and having non-trivial commit messages in the first place. You can even selectively stage part of a file to be committed using `git add -p`. 

And besides that, it can also be helpful when reviewing because the reviewer can go commit by commit through fairly isolated changes.

new-dti-website/components/team/MemberDisplay.tsx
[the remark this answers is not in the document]
henry-li-06: w.r.t. `NovaMember` vs ` IdolMember`, that's really just about whether to use the same data and fields bewteen idol and the dti website. The `NovaMember` type can of course be changed to fit whatever the needs of the new dti website are. 

w.r.t. profile stuff:
- The subroles situation i kinda touched on in #588. Related to #471 as well. I think that's the best way to handle this. 
- In terms of standardizing majors, there's a tradeoff that needs to be considered there. The IDOL profile editor is intended to allow self-servicing of changing things like your major. That's just a free-form text box to type your major in. The only "verification" happens at the stage when an idol admin has to check the diff before approving it. I'm not sure at that stage if rejecting a change is worth it because someone put "CS" instead of "Computer Science." The way to prevent that might be to use a dropdown with all the majors, but that requires maintaining a list of majors which is a pain. From a design perspective, I'm not sure where they stand on that, but from an engineering perspective, it might be fine imo as long as it's not something egregious like the major is misspelled or it's a fake major or something.

new-dti-website/components/team/MemberGroup.tsx
```diff
@@ -101,71 +87,14 @@ type MemberDetailsProps = {
 
 export const MemberDetails: React.FC<MemberDetailsProps> = (props: MemberDetailsProps) => {
   const [hover, setHover] = useState<boolean>(false);
+  const mouseHandler = () => setHover(!hover);
```
henry-li-06, line 90:
```suggestion
  const mouseHandler = () => setHover(prev => !prev)
```
https://legacy.reactjs.org/docs/hooks-reference.html#functional-updates
https://ogzhanolguncu.com/blog/how-does-prevstate-works-under-the-hood

new-dti-website/components/team/MemberGroup.tsx
```diff
@@ -0,0 +1,326 @@
...
+  | 'queuemein'
+  | 'cuapts'
+  | 'idol'
+  | 'cornellgo'
+  | 'carriage'
+  | 'leads'
+  | 'business'
+  | 'zing'
+  | 'dac';
+
```
andrew032011, line 103: I'd remove this type. We can just set it to a string (Subteam is not a fixed type).

new-dti-website/components/team/MemberDisplay.tsx
```diff
@@ -0,0 +1,101 @@
+import { useRef, useState } from 'react';
+import MemberGroup from './MemberGroup';
+import Icon from '../icons';
+import FA23Members from '../../../backend/src/members-archive/fa23.json';
+import teamRoles from './data/roles.json';
+import roleIcons from './data/roleIcons.json';
+import { populateMembers } from '../../src/app/utils';
+
+type DisplayCategory = 'Full Team' | 'Leads' | 'Design' | 'Product' | 'Business' | 'Development';
```
andrew032011, line 9: Delete this type and use `string` instead, since we're reading these values directly from the JSON anyways. So we're not benefitting from any type-checking going on here.

new-dti-website/src/app/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+    [key in Role]: {
+      roleName: string;
+      description: string;
+      members: IdolMember[];
+      order: string[];
+    };
+  },
+  allMembers: IdolMember[]
+) => {
+  for (const [key, value] of Object.entries(roles)) {
```
andrew032011, line 12: Instead of the loop, try to do this more functionally with return ` Object.entries(roles).map(...)` instead of editing this in place, which makes it harder to follow (i.e. if this breaks, we'll have a harder time fixing it). Slack me if you need help!

This might involve changing more than just this function, but the refactor will be worth it for code readability.

image

new-dti-website/components/team/MemberGroup.tsx
```diff
@@ -0,0 +1,326 @@
...
+}) => {
+  const selectedMemberIndex: number = useMemo(
+    () => (selectedMember ? members.indexOf(selectedMember) : -1),
+    [members, selectedMember]
+  );
+
+  const { width } = useScreenSize();
+
+  const canInsertMemberDetails = (index: number): boolean => {
+    let columns = 4;
```
andrew032011, line 274: Can you leave a comment explaining what `columns` and `index` is and what's going on here in general as well (especially with the checks we're doing underneath)?

Also we shouldn't really be using magic numbers `768` and `1024`, and `2`, `3`, `4`. Instead, let's define consts so that these numbers are self-documenting (i.e. I know what each number means).

Nit: it might be make sense to use a switch/case here instead of if/else.

andrew032011 (approved):

andrew032011:
Also I'm taking this review over from @henry-li-06 (spoke with him already), so no need to wait for his approval.

andrew032011:
FYI the deploy preview: https://deploy-preview-591--new-cornell-dti.netlify.app/ is crashing now.

andrew032011 (approved):

PR #593, opened 2024-03-24
Set up firebase for coffee chat bingo

alyssayzhang:
### Summary 

This pull request is the first step towards implementing coffee chat bingo. 

- [x] created a new collection called `coffee-chats` in `firebase.ts`
- [x]  defined a  DBCoffeeChat type

### Notion/Figma Link 
https://www.notion.so/Set-up-Firebase-for-coffee-chat-bingo-3d6c635217ec48a8aaf08ed7a5e4f197


patriciaahuang (commented):
The set up looks good in the prod database! You probably did some testing in the dev database, so there's an extra coffee-chat-requirement collection. In order to keep the dev database updated with the prod, you may want to remove it.

backend/src/types/DataTypes.d.ts
```diff
@@ -97,3 +97,12 @@ export type DevPortfolioSubmissionRequestLog = {
...
   };
 };
+
+export type DBCoffeeChat = {
+  uuid: string;
+  members: firestore.DocumentReference[];
+  image: string;
+  category: string;
+  description: string;
+  status: string;
```
oscarwang20, line 107: Make sure to add this new type to `common-types/index.d.ts` as well!

oscarwang20 (changes requested):

backend/src/types/DataTypes.d.ts
```diff
@@ -97,3 +97,13 @@ export type DevPortfolioSubmissionRequestLog = {
...
```
andrew032011, line 107: Change `status` type to be `Status`, not `string`.

common-types/index.d.ts
```diff
@@ -191,3 +191,13 @@ interface Shoutout {
   readonly hidden: boolean;
   readonly uuid: string;
 }
+
+interface CoffeeChat {
+  readonly uuid: string;
```
andrew032011, line 196: As you can see (from above), though we never followed this rule, we theoretically should be making all these fields `readonly`. Don't worry about the above ones but let's fix this one.

andrew032011 (approved):
A few nits, but overall looks good

oscarwang20 (approved):
Looks good to me! Thanks again for finishing this :D

PR #594, opened 2024-03-24
Create export to CSV feature for TEC dashboard

patriciaahuang:
### Summary

The leads needed a way to export the TEC information for each member to a CSV format. The TEC dashboard page now has the ability to export the entire table to a CSV format. The exported file will contain each member's name, netID, total TECs, initiative TECs, and the corresponding TECs they have from each team event.

For each member in `allMembers`, the `totalCredits` and `initiativeCredits` are calculated and added to the CSV data. Also, the number of credits they have for each team event in `teamEvents`  is added to the CSV data. After configuring the CSV format, the CSV exporter generates the file based on the data.

### Notion/Figma Link

https://www.notion.so/TEC-Export-to-csv-30f17a9b406243f0a24ec297626c1025?pvs=4

### Test Plan 

I checked to see that the data in the exported CSV file aligns with what is shown in the TEC dashboard page. 

Screenshot 2024-03-23 at 8 01 14 PM

Screenshot 2024-03-23 at 8 01 27 PM

### Notes

Depending on if the C&I lead wants this, there could be an additional column that shows if a member has met the credit requirement.

### DIFF
--- a/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
+++ b/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { Table, Header, Loader, Button, Icon, Checkbox } from 'semantic-ui-react';
+import { ExportToCsv, Options } from 'export-to-csv';
 import { useMembers } from '../../Common/FirestoreDataProvider';
 import { TeamEventsAPI } from '../../../API/TeamEventsAPI';
 import {
@@ -10,7 +11,6 @@ import {
 } from '../../../consts';
 import styles from './TeamEventDashboard.module.css';
 import NotifyMemberModal from '../../Modals/NotifyMemberModal';
-import { ExportToCsv, Options } from 'export-to-csv';
 
 const calculateMemberCreditsForEvent = (
   member: IdolMember,
@@ -64,12 +64,10 @@ const TeamEventDashboard: React.FC = () => {
       );
 
       const data = teamEvents.reduce(
-        (prev, event) => {
-          return {
-            ...prev,
-            [event.name]: calculateTotalCreditsForEvent(member, event)
-          };
-        },
+        (prev, event) => ({
+          ...prev,
+          [event.name]: calculateTotalCreditsForEvent(member, event)
+        }),
         {
           Name: `${member.firstName} ${member.lastName}`,
           NetID: `${member.netid}`,
--- a/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
+++ b/frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
@@ -36,6 +36,17 @@ const calculateTotalCreditsForEvent = (member: IdolMember, event: TeamEvent): nu
 const calculateInitiativeCreditsForEvent = (member: IdolMember, event: TeamEvent): number =>
   calculateMemberCreditsForEvent(member, event, true);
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
```diff
@@ -51,11 +52,58 @@ const TeamEventDashboard: React.FC = () => {
...
+  const handleExportToCsv = () => {
+    const csvData = allMembers.map((member) => {
+      const totalCredits = teamEvents.reduce(
+        (val, event) => val + calculateTotalCreditsForEvent(member, event),
+        0
+      );
+      const initiativeCredits = teamEvents.reduce(
+        (val, event) => val + calculateInitiativeCreditsForEvent(member, event),
+        0
+      );
```
cchrischen, line 64: The previous two functions are the same as the ones found on lines 119-122 and 150-157. I think it would be good to refactor them to something like `getAllCredits(member)` and `getAllInitiativeCredits(member)`, or something like that for code reusability.

frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
```diff
@@ -51,11 +52,58 @@ const TeamEventDashboard: React.FC = () => {
...
+
   return (
     <div className={styles.dashboardContainer}>
-      <Header as="h1" textAlign="center">
-        Team Event Dashboard
-      </Header>
+      <div className={styles.headerContainer}>
+        <Header as="h1">Team Event Dashboard</Header>
+        <div className={styles.csvButton}>
+          <Button onClick={() => handleExportToCsv()}>Export to CSV</Button>
```
cchrischen, line 104: nit: the arrow function can just be replaced with `handleExportToCsv`. (also, henry mentioned in my PR that using anonymous functions as inputs is frowned upon since its less performant.)

cchrischen:
Looks very cool :))!!

I just commented a few things that I stuck out to me, but other than that great work!

frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
```diff
@@ -51,11 +52,58 @@ const TeamEventDashboard: React.FC = () => {
...
+      const data = teamEvents.reduce(
+        (prev, event) => ({
+          ...prev,
+          [event.name]: calculateTotalCreditsForEvent(member, event)
+        }),
+        {
+          Name: `${member.firstName} ${member.lastName}`,
+          NetID: `${member.netid}`,
+          Total: totalCredits,
+          Initiative: initiativeCredits
```
cchrischen, line 75: nit: camelCase the keys :)

patriciaahuang: The keys show up exactly as the title of a column in the CSV file, which I feel like would look better starting with an uppercase letter since it's consistent with the names of the team events.

frontend/src/components/Admin/TeamEvent/TeamEventDashboard.tsx
henry-li-06: fyi the performance decrease is _extremely marginal_, we're talking like a couple milliseconds maybe. If there's a lot of these and _a lot of re-renders_ (like _A LOT_) ok sure maybe there's noticeable performance issues but for the most part it's nbd. 

but mostly for readability it's a little nicer to use a pre-defined handler function. that func could potentially also be re-used across different handlers

oscarwang20 (approved):
Looks good! Thanks you for addressing those comments :D

PR #598, opened 2024-03-25
Implement Coffee Chat Dao

alyssayzhang:
### Summary 

This PR implements the creates the functions that are used to set up coffee chat DAO.
- [ X ] get all coffee chat instances
- [ X ]  get coffee chat instances for user
- [ X ] create new coffee chat instance (1 submission)
- [ X ]  update coffee chat instance
- [ X ]  delete coffee chat instance
- [ X ] delete all coffee chats

Some things to keep in mind:
* users cannot coffee chat members from previous semesters (prevent users from creating new instances in this case)

### Notion/Figma Link 
https://www.notion.so/Dev-Roadmap-741d645c627a466da392fc8038ef0143?p=9719280ab9634834bba9d51a94ddeab8&pm=s


backend/src/dao/CoffeeChatDao.ts
```diff
@@ -0,0 +1,112 @@
...
+  async createCoffeeChat(coffeeChat: CoffeeChat): Promise<CoffeeChat> {
+    const [member1, member2] = coffeeChat.members;
+    const prevChats1 = await this.getCoffeeChatsByUser(member1);
+    const prevChats2 = await this.getCoffeeChatsByUser(member2);
+
+    const prevChats = [...prevChats1, ...prevChats2];
+
+    if (prevChats.some((c) => c.members.includes(member1) && c.members.includes(member2))) {
+      throw new Error(
+        `Cannot create coffee chat with member. Previous coffee chats from previous semesters exist.`
```
patriciaahuang, line 50: There should probably also be a check to make sure a member cannot coffee chat themselves. There doesn't seem to be anything preventing this right now.

patriciaahuang (commented):
Good job! The design aligns with the other DAO classes. I left a comment on creating a new coffee chat instance, but otherwise this looks good.

backend/src/dao/CoffeeChatDao.ts
```diff
@@ -0,0 +1,115 @@
...
+   * If provided, the object uuid will be used. If not, a new one will be generated.
+   * The pending field will be set to true by default.
+   * A member can not create a coffee chat the same person from previous semesters
+   */
+
+  async createCoffeeChat(coffeeChat: CoffeeChat): Promise<CoffeeChat> {
+    const [member1, member2] = coffeeChat.members;
+
+    if (member1.netid === member2.netid) {
+      throw new Error(`Cannot create coffee chat with yourself.`);
```
andrew032011, line 45: nit: no need for backticks if the string is not a template string

backend/src/dao/CoffeeChatDao.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+    const prevChats1 = await this.getCoffeeChatsByUser(member1);
+    const prevChats2 = await this.getCoffeeChatsByUser(member2);
+    const prevChats = [...prevChats1, ...prevChats2];
+
+    if (prevChats.some((c) => c.members.includes(member1) && c.members.includes(member2))) {
```
andrew032011, line 51: I would be very careful about this check. `includes(...)` uses object equality and we've had a bug introduced in candidate decider in the past from this. Consider instead checking against netIDs or emails.

andrew032011 (commented):
Generally, business logic such as rejecting duplicate chats  should be done on the API layer, not Dao. Can you remove this here?

Feel free to read the git history/copy paste somewhere to use for the API implementation.

backend/src/dao/CoffeeChatDao.ts
andrew032011: Good catch, let's put this and the other validation logic in the API layer, not Dao.

backend/src/dao/CoffeeChatDao.ts
henry-li-06: Alternatively, we use lodash's `isequal` which checks for deep equality for the profile update notification job. That can be used here.

backend/src/dao/CoffeeChatDao.ts
henry-li-06: interesting that eslint doesn't complain about this... 🤔

PR #605, opened 2024-04-18
Test CandidateDeciderDao

patriciaahuang:
Tested CandidateDeciderDao with a test file by creating mock data, and ensuring certain properties are met after using CandidateDeciderDao functions.


andrew032011 (approved):
Fast following in a separate PR for transactions tests is fine too, IMO. Otherwise these tests look good.

PR #606, opened 2024-04-29
upgrade to node 20

oscarwang20:
### Summary 

This PR upgrades our node version from 16 to 20.

### Test Plan 

Tested by running `build`,  `type-check`, and `lint`

### Breaking Changes 

- Make sure to upgrade your node version from 16 to 20 locally as well


frontend/next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
```
andrew-chen-figma, line 2: Are we sure we don't need this?

package.json
```diff
@@ -8,7 +8,6 @@
     "backend",
     "dti-website",
     "image-cli",
-    "dti-website-redesign",
```
andrew-chen-figma, line 11: 👍, maybe another follow up PR to remove the project entirely

dti-website/next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
```
andrew-chen-figma, line 2: See below

andrew032011, line 2: Are we sure we don't need this?

frontend/next-env.d.ts
andrew032011, line 2: See above, Are we sure we don't need this?

package.json
andrew032011, line 11: 👍 any maybe a follow up PR to remove this project entirely.

andrew032011 (approved):

frontend/next-env.d.ts
oscarwang20: This was removed when I ran `yarn run build` locally for testing purposes – everything runs fine without it, but I'll add it back just in case.

henry-li-06:
fyi i think some of the Github actions CI/cronjob should also be upgraded like setup-node, yarn cache, stuff like that. they're all outdated from 3 years ago

PR #610, opened 2024-05-02
Downgrade Nextjs and set config for new website

oscarwang20:
### Summary 

This PR fixes the netlify deploy error by downgrading the nextjs version for the new dti website.
See [this thread](https://stackoverflow.com/questions/76083438/module-not-found-cant-resolve-swc-helpers-src-class-private-field-init-mjs) for details

### Test Plan 

- Checked deployment preview and confirmed it works again

### DIFF
--- a/new-dti-website/package.json
+++ b/new-dti-website/package.json
@@ -20,7 +20,7 @@
     "embla-carousel-autoplay": "^8.0.0",
     "embla-carousel-react": "^8.0.0",
     "init": "^0.1.2",
-    "next": "latest",
+    "next": "13.3.0",
     "react": "latest",
     "react-dom": "latest",
     "shadcn-ui": "^0.4.0",
--- a/new-dti-website/next.config.js
+++ b/new-dti-website/next.config.js
@@ -1,4 +1,8 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {};
+const nextConfig = {
+  experimental: {
+    appDir: true
+  }
+};
 
 module.exports = nextConfig;
--- a/yarn.lock
+++ b/yarn.lock
@@ -2270,10 +2270,10 @@
   resolved "https://registry.yarnpkg.com/@next/env/-/env-12.3.4.tgz#c787837d36fcad75d72ff8df6b57482027d64a47"
   integrity sha512-H/69Lc5Q02dq3o+dxxy5O/oNxFsZpdL6WREtOOtOM1B/weonIwDXkekr1KV5DPVPr12IHFPrMrcJQ6bgPMfn7A==
 
-"@next/env@13.5.4":
-  version "13.5.4"
-  resolved "https://registry.yarnpkg.com/@next/env/-/env-13.5.4.tgz#777c3af16de2cf2f611b6c8126910062d13d222c"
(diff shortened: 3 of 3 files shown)


cchrischen (approved):
Looks good! Thank you for fixing the netlify issue :)

new-dti-website/package.json
```diff
@@ -20,7 +20,7 @@
     "embla-carousel-autoplay": "^8.0.0",
     "embla-carousel-react": "^8.0.0",
     "init": "^0.1.2",
-    "next": "latest",
+    "next": "13.3.0",
```
henry-li-06, line 23: Ok first of all none of these packages should be `latest`. That's a recipe for one day builds starting to fail for no good apparent reason. 

Secondly nextjs 14 is the current latest version with the latest features and performances upgrades. Unless I'm missing some context, is there a reason the package version needs to be downgraded to nextjs 13, and even more specifically 13.3?

oscarwang20: 1. Thanks for the tip :D. I'd like to address this so that scenarios like this don't come up unexpectedly again in the future. What do you think is the best course of action here – should I create another PR replacing `latest` with the current working version of each dependency?

2. I rolled back nextjs to 13.3.0 for now to allow @cchrischen and @keverie to merge their current work on the member page and apply page ASAP because we'd like to run a quick demo during the last All Hands on Saturday. The ongoing issue was that our netlify deploy previews were down for [new.cornelldti.org](http://new.cornelldti.org) and we kept getting an error where `@swc/helpers` could not be resolved. I read [this thread](https://github.com/vercel/next.js/issues/48593#issuecomment-1519914997) and learned that downgrading to 13.3 worked for some and it ended up working for us as well, so that's what I have it as for now until I resolve the peer dependency issue. I understand that v14 is the latest version, but I had no luck with fixing the deployment issue on that version. So, I decided the best course of action to take given our time constraints was just to roll back for now and release the features we've been working on throughout the semester and figure out how to upgrade nextjs without causing more errors in the deployment preview.

henry-li-06: Yeah given that `latest` is currently working just find out what version `latest` is actually referencing and replace it with that. I'd set the version to allow at least the patch but possibly the minor version to be flexible tho. 

For context, nodejs uses semantic versioning which means the version is in the form of `x.y.z` where _x_ is the major version, _y_ is the minor version, and _z_ is the patch version. Major changes contain breaking, non-backwards compatible changes, minor changes contain additional features that are backwards compatible, and patches are bug fixes that are also backwards compatible. 

`latest` is no good because you can get major version updates that have breaking functionality. API changes can happen in major version updates so it'd break code functionality unexpectedly. 

Nodejs has some docs about their semantic versioning usage [https://docs.npmjs.com/about-semantic-versioning](https://docs.npmjs.com/about-semantic-versioning). It looks like you'd either wanna use a `~` or `^` to allow for patches or minor versions to be  upgraded.

henry-li-06: Fair enough for the nextjs downgrade, but I'd still resolve it tho. If v14 has been working then it's unlikely that you _actually_ have to be downgrading to v13 to get the build to work.

PR #618, opened 2024-07-09
Add ability to archive members

cchrischen:
### Summary 
This pull request adds a section to the `admin/edit` page to download an archive, in json form, of all past and present DTI members given a returning members survey. This will be used by the ops leads at the end of each semester to generate the next semester's roster. The json will contain three keys: current, alumni, inactive, each storing a list of member details. 
* `NovaMember` now has a purpose again! It represents a member on the new DTI website, specifically with a coffee chat and image field.
* A sample returning members survey is attached.
* Endpoint `/member-archive?semesters=` allows you to define a limit to how many semesters back, if any, to include in the archive. 

### Notion/Figma Link 
https://www.notion.so/Member-archive-script-0540cb5e7d0349c4921f344b1875387e?pvs=4

### Test Plan 
image
Test CSV:
```
NetID,Graduation Semester,Returning
cc2785,May 2027,Yes
```
Output JSON:
[archive.json](https://github.com/user-attachments/files/16135513/archive.json)

### Breaking Changes 

* `NovaMember` as mentioned above, now extends `IdolMember`.

### DIFF
--- a/common-types/index.d.ts
+++ b/common-types/index.d.ts
@@ -42,25 +42,7 @@ interface IdolMemberDiff {
 }
 
 /** The data type used by Nova site to represent a DTI member. */
-interface NovaMember {
-  readonly netid: string;
-  readonly email: string;
-  readonly firstName: string;
-  readonly lastName: string;
-  readonly pronouns: string;
-  readonly graduation: string;
-  readonly major: string;
-  readonly doubleMajor?: string | null;
-  readonly minor?: string | null;
-  readonly website?: string | null;
-  readonly linkedin?: string | null;
-  readonly github?: string | null;
-  readonly hometown: string;
-  readonly about: string;
-  readonly subteams: readonly string[];
-  readonly formerSubteams?: readonly string[] | null;
-  readonly role: Role;
-  readonly roleDescription: RoleDescription;
+interface NovaMember extends IdolMember {
   readonly image?: string | null;
   readonly coffeeChatLink?: string | null;
 }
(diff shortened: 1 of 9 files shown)


common-types/index.d.ts
```diff
@@ -42,24 +42,9 @@ interface IdolMemberDiff {
 }
 
 /** The data type used by Nova site to represent a DTI member. */
```
oscarwang20, line 44: Can we update the comment to say that it's used by the new DTI website?

common-types/index.d.ts
```diff
@@ -42,24 +42,9 @@ interface IdolMemberDiff {
...
-  readonly website?: string;
-  readonly linkedin?: string;
-  readonly github?: string;
-  readonly hometown: string;
-  readonly about: string;
-  readonly subteams?: string[];
-  readonly formerSubteams?: string[];
-  readonly roleId: string;
-  readonly roleDescription: string;
+interface NovaMember extends IdolMember {
```
oscarwang20, line 45: nit: Maybe change `NovaMember` to something more descriptive like `RenderableMember` or `MemberProfile`?

frontend/public/sample_returning_members.csv
```diff
@@ -0,0 +1 @@
+NetID,Graduation Semester,Returning
```
oscarwang20, line 1: The returning member survey doesn't include Graduating Semester as a field, can we fetch this from our backend instead?

oscarwang20 (changes requested):
Thanks for getting this up. I requested a couple of changes, but overall it looks pretty good.

Another thing: Since they are no longer being maintained, it might be a good idea to add `dti-website` and `dti-website-redesign` to the `.eslintignore`.

PR #630, opened 2024-09-09
TEC Form: remove tec info card and fix dropdown selection bug

andrew032011:
### Summary 
There was a bug where clicking on any option in the dropdown would always default to the first option in the list, and the dropdown would close, preventing users from clicking on any other options.

Remove TEC info card so that it's possible to select other options. Frankly, I don't see why we need the card in the first place and it adds an unnecessary layer of complexity.

Before:

https://github.com/user-attachments/assets/72d10e30-89f7-4216-8a9e-8771d7841ce1

After:

https://github.com/user-attachments/assets/f3899ba2-a014-4027-a07e-781b95c71195

### Notion/Figma Link 

https://cornelldti.slack.com/archives/C044T59JD34/p1713279418092159


cchrischen (approved):
Looks good! Thanks for removing the unnecessary UI.

PR #631, opened 2024-09-09
TEC Form: Enforce TEC Limit for Bulk Submissions

andrew032011:
### Summary 
Currently we enforce TEC limit for single submissions that exceed the total credit limit of the TEC event. But we missed this check on bulk submissions, where a user can submit multiple events at a time that in total will exceed the max credit limit.

This PR enforces the TEC Limit for bulk submissions.

### Notion/Figma Link 

https://cornelldti.slack.com/archives/C07K7HFJLPJ/p1725836378114069

### Test Plan 

Manual

https://github.com/user-attachments/assets/27ae2aa6-f05e-4def-9dd3-7e04280623a1

### DIFF
--- a/frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
+++ b/frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
@@ -59,21 +59,18 @@ const TeamEventCreditForm: React.FC = () => {
   };
 
   const submitTeamEventCredit = async () => {
+    const creditsWithHours = teamEvent?.hasHours
+      ? Number(teamEvent?.numCredits || 0) * Number(hours)
+      : Number(teamEvent?.numCredits || 0);
     const getCredits: (attendance: TeamEventAttendance[]) => number = (attendance) => {
       const filteredAttendance = attendance.filter((event) => event.eventUuid === teamEvent?.uuid);
-      const sum = filteredAttendance.reduce(
-        (acc, event) =>
-          acc +
-          (teamEvent?.hasHours
-            ? Number(teamEvent?.numCredits || 0) * Number(hours)
-            : Number(teamEvent?.numCredits || 0)),
-        0
-      );
+      const sum = filteredAttendance.reduce((acc, event) => acc + creditsWithHours, 0);
 
       return sum;
     };
 
-    const totalCredits = getCredits(approvedAttendance) + getCredits(pendingAttendance);
+    const submittedCredits = getCredits(approvedAttendance) + getCredits(pendingAttendance);
+    const creditsToSubmit = images.length * creditsWithHours;
 
     if (!teamEvent) {
       Emitters.generalError.emit({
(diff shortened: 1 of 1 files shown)


frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
```diff
@@ -59,21 +59,18 @@ const TeamEventCreditForm: React.FC = () => {
...
       const filteredAttendance = attendance.filter((event) => event.eventUuid === teamEvent?.uuid);
-      const sum = filteredAttendance.reduce(
-        (acc, event) =>
-          acc +
-          (teamEvent?.hasHours
-            ? Number(teamEvent?.numCredits || 0) * Number(hours)
-            : Number(teamEvent?.numCredits || 0)),
-        0
-      );
+      const sum = filteredAttendance.reduce((acc, event) => acc + creditsWithHours, 0);
```
cchrischen, line 67: Thanks so much on your bug fixes. Though we don't use it much, I noticed something with team events with hourly credits that allows users to submit more than the max or prevent users from submitting. `creditsWithHours` is the total credits for the user's current submission not each event's hours. 

Here is an example of a user submitting more than the max team credits:

https://github.com/user-attachments/assets/5567e4c2-c3e1-4ce9-8dde-9e0c7ac170ff

A similar process can be repeated for a bug preventing users from submitting when below the max threshold.

cchrischen (approved):
I'm not sure if it's just me, but I did not know of this feature. Maybe in the description of uploading the event picture, we can describe how to do a bulk submission.

frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
andrew032011: Thanks for catching that! That appears to have been a long standing bug that went unnoticed. I definitely see usecases for this feature this semester so I'm glad we're fixing that now.

Can you double check again that the validation check now handle all the edge-cases you had in mind? I'm not fully sure what

> A similar process can be repeated for a bug preventing users from submitting when below the max threshold.

is referring to. Let me know if I've already addressed it!

andrew032011:
> Looks great! Thanks for catching and fixing these bugs. I'm not sure if it's just me, but I did not know of this feature. Maybe in the description of uploading the event picture, we can describe how to do a bulk submission.

Good idea! Done: https://github.com/cornell-dti/idol/pull/631/commits/e17c2db2722d99188d3dcbfb704fa83fcfa20cb8

frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
andrew032011: Ooh, I think I know what you mean. The problem before was that we weren't multiplying the number of hours with the credits correctly. We used the `hours` variable instead of `teamEventAttendance.hoursAttended`. I think that should all be resolved now since we calculate it correctly. Let me know if it's not and I can follow with a PR to fix!

PR #635, opened 2024-09-22
Set Up Coffee Chat API

patriciaahuang:
### Summary

This PR sets up the Coffee Chat API (and DAO). Members are not allowed to submit a coffee chat with themselves, or with someone they have coffee chatted in a previous semester.

### Notion/Figma Link

[Coffee Chat DAO Design Doc](https://www.notion.so/Dev-Roadmap-741d645c627a466da392fc8038ef0143?p=9719280ab9634834bba9d51a94ddeab8&pm=s)
[Coffee Chat API Design Doc](https://www.notion.so/Set-up-Coffee-Chat-API-a60dd100ca7c4430a0e0600b9639e2d1)

### Test Plan

The endpoints were tested using Jest ([PR #363](https://github.com/cornell-dti/idol/pull/636)).


backend/src/dao/CoffeeChatDao.ts
```diff
@@ -0,0 +1,102 @@
...
+   * @param uuid - DB uuid of CoffeeChat
+   */
+  async deleteCoffeeChat(uuid: string): Promise<void> {
+    await this.deleteDocument(uuid);
+  }
+
+  /**
+   * Deletes all coffee chats for all users
+   */
+  static async clearAllCoffeeChats(): Promise<void> {
```
andrew032011, line 95: You should be able to use `deleteCollection` here.

https://github.com/cornell-dti/idol/blob/95ee0b88aaaf2b6bc56ee4783b7edf1cf59d84d8/backend/src/utils/firebase-utils.ts#L57-L68

backend/src/API/coffeeChatAPI.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export const deleteCoffeeChat = async (uuid: string, user: IdolMember): Promise<void> => {
+  const isLeadOrAdmin = await PermissionsManager.isLeadOrAdmin(user);
+  const coffeeChat = await coffeeChatDao.getCoffeeChat(uuid);
+
+  if (!coffeeChat) return;
+
+  if (
+    !isLeadOrAdmin &&
+    (coffeeChat.members[0].email !== user.email || coffeeChat.members[1].email !== user.email)
```
andrew032011, line 87: Hm, I'm thinking about this more and I think it probably only makes sense for a lead to delete a coffee chat for now. i.e. say that a person submits a coffee chat and it gets approved for both members. And then one member for some reason deletes the coffee chat, since we don't have two "copies" of it, the other members gets affected too.

Right now since we just use the google form and the lead already has control over the coffee chat, I think it makes sense to make this not allowed for any non-lead. Seems much easier to manage permissions-wise this way, since coffee chats concern two people.

andrew032011 (approved):

PR #637, opened 2024-09-24
New Website: Compress images to under 1 MB

andrew032011:
### Summary 
Manually compress all images to be under 1 MB.

I don't see any differences immediately locally. Going to check deployment to see if there's any improvements.

### Test Plan 

Manually verify.


andrew032011:
We might still need this, since we did see image loading issues on the website. But the 3 second delay was caused by something else: https://github.com/cornell-dti/idol/pull/639

kevinmram (approved):
the new images are lookin' good to me, and the website seems to run smoothly- good decision to use .jpgs instead!

PR #638, opened 2024-09-24
TEC Admin tool: disable "Approve" when reason for reject isn't empty.

kevinmram:
Previously, a TEC Admin would be able to Approve a request even if the "reason for reject" text field was not empty, leading to accidentally approvals. However, this change disables the "Approve" button if there's text in the "reason for reject" field. (see attached video)

https://github.com/user-attachments/assets/07ac6966-e6e9-4392-a8e5-9649fcabb2f2

### DIFF
--- a/frontend/src/components/Admin/TeamEvent/TeamEventCreditReview.tsx
+++ b/frontend/src/components/Admin/TeamEvent/TeamEventCreditReview.tsx
@@ -99,6 +99,7 @@ const TeamEventCreditReview = (props: {
         onClick={() => {
           approveCreditRequest(teamEventAttendance);
         }}
+        disabled={reason !== ''}
       >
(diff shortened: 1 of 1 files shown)


frontend/src/components/Admin/TeamEvent/TeamEventCreditReview.tsx
```diff
@@ -99,6 +99,7 @@ const TeamEventCreditReview = (props: {
         onClick={() => {
           approveCreditRequest(teamEventAttendance);
         }}
+        disabled={reason !== ''}
```
andrew032011, line 102: Maybe don't need this?

andrew032011 (approved):

andrew032011 (approved):

PR #639, opened 2024-09-24
New Website: Fix 3 Second Slideshow Delay Bug

andrew032011:
### Summary 

There's been a mysterious 3-second delay on the slideshow whenever it's loaded. Originally, we thought it was due to image size, but it was actually a bug caused by how the initial state of the slideshow is set!

Set the initial value to be image 0, so that the slideshow is loaded upon page load.

### Test Plan 

Load up the page and observe that the slideshow loads immediately and doesn't take 3 seconds.


JasonMun7 (approved):
Great find!

PR #640, opened 2024-09-25
New DTI Website: Remove Fading Transition for Photos in Slideshow

andrew032011:
### Summary 
I forget whether we added the fading transition for flair or if it was actually part of the design. Frankly I think it looks borderline buggy, and I don’t like how the header and the slide aren’t perfectly fading in alignment (even if we fixed that, I don’t know if I’d like it).

Let's try switching the slideshow component to switch between photos without the fading transition

### Test Plan 
Before:

https://github.com/user-attachments/assets/7e43744c-c6c3-4859-b632-fd422ca858d8

After:

https://github.com/user-attachments/assets/3d58d0f3-e071-4e52-8a86-95756ec0d73e


JasonMun7 (approved):
Looking at the videos, I definitely think that it looks much better without the fade! I guess we can't really ask Hope anymore but if we really wanted to keep the fade we could make the duration longer in the future! Other than that looks great!

PR #642, opened 2024-09-27
Implement Coffee Chat Image API

patriciaahuang:
### Summary

Created an API for coffee chat images, allowing setting, getting, and deleting images.

### Notion/Figma Link

[Coffee Chat API Design Doc](https://www.notion.so/Set-up-Coffee-Chat-API-a60dd100ca7c4430a0e0600b9639e2d1?pvs=4)

### Test Plan

I manually tested each of the endpoints by replacing team event image endpoints with the corresponding coffee chat image endpoints. I set the images by submitting a team event credit, then tested if the images were set properly by retrieving the images in "Check Team Event Credits" section. I checked that the image deleted properly by opening the image as a URL, then deleting the team event, then refreshing the URL to see if it still exists in storage.

https://github.com/user-attachments/assets/b7a00aca-0f91-4b6b-adea-90bf291b5fe6

### Notes

I noticed some of the endpoints in team events API did not use the "user" parameter, so I removed it.

### Breaking Changes

I renamed the "EventProofImage" type to just "ProofImage" but ensured all use cases were modified properly.

### DIFF
--- a/backend/src/API/coffeeChatImageAPI.ts
+++ b/backend/src/API/coffeeChatImageAPI.ts
@@ -1,0 +1,48 @@
+import { bucket } from '../firebase';
+import { NotFoundError } from '../utils/errors';
+
+/**
+ * Gets Coffee Chat proof image for member
+ * @param name - the name of the image
+ * @param user - the member who made the request
+ * @throws NotFoundError if the requested image does not exist
+ * @returns a Promise to the signed URL to the image file
+ */
+export const getCoffeeChatProofImage = async (name: string): Promise<string> => {
+  const file = bucket.file(`${name}.jpg`);
+  const fileExists = await file.exists().then((result) => result[0]);
+  if (!fileExists) {
+    throw new NotFoundError(`The requested image (${name}) does not exist`);
+  }
+  const signedUrl = await file.getSignedUrl({
+    action: 'read',
+    expires: Date.now() + 15 * 60000
+  });
+  return signedUrl[0];
+};
+
+/**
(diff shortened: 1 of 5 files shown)


andrew032011 (commented):
Thoughts on unifying the logic between `coffeeChatImageAPI.ts` and `teamEventsImageAPI.ts`? We could just create something like `image-utils.ts` under `backend/src/utils/` and add the logic for getting the signed URL by name, uploading the image by name, and deleting all in that and just call them from `coffeeChatImageAPI` and `teamEventsImageAPI`.

andrew032011 (approved):

patriciaahuang:
> Thoughts on unifying the logic between `coffeeChatImageAPI.ts` and `teamEventsImageAPI.ts`? We could just create something like `image-utils.ts` under `backend/src/utils/` and add the logic for getting the signed URL by name, uploading the image by name, and deleting all in that and just call them from `coffeeChatImageAPI` and `teamEventsImageAPI`.
> 
> This also makes it easier for testing too since both coffee chat and team event image are using the same code path, so in theory if one of them works, the other one does.

I agree that we should unify the logic in `coffeeChatImageAPI.ts` and `teamEventsImageAPI.ts`. I'm thinking we could also unify the logic in `ImagesAPI.ts` since that is also starting to get repetitive.

kevinmram:
> Member

and similarly to what Patricia mentioned in my PR (#643), a unified `ImagesAPI `would be used for adding images to shoutouts (instead of reusing `teamEventsImageAPI`) and anywhere else we made need similar functionality in the future!

PR #643, opened 2024-09-29
[IDOL, shoutouts] - add image support

kevinmram:
Previously, IDOL shoutouts could only be a text message. However, this PR adds (optional) support for attaching an image with the shoutout!

(WIP - fixing some bugs with deleting shoutouts and image sizing)

https://github.com/user-attachments/assets/a6137da6-9954-4af3-a666-f48b5b7d9583


frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -27,6 +28,7 @@ type DBShoutout = {
   timestamp: number;
   hidden: boolean;
   uuid: string;
+  images?: string[]; // Ensure that images field is part of the shoutout
```
patriciaahuang, line 31: Remove comment

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -36,6 +38,7 @@ const AdminShoutouts: React.FC = () => {
   const [lastDate, setLastDate] = useState<Date>(new Date());
   const [hide, setHide] = useState(false);
   const [loading, setLoading] = useState(false);
+  const [imageUrls, setImageUrls] = useState<{ [key: string]: string }>({}); // Store image URLs in an object
```
patriciaahuang, line 41: Remove comment

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -83,8 +86,19 @@ const AdminShoutouts: React.FC = () => {
     }
   }, [allShoutouts, earlyDate, lastDate, view]);
 
+  const fetchImages = useCallback((shoutouts: Shoutout[]) => {
+    shoutouts.forEach((shoutout) => {
+      if (shoutout.images?.length) {
+        ImagesAPI.getEventProofImage(shoutout.images[0]).then((url) => {
```
patriciaahuang, line 92: Since this isn't an event proof image, we should probably have a separate `getShoutoutImage` function. Same with the upload image function. However, the logic behind all these functions will be the same and will be repetitive, so I'm thinking maybe we could just have an overall `getImage` and `uploadImage` function in the Images API. @andrew032011 What do you think? This will also affect the coffee chat images, so I could make this change in PR #642.

patriciaahuang (commented):
I left a comment that might change the Images API. Also, make sure to address the GitHub Actions checks :)

frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
```diff
@@ -24,7 +43,20 @@ const ShoutoutCard = (props: {
...
       </Card.Group>
       <Card.Content description={shoutout.message} />
+
+      {isLoading ? (
+        <Loader active inline />
+      ) : (
+        image && (
+          <Card.Content>
+            <div className={styles.imageContainer}>
+              <Image src={image} size="small" />
```
JasonMun7, line 53: The Image component should have an alt attribute for accessibility purposes since it seems that the image is being dynamically loaded!

frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
```diff
@@ -24,7 +43,20 @@ const ShoutoutCard = (props: {
         <ShoutoutDeleteModal uuid={shoutout.uuid} setGivenShoutouts={setGivenShoutouts} />
       </Card.Group>
       <Card.Content description={shoutout.message} />
+
+      {isLoading ? (
+        <Loader active inline />
+      ) : (
+        image && (
```
JasonMun7, line 50: Maybe also check to see if the image string is non-empty and non-null like image?.length > 0 since it could be possible that an empty string or a malformed image URL might still pass the condition. Might need clarification on this.

frontend/src/components/Forms/ShoutoutsPage/ShoutoutCard.tsx
```diff
@@ -1,18 +1,37 @@
...
   setGivenShoutouts: React.Dispatch<React.SetStateAction<Shoutout[]>>;
 }): JSX.Element => {
   const { shoutout, setGivenShoutouts } = props;
 
+  const [image, setImage] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+
   const fromString = shoutout.isAnon
     ? 'From: Anonymous'
     : `From: ${shoutout.giver?.firstName} ${shoutout.giver?.lastName}`;
```
JasonMun7, line 18: I'm wondering if there could be a case where either the firstname and the lastName are null/undefined? Would be helpful to take these cases and provide a fallback for both in that case!

JasonMun7 (commented):
Hey Kevin this is looking fantastic! Can't wait to use this during All hands! I left some comments that are pretty nit picky for the states in the code as I don't have too much experience with the api yet! Other than the action checks and my comments, it's looking great!

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
andrew032011: Thanks for bringing this up! Yeah, I agree we should unify it into one API. getImage and estimate (or uploadImage). @patriciaahuang do you want to repurpose your Coffee Chat image API PR to instead make the generalizable API? I thiiink it should be as simple as some renaming.

common-types/index.d.ts
```diff
@@ -207,6 +207,7 @@ interface Shoutout {
   readonly timestamp: number;
   readonly hidden: boolean;
   readonly uuid: string;
+  readonly images?: string[];
```
andrew032011, line 210: Make sure you update `DBShoutout` in [DataTypes.d.ts](https://github.com/cornell-dti/idol/blob/f64bb993ab2646586e5f476f375368418aab5ce6/backend/src/types/DataTypes.d.ts#L11-L19) as well.

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -209,6 +224,21 @@ const AdminShoutouts: React.FC = () => {
                   className={styles.presentShoutoutMessage}
                   content={shoutout.message}
                 />
+                {(() => {
+                  let content;
```
andrew032011, line 228: This looks like the exact same as the code below. It's also not the cleanest pattern to put a function and then call it here. Instead, can you abstract this code out into a separate component?

andrew032011:
We should also delete the corresponding images for a shoutout if the shoutout itself is deleted.

andrew032011:
Sorry I think I might've not worded this message properly. 

> We should also delete the corresponding images for a shoutout if the shoutout itself is deleted.

I meant that if a user deletes a shoutout on the member side, that the image should also be deleted from Google Storage! Not to add another feature delete shoutouts from the admin side. Could you revert the change that adds the "Delete" button?

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -57,18 +59,15 @@ const AdminShoutouts: React.FC = () => {
       const filteredShoutouts = allShoutouts
         .filter((shoutout) => {
           const shoutoutDate = new Date(shoutout.timestamp);
-          // Set time to be 4:59:59AM UTC/11:59PM EST/12:59AM EDT
```
andrew032011, line 60: Can we keep the comments?

frontend/src/components/Admin/AdminShoutouts/AdminShoutouts.tsx
```diff
@@ -57,18 +59,15 @@ const AdminShoutouts: React.FC = () => {
...
               4,
               59,
               59,
               59
             ) +
-              60 * 60 * 1000 * 24
+            60 * 60 * 1000 * 24
           );
-
-          // Set time to be 5AM UTC/12AM EST/1AM EDT
```
andrew032011, line 71: ^ same here, can we keep this?

andrew032011:
I see a slight dilemma with this feature that I didn’t anticipate before…real-time shoutouts is going to be really annoying if we’re constantly having to refetch images. Each time we refetch, the image always flickers. I wonder if we can cache all images that we have already loaded in on the frontend, and only refetch images that we haven't stored yet?

Also, when I click between tabs, the images are always refetched, but I think we should fetch images on the same cadence that we fetch shoutouts....so we should also store image in some state and avoid refetching every time we switch tabs.

andrew032011 (approved):

PR #647, opened 2024-10-01
Unify Logic in Image APIs

patriciaahuang:
### Summary

Unified the logic in the backend image APIs and the frontend image API, since there are starting to be more use cases for images but the logic is all the same.

### Test Plan

I checked that team event proof images and member profile images still works properly, using the same method as in PR #642.


backend/src/api.ts
```diff
@@ -315,6 +321,20 @@ loginCheckedPut('/coffee-chat', async (req, user) => ({
   coffeeChats: await updateCoffeeChat(req.body, user)
 }));
 
+// Coffee Chat Proof Image
```
andrew032011, line 324: What if we unified this on the backend as well? So just have a generic `/image/:name(*)` (get), `/image-signed-url/:name(*)` (get), `/image/:name(*)` (delete) endpoints? I think the logic is the same across all the different use-cases anyways.

Instead of the utils, we'd have just three endpoints total (as opposed to 9)

andrew032011: You should keep the image names the same though, so I would still make sure the images, when they're created, are prefixed correctly (i.e. `eventProofs/`, `coffeeChatProofs/`, `member-image/`)

patriciaahuang: Oops, I missed unifying this file as well. Will do that!

henry-li-06: Even better if it's just prefixed with `/files` to be super generic :D

backend/src/utils/image-utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  const fileExists = await file.exists().then((result) => result[0]);
+  if (!fileExists) {
+    throw new NotFoundError(`The requested image (${name}) does not exist`);
+  }
+  const signedUrl = await file.getSignedUrl({
+    action: 'read',
+    expires: Date.now() + 15 * 60000
+  });
+  return signedUrl[0];
+};
```
henry-li-06, line 37: ok so let's rename these (ik i'm the one that gave them shitty names in the first place)
- `getWriteSignedURL`
- `getReadSignedURL` 

or something a long those lines

backend/src/utils/image-utils.ts
```diff
@@ -0,0 +1,97 @@
...
+import { getNetIDFromEmail } from './memberUtil';
+import { NotFoundError } from './errors';
+
+/**
+ * Sets image for member
+ * @param name - the name of the image
+ * @returns a Promise to the signed URL to the image file
+ */
+export const setImage = async (name: string): Promise<string> => {
+  const file = bucket.file(`${name}.jpg`);
```
henry-li-06, line 11: on a side note - probably better off including the file extension in the actual file path itself instead of assuming they're all `.jpg`

backend/src/API/imageAPI.ts
```diff
@@ -25,37 +59,61 @@ export const allMemberImages = async (): Promise<readonly ProfileImage[]> => {
...
+      });
+      const fileName = await file.getMetadata().then((data) => data[1].body.name);
+      return {
+        fileName,
+        url: signedURL[0]
+      };
+    })
+  );
+
+  images
```
andrew032011, line 87: I think this value looks like it's floating and nothing is assigned to it?

andrew032011:
I'm unable to fetch existing images anymore here. Did we change how we construct the image paths? The image paths should still be the same, even if we're changing the endpoints.

patriciaahuang:
> I'm unable to fetch existing images anymore here. Did we change how we construct the image paths? The image paths should still be the same (to maintain backwards compatibility with the existing images we're storing), even if we're changing the endpoints.

I included the file extension in the file path to avoid assuming all images are `.jpg`. Before, `.jpg` was manually added to the file path name when getting an image in `imageAPI.ts`, so removing that will cause existing images to be fetched incorrectly (instead of fetching `pathname.jpg` it'll try fetching just `pathname`). Should I revert this change to maintain backwards compatibility?

andrew032011:
> > I'm unable to fetch existing images anymore here. Did we change how we construct the image paths? The image paths should still be the same (to maintain backwards compatibility with the existing images we're storing), even if we're changing the endpoints.
> 
> I included the file extension in the file path to avoid assuming all images are `.jpg`. Before, `.jpg` was manually added to the file path name when getting an image in `imageAPI.ts`, so removing that will cause existing images to be fetched incorrectly (instead of fetching `pathname.jpg` it'll try fetching just `pathname`). Should I revert this change to maintain backwards compatibility?

Wow, good find! In an ideal world we would go in and change all images to not have the `.jpg` suffix and patch this, so that your fix can stay and we no longer suffix our images with `.jpg`. Of course, there is always a risk with such a change and trying to change all these images might be more work than necessary.

Let's keep the `.jpg` for now! (rationale: This gets us unblocked the quickest. What might be easier is at the end of this semester, we can remove all images in firestore and start afresh next semester. And if it really bothers us this semester, we can always remove it later by writing a script to change all the images)

patriciaahuang:
> > > I'm unable to fetch existing images anymore here. Did we change how we construct the image paths? The image paths should still be the same (to maintain backwards compatibility with the existing images we're storing), even if we're changing the endpoints.
> > 
> > 
> > I included the file extension in the file path to avoid assuming all images are `.jpg`. Before, `.jpg` was manually added to the file path name when getting an image in `imageAPI.ts`, so removing that will cause existing images to be fetched incorrectly (instead of fetching `pathname.jpg` it'll try fetching just `pathname`). Should I revert this change to maintain backwards compatibility?
> 
> Wow, good find! In an ideal world we would go in and change all images to not have the `.jpg` suffix and patch this, so that your fix can stay and we no longer suffix our images with `.jpg`. Of course, there is always a risk with such a change and trying to change all these images might be more work than necessary.
> 
> Let's keep the `.jpg` for now! (rationale: This gets us unblocked the quickest. What might be easier is at the end of this semester, we can remove all images in firestore and start afresh next semester. And if it really bothers us this semester, we can always remove it later by writing a script to change all the images)

Sounds good, thanks for catching this!

frontend/src/components/Forms/UserProfile/UserProfileImage.tsx
```diff
@@ -31,7 +31,7 @@ const UserProfileImage: React.FC = () => {
         .then((res) => res.blob())
         .then((blob) => {
           imageURL = window.URL.createObjectURL(blob);
-          ImagesAPI.uploadMemberImage(blob);
+          ImagesAPI.uploadImage(blob, `image-signed-url/${userInfo ? userInfo.email : ''}`);
```
andrew032011, line 34: we can move this prefix `image-signed-url/` into the ImagesAPI.uploadImage function

frontend/src/components/Forms/TeamEventCreditsForm/TeamEventsCreditDashboard.tsx
```diff
@@ -63,7 +63,7 @@ const TeamEventCreditDashboard = (props: {
           headerMsg: 'Team Event Attendance Deleted!',
           contentMsg: 'Your team event attendance was successfully deleted!'
         });
-        ImagesAPI.deleteEventProofImage(attendance.image);
+        ImagesAPI.deleteImage(`image/${attendance.image}`);
```
andrew032011, line 66: we can move this prefix `image/` into the ImagesAPI.deleteImage function

frontend/src/API/ImagesAPI.ts
```diff
@@ -17,48 +14,19 @@ export default class ImagesAPI {
...
+  private static getSignedURL(name: string): Promise<string> {
+    const responseProm = APIWrapper.get(`${backendURL}/${name}`).then((res) => res.data);
     return responseProm.then((val) => val.url);
   }
 
-  public static uploadMemberImage(body: Blob): Promise<void> {
-    return this.getSignedURL().then((url) => {
-      const headers = { 'content-type': 'image/jpeg' };
+  public static uploadImage(body: Blob, name: string): Promise<void> {
+    return this.getSignedURL(`${name}`).then((url) => {
```
andrew032011, line 23:
```suggestion
    return this.getSignedURL(name).then((url) => {
```

frontend/src/API/ImagesAPI.ts
```diff
@@ -17,48 +14,19 @@ export default class ImagesAPI {
...
+      const headers = { 'content-type': `${body.type}` };
```
andrew032011, line 24:
```suggestion
      const headers = { 'content-type': body.type };
```

frontend/src/API/ImagesAPI.ts
```diff
@@ -17,48 +14,19 @@ export default class ImagesAPI {
     });
   }
 
-  private static getSignedURL(): Promise<string> {
-    const responseProm = APIWrapper.get(`${backendURL}/member-image-signedURL`).then(
-      (res) => res.data
-    );
+  private static getSignedURL(name: string): Promise<string> {
+    const responseProm = APIWrapper.get(`${backendURL}/${name}`).then((res) => res.data);
```
andrew032011, line 18:
```suggestion
    const responseProm = APIWrapper.get(`${backendURL}/image-signed-url/${name}`).then((res) => res.data);
```

frontend/src/API/ImagesAPI.ts
```diff
@@ -17,48 +14,19 @@ export default class ImagesAPI {
...
     return responseProm.then((val) => val.url);
   }
 
-  public static uploadMemberImage(body: Blob): Promise<void> {
-    return this.getSignedURL().then((url) => {
-      const headers = { 'content-type': 'image/jpeg' };
+  public static uploadImage(body: Blob, name: string): Promise<void> {
+    return this.getSignedURL(`image-signed-url/${name}`).then((url) => {
```
andrew032011, line 23:
```suggestion
    return this.getSignedURL(name).then((url) => {
```

andrew032011 (approved):

andrew032011:
Hold on, just tested manually and if you upload and then try to get a TEC image, it does not fetch the image properly. I think it has something to do with how the file is named when it is uploaded.

PR #651, opened 2024-10-06
Filter Admin Only Cards for Home Page

andrew032011:
### Summary 
We previously were only filtering navigation cards on the "Forms" tab, but not the home page. This is because we had two navigation card components, `NavCard` and `NavigationCard`, the first of which did not contain logic to filter against the "adminOnly" flag. We should allow `adminOnly` to be used on the homepage if we want to gate something through a feature flag.

Also, update the Homepage to use `NavigationCard` and delete `NavCard`. This consequently also fixes an alignment issue on the homepage too.

### Test Plan 

Verification steps:
1. Check that the "Admin" and "Forms" tab still look the same
2. Check that the homepage styling still looks acceptable (in terms of spacing and alignment of the cards)

Before:
Screenshot 2024-10-06 at 12 46 27 AM

After:
Screenshot 2024-10-06 at 12 46 21 AM

### DIFF
--- a/frontend/src/components/Common/NavigationCard/NavigationCard.module.css
+++ b/frontend/src/components/Common/NavigationCard/NavigationCard.module.css
@@ -1,9 +1,3 @@
-.content {
-  margin-left: 2rem;
-  margin-right: 2rem;
-  padding-top: 2rem;
-}
-
 @media (max-width: 900px) {
   .cardsContainer {
     display: flex;
--- a/frontend/src/components/Common/NavigationCard/NavigationCard.tsx
+++ b/frontend/src/components/Common/NavigationCard/NavigationCard.tsx
@@ -19,30 +19,28 @@ export default function NavigationCard({ testID, items }: Props): JSX.Element {
 
   return (
     <div data-testid={testID}>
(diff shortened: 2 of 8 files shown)


kevinmram (approved):
It's a good catch that we were using NavCard and NavigationCard- using just NavigationCard across both the homepage and other sections, reducing redundancy and simplifying our codebase. However, there are a few CSS changes, especially concerning responsive design. I'd double-checking that the adjustments maintain consistency in mobile views, particularly for the .card and .quick containers, but everything should be lookin' good! (you're the goat A.C.)

andrew032011:
> It's a good catch that we were using NavCard and NavigationCard- using just NavigationCard across both the homepage and other sections, reducing redundancy and simplifying our codebase. However, there are a few CSS changes, especially concerning responsive design. I'd double-checking that the adjustments maintain consistency in mobile views, particularly for the .card and .quick containers, but everything should be lookin' good! (you're the goat A.C.)

Thanks for the reminder! I checked this and it looks like the cards were previously not aligned, and now they are. So net-positive!


Before:

image


After:
image

PR #668, opened 2024-11-06
Home Page: Remove Image and Add Red Blobs for Hero Component/Navbar Region

andrew032011:
### Summary 
Screenshot 2024-11-06 at 10 15 26 AM
Remove the old image and add in `Redblob` 's for the hero component section.

I'm leaving out the red blobs on the bottom portion since we're still missing the line component. Since we're using absolute positioning for the red blobs, I think it'll be easier to add the rest of the blobs in once the line component is added.

I also fixed the display of the logo for the navbar in mobile view (blur the rest of the page except for the logo):

image

### Notion/Figma Link 

https://www.notion.so/Fix-Harsh-Transition-from-Header-to-Rest-of-Home-Page-1360ad723ce1806e8c62dbfa3b70ba9d?pvs=4


new-dti-website/src/app/page.tsx
```diff
@@ -116,11 +116,12 @@ const Home: React.FC = () => {
           <div className="flex-grow">
             <Slideshow selectedImage={selectedIcon} />
           </div>
+          <RedBlob intensity={0.6} className="left-[1200px] top-[0px]" />
```
cchrischen, line 119: This makes the page horizontally scrollable. One way that I found to solve this is to wrap the blob in a div and make the most parent div have `overflow-hidden`. This also introduces some z-index issues that show up.

new-dti-website/components/navbar.tsx
```diff
@@ -77,13 +77,15 @@ const Navbar: React.FC = () => {
...
+          ) : (
+            <div className="h-12 w-auto md:h-14" />
           )}
         </div>
       </div>
       {isMenuOpen && (
-        <div className="fixed inset-0 bg-black bg-opacity-25 backdrop-blur-sm z-40 lg:hidden">
-          <div className="w-full h-full flex flex-col items-end">
-            <div className="p-4">
+        <div className="fixed inset-0 bg-black bg-opacity-25 lg:hidden">
```
cchrischen, line 86: Thoughts on keeping the backdrop blur? There's now a harsh, but less noticeable, transition in the navbar.
image

cchrischen (approved):
Thanks for the quick turnaround! A little note just to have documented (doesn't have to be addressed in this PR) is that the slideshow isn't responsive is smaller viewports.

new-dti-website/components/navbar.tsx
andrew032011: You mean remove the black bar entirely and just keep the original background?

andrew032011:
> Thanks for the quick turnaround! A little note just to have documented (doesn't have to be addressed in this PR) is that the slideshow isn't responsive is smaller viewports.

Thanks for the quick review! Slideshow mobile styling is tracked in this ticket: https://www.notion.so/Hero-Page-Mobile-Styling-a3192f43d4c44dc6a8f6ede779edb575?pvs=4, we'll definitely work on that before launch.

new-dti-website/components/navbar.tsx
cchrischen: Hmm.. I meant that the background behind the cornell dti logo on top should also be blurred, so we don't get the same harsh transition in this screen.

andrew032011: Oh hmm, I was following the design which deliberately did not blur the logo, I can see how it would make sense to get rid of the black bar there. I'll double check with Vannessa.

PR #670, opened 2024-11-07
Coffee Chat: Indicate If Member Meets Category

patriciaahuang:
### Summary 

On the submission form, a warning will appear if a selected member does not meet the selected category. If there is no data for the selected member, no warning will appear. 

On the admin side, each submission will have a different color depending on whether the member meets/doesn't meet the category or if there is not enough data to know.

### Demo

https://github.com/user-attachments/assets/ef83ab5b-21e8-4595-bbb0-464cf834e537


backend/src/API/memberAPI.ts
```diff
@@ -169,3 +169,89 @@ export const generateMemberArchive = async (
...
+
+/**
+ * Gets the properties of a member given their email.
+ * @param email - the email of a member.
+ * @returns the `MemberProperties` where the email matches the given email.
+ */
+export const getMemberProperties = async (email: string): Promise<MemberProperties | undefined> =>
+  MembersDao.getMemberProperties(email);
+
+/**
```
andrew032011, line 181: Sorry for the back and forth. This feels very coffee chat specific, so I lean towards putting it in the CoffeeChatAPI.

frontend/src/API/MembersAPI.ts
```diff
@@ -64,4 +64,24 @@ export class MembersAPI {
       member
     ).then((res) => res.data);
   }
+
+  public static checkMemberMeetsCategory(
+    otherMember: IdolMember,
+    category: string
+  ): Promise<boolean | undefined> {
+    const res = APIWrapper.get(
+      `${backendURL}/member/${otherMember.email}/${Buffer.from(category).toString('base64')}`
```
andrew032011, line 73: hm, this doesn't feel the most idiomatic and will make debugging tricky. Is putting the category name directly here too long? If so, we could maybe consider using some other key that can be easily traced back to the coffee chat category? Something that's more human-readable would make it easier to debug.

patriciaahuang: True. I encoded it because if a category included '/', it messed with getting the request parameters.

andrew032011: Try:

```
const category = "category/with/slash";
const encodedCategory = encodeURIComponent(category);
```

backend/src/api.ts
```diff
@@ -238,6 +239,11 @@ loginCheckedPost('/member-archive', async (req, user) => ({
   archive: await generateMemberArchive(req.body, user, req.query.semesters as number | undefined)
 }));
 
+loginCheckedGet('/member/:email/:category', async (req, user) => {
+  const result = await checkMemberMeetsCategory(req.params.email, user, req.params.category);
+  return { result };
```
andrew032011, line 244: Right now if `result` returns `undefined` the response just looks like:

```
{}
```
How about we make the response a little more clear for the uncertain cases? Maybe "pass" | "fail" | "no data"? Feel free to adjust naming as you see fit.

backend/src/API/memberAPI.ts
```diff
@@ -169,3 +170,81 @@ export const generateMemberArchive = async (
...
 };
+
+/**
+ * Checks if a member meets a category.
+ * @param otherMemberEmail - the email of the member we are checking.
+ * @param submitter - the member that submitted the coffee chat.
+ * @param encodedCategory - the category we are checking with (encoded with base64).
+ * @returns true if a member meets a category, false if not, undefined if not enough data to know.
+ */
+export const checkMemberMeetsCategory = async (
```
andrew032011, line 181: Sorry, what I meant was, can you move this function into `coffeeChatAPI`, delete `getMemberProperties` from `coffeeChatAPI` (since we don't want to expose the functionality of actually showing a user's properties), and then just call the `CoffeeChatDao.getMemberProperties` function directly from `checkMemberMeetsCategory`?

frontend/src/API/MembersAPI.ts
```diff
@@ -69,8 +69,9 @@ export class MembersAPI {
     otherMember: IdolMember,
     category: string
   ): Promise<boolean | undefined> {
+    const encode = (input: string): string => input.replace(/\//g, '*');
     const res = APIWrapper.get(
-      `${backendURL}/member/${otherMember.email}/${Buffer.from(category).toString('base64')}`
+      `${backendURL}/member/${otherMember.email}/${encode(category)}`
```
andrew032011, line 74: Can you try `encodeURIComponent`? Replacing each `/` with a `*` works but it feels a little hacky to me, since we're just converting it into a random character. What if a category contained a `*` for some reason?

frontend/src/components/Forms/CoffeeChatsForm/CoffeeChatsForm.tsx
```diff
@@ -30,7 +31,12 @@ const CoffeeChatsForm: React.FC = () => {
       setIsChatLoading(false);
     });
     CoffeeChatAPI.getCoffeeChatBingoBoard().then((board) => setBingoBoard(board));
-  }, [userInfo]);
+    if (member && category) {
```
andrew032011, line 34: Lumping these API calls into a `useEffect` makes it so if one of them is triggered, then all calls are made. So each time I change the name in the dropdown, all three endpoints are called (instead of just `checkMemberMeetsCategory`). Maybe we can split this off into its own `useEffect` hook with the proper dependencies? And we can leave this one the way it was.

frontend/src/API/CoffeeChatAPI.ts
```diff
@@ -44,4 +44,24 @@ export default class CoffeeChatAPI {
     );
     return res.board as string[][];
   }
+
+  public static checkMemberMeetsCategory(
+    otherMember: IdolMember,
+    category: string
+  ): Promise<'pass' | 'fail' | 'no data'> {
```
andrew032011, line 51: nit: Make this a type, so that we don't have to copy and paste this a bunch.

andrew032011 (approved):

frontend/src/components/Admin/CoffeeChats/CoffeeChatDetails.tsx
```diff
@@ -109,6 +133,11 @@ const CoffeeChatDetails: React.FC = () => {
       </div>
 
       <h1 className={styles.categoryName}>Category: {category}</h1>
+      <p className={styles.description}>
+        Coffee chats will appear <strong style={{ color: '#02c002' }}>green</strong> if the coffee
+        chatted member meets the category, <strong style={{ color: '#f23e3e' }}>red</strong> if they
+        do not, and <strong>white</strong> if their status is uncertain.
```
andrew032011, line 139:
```suggestion
        do not, and <strong>white</strong> if we don't have the data to determine whether or not the member meets the category.
```

andrew032011 (approved):

PR #674, opened 2024-11-07
Coffee Chat: Fix Infinite Calls to Coffee Chat API Bug on Admin Tool

andrew032011:
### Summary 
- Fix infinite loop bug on admin view: we put an API call directly inside of a component instead of a `useEffect` call. Since the callback function updates the state, this causes the component to re-render, which then makes the API call again, resulting in an infinite loop
- Improve loading state so that the colors don't flash in the user's face (wait for the auto-checker to finish running against all the coffee chats first, then display all the cards with their auto-checker results)

### Test Plan 
- Manually observe only the correct number of API calls are made, no more infinite loop when trying to repro

### DIFF
--- a/frontend/src/components/Admin/CoffeeChats/CoffeeChatDetails.tsx
+++ b/frontend/src/components/Admin/CoffeeChats/CoffeeChatDetails.tsx
@@ -8,65 +8,79 @@ import CoffeeChatAPI from '../../../API/CoffeeChatAPI';
 import { Emitters } from '../../../utils';
 import { ALL_STATUS } from '../../../consts';
 
+type CoffeeChatCardProps = {
+  status: Status;
+  chat: CoffeeChat;
+};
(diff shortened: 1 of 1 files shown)


frontend/src/components/Admin/CoffeeChats/CoffeeChatDetails.tsx
```diff
@@ -85,10 +99,6 @@ const CoffeeChatDetails: React.FC = () => {
   const [status, setStatus] = useState<Status | string>();
   const [coffeeChats, setCoffeeChats] = useState<CoffeeChat[]>([]);
 
-  useEffect(() => {
```
andrew032011, line 88: This is a duplicate call. If you look further down, we already fetch all coffee chats.

patriciaahuang (approved):

PR #675, opened 2024-11-08
Coffee Chat: Show Error on Admin Side, Add Re-run Auto Checker Button, Fix checkMemberMeetsCategory Logic

patriciaahuang:
### Summary 

This PR customizes each error message depending on the category, and displays the error message on the admin side if a member does not meet the category. There is also now a button to re-run the auto checker on the admin side.

The previous logic for `checkMemberMeetsCategory` didn't correctly identify between false/fail and undefined/no data, so that is fixed.

### Demo

https://github.com/user-attachments/assets/1cc0b432-a7c4-4cd4-925f-bbf9a1d51e7d

### DIFF
--- a/backend/src/API/coffeeChatAPI.ts
+++ b/backend/src/API/coffeeChatAPI.ts
@@ -135,7 +135,7 @@ export const runAutoChecker = async (uuid: string, user: IdolMember): Promise<Co
   const canRunAutoChecker = await PermissionsManager.isLeadOrAdmin(user);
   if (!canRunAutoChecker)
     throw new PermissionError(
-      `User with email ${user.email} does not have permission to regrade dev portfolio submissions`
+      `User with email ${user.email} does not have permission to run auto checker`
     );
 
   const coffeeChat = await coffeeChatDao.getCoffeeChat(uuid);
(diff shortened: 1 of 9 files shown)


frontend/src/components/Admin/CoffeeChats/CoffeeChatDetails.tsx
```diff
@@ -114,6 +117,31 @@ const CoffeeChatDetails: React.FC = () => {
     return map;
   }, [coffeeChats]);
 
+  const runAutoCheckerForCategory = async (category: string) => {
```
andrew032011, line 120: I would move this logic to the backend/API layer, since you're doing multiple data operations here. Take a look at how `regradeSubmissions` is implemented for Dev Portfolios for inspiration.

backend/src/API/coffeeChatAPI.ts
```diff
@@ -125,18 +125,52 @@ export const clearAllCoffeeChats = async (user: IdolMember): Promise<void> => {
...
+ * Checks if a member meets a category for the specified coffee chat.
+ * @param uuid - the uuid of the coffee chats we are checking.
+ * @param user - the IdolMember making the request.
+ * @returns the updated coffee chat
+ */
+export const runAutoChecker = async (uuid: string, user: IdolMember): Promise<CoffeeChat> => {
+  const canRunAutoChecker = await PermissionsManager.isLeadOrAdmin(user);
+  if (!canRunAutoChecker)
+    throw new PermissionError(
+      `User with email ${user.email} does not have permission to regrade dev portfolio submissions`
```
andrew032011, line 138: Fix this error message ("regrade dev portfolio submissions")

backend/src/API/coffeeChatAPI.ts
```diff
@@ -145,61 +179,135 @@ export const checkMemberMeetsCategory = async (
...
+  if (otherMember && submitter) {
+    if (category === 'an alumni') {
+      status = (await allMembers()).every((member) => member.email !== otherMember.email)
+        ? 'pass'
+        : 'fail';
+      status === 'fail' &&
+        (message = `${otherMember.firstName} ${otherMember.lastName} is not an alumni`);
+    } else if (category === 'courseplan member') {
+      status = otherMember.subteams.includes('courseplan') ? 'pass' : 'fail';
+      status === 'fail' &&
```
andrew032011, line 200: just curious: is this idiomatic? not saying we need to get rid of it. It's possible that I've just not seen this pattern before!

backend/src/API/coffeeChatAPI.ts
```diff
@@ -145,61 +179,135 @@ export const checkMemberMeetsCategory = async (
...
+          message = `${otherMember.firstName} ${otherMember.lastName} is a PM, but on the same team as ${submitter?.firstName} ${submitter?.lastName}`;
+        }
+      }
+    } else if (category === 'a tpm (not your team)') {
+      const isTpm = otherMember.role === 'tpm';
+      const notSameTeam = haveNoCommonSubteams(submitter, otherMember);
+      status = isTpm && notSameTeam ? 'pass' : 'fail';
+      if (status === 'fail') {
+        if (!isTpm && !notSameTeam) {
+          message = `${otherMember.firstName} ${otherMember.lastName} is not a TPM and is on the same team as ${submitter?.firstName} ${submitter?.lastName}`;
```
andrew032011, line 249: similar to above ^:

I feel like we don't need to get that specific here. For a category like this, the person not being a TPM is already enough to justify why the coffee chat category is wrong. I don't think it's necessary to state both reasons

backend/src/API/coffeeChatAPI.ts
```diff
@@ -145,61 +179,135 @@ export const checkMemberMeetsCategory = async (
...
+          diffRole = undefined;
+        }
+
+        // If submitter is a lead, but submitterProperties doesn't exist, status should stay undefined
+        if (diffRole !== undefined) {
+          status = isLead && diffRole;
+
+          if (!status) {
+            if (!isLead && !diffRole) {
+              message = `${otherMember.firstName} ${otherMember.lastName} is not a lead and is from the same role as ${submitter.firstName} ${submitter.lastName}`;
```
andrew032011, line 301: Not being a lead should be enough to disqualify the coffee chat. I probably wouldn't add the part about being in the same role.

backend/src/API/coffeeChatAPI.ts
```diff
@@ -145,61 +179,135 @@ export const checkMemberMeetsCategory = async (
...
+      status = otherMember.subteams.includes('cuapts') ? 'pass' : 'fail';
+      status === 'fail' &&
+        (message = `${otherMember.firstName} ${otherMember.lastName} is not a CUApts member`);
+    } else if (category === 'a pm (not your team)') {
+      const isPm = otherMember.role === 'pm';
+      const notSameTeam = haveNoCommonSubteams(submitter, otherMember);
+      status = isPm && notSameTeam ? 'pass' : 'fail';
+      if (status === 'fail') {
+        if (!isPm && !notSameTeam) {
+          message = `${otherMember.firstName} ${otherMember.lastName} is not a PM and is on the same team as ${submitter?.firstName} ${submitter?.lastName}`;
```
andrew032011, line 236: I feel like we don't need to get that specific here. For a category like this, the person not being a PM is already enough to justify why the coffee chat category is wrong. I don't think it's necessary to state both reasons

backend/src/API/coffeeChatAPI.ts
```diff
@@ -145,61 +179,135 @@ export const checkMemberMeetsCategory = async (
   const haveNoCommonSubteams = (member1: IdolMember, member2: IdolMember): boolean =>
     member2.subteams.every((team) => !member1.subteams.includes(team)) &&
     member1.subteams.every((team) => !member2.subteams.includes(team));
+  let status: MemberMeetsCategoryStatus = 'no data';
```
andrew032011, line 182: I know we're trying to ship this out fast, but I think it would help us bash bugs significantly if we wrote a couple unit tests for some of the cases outlined here. Just enough to convince ourselves that we handled some of the more complicated cases (i.e. PM (not on your team)) properly

backend/src/API/coffeeChatAPI.ts
```diff
@@ -145,61 +179,135 @@ export const checkMemberMeetsCategory = async (
...
+      // If submitterProperties doesn't exist, status should stay undefined
+      if (submitterProperties) {
+        if (category === 'from a different college') {
+          status = otherMemberProperties.college !== submitterProperties.college ? 'pass' : 'fail';
+          status === 'fail' &&
+            (message = `${otherMember.firstName} ${otherMember.lastName} is from the same college as ${submitter.firstName} ${submitter.lastName} (${otherMemberProperties.college})`);
+        }
+      }
+
+      if (category === 'a lead (not your role)') {
```
andrew032011, line 283: I think...we should be able to rule out some people even if they haven't filled out the form. Reason is...some people are clearly not leads, so the warning should pop up anyways. If the lead didn't fill in the form for what kind of lead they are, then it's harder to tell, but say I coffee chat "Patricia Huang". Just because there isn't a form response doesn't mean I shouldn't deduce that the "a lead (not your role)" category would be invalid.

backend/src/API/coffeeChatAPI.ts
```diff
@@ -145,61 +179,135 @@ export const checkMemberMeetsCategory = async (
   const haveNoCommonSubteams = (member1: IdolMember, member2: IdolMember): boolean =>
     member2.subteams.every((team) => !member1.subteams.includes(team)) &&
     member1.subteams.every((team) => !member2.subteams.includes(team));
+  let status: MemberMeetsCategoryStatus = 'no data';
+  let message: string = '';
```
andrew032011, line 183: Can we add a warning for "can't coffee chat yourself" if the person tries submitting themselves?

backend/src/API/coffeeChatAPI.ts
patriciaahuang: Hmm it's probably more idiomatic to use `if`

backend/src/API/coffeeChatAPI.ts
patriciaahuang: Agree! I'll write unit tests for the more complicated cases.

backend/src/API/coffeeChatAPI.ts
andrew032011: How about we add the warning to this function too?

andrew032011 (approved):

backend/src/API/coffeeChatAPI.ts
patriciaahuang: Since the "can't coffee chat yourself" warning appears under "Select a team member" instead of under "Select a category", I think it might be easier to handle this case separately. Also the admin side will never show this warning either, since we don't allow coffee chats with yourself to be submitted.

PR #677, opened 2024-11-09
Coffee Chat: Fix deleteCoffeeChat Bug

patriciaahuang:
### Summary 

`deleteCoffeeChat` didn't allow members to delete their own coffee chat (IdolMember object comparison didn't work). Also someone inputted values in the form, then tried deleting a chat, then pressed submit (the form values stayed), and got application side "cannot read properties of undefined" error, so the state needs to be reset after someone deletes a coffee chat.

### DIFF
--- a/backend/tests/CoffeeChatAPI.test.ts
+++ b/backend/tests/CoffeeChatAPI.test.ts
@@ -11,6 +11,10 @@ import {
 } from '../src/API/coffeeChatAPI';
 import { PermissionError } from '../src/utils/errors';
 
+const user = fakeIdolMember();
+const user2 = fakeIdolMember();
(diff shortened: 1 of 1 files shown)


andrew032011 (approved):

backend/src/API/coffeeChatAPI.ts
```diff
@@ -93,7 +93,7 @@ export const deleteCoffeeChat = async (uuid: string, user: IdolMember): Promise<
 
   if (!coffeeChat) return;
 
-  if (!isLeadOrAdmin && coffeeChat.submitter !== user) {
+  if (!isLeadOrAdmin && coffeeChat.submitter.email !== user.email) {
```
andrew032011, line 96: Would be good to add a test for this.

andrew032011 (approved):

PR #680, opened 2024-11-12
Coffee Chat: Suggestions for Categories with No Submissions

patriciaahuang:
### Summary 

This PR lists all members who meet a category if there is no submission in that category yet. The members listed won't already have a pending/approved coffee chat with the submitter, and won't be the submitter themselves.

Requires addition of `coffee-chat-bingo-board` collection, with documents formatted like 'fall-2024' to store boards from multiple semesters. Each document has format `category: [{netid: __, name: __}]`

### Demo

https://github.com/user-attachments/assets/59b66b68-1dda-43df-a04c-54a589d75a4c

### Notion/Figma Link
https://www.notion.so/Coffee-Chat-Suggestions-for-Categories-with-No-Submissions-13c0ad723ce1804ab3c0d35531f4112b?pvs=4

### DIFF
--- a/backend/src/dao/CoffeeChatDao.ts
+++ b/backend/src/dao/CoffeeChatDao.ts
@@ -178,6 +178,8 @@ export default class CoffeeChatDao extends BaseDao<CoffeeChat, DBCoffeeChat> {
    * Gets coffee chat suggestions for a specifc member
    * @param email - the email of the member
    * @returns A promise that resolves to a CoffeeChatSuggestions object, or undefined.
+   * Note: This data can be stale, so changes made to member or coffee chat info are not immediately reflected.
+   * The data is currently being updated manually.
    */
   static async getCoffeeChatSuggestions(email: string): Promise<CoffeeChatSuggestions | undefined> {
     return coffeeChatSuggestionsCollection
--- a/backend/src/dao/CoffeeChatDao.ts
+++ b/backend/src/dao/CoffeeChatDao.ts
@@ -175,15 +175,13 @@ export default class CoffeeChatDao extends BaseDao<CoffeeChat, DBCoffeeChat> {
   }
 
   /**
-   * Gets coffee chat suggestions for all members, given the semester
-   * @param semester - the semester of the bingo board (i.e. fall-2024).
-   * @returns A promise that resolves to a map of member emails to a BingoBoard object, or undefined.
+   * Gets coffee chat suggestions for a specifc member
+   * @param email - the email of the member
+   * @returns A promise that resolves to a CoffeeChatSuggestions object, or undefined.
    */
-  static async getCoffeeChatSuggestions(
-    semester: string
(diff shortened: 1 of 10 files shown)


andrew032011:
I see this in the deploy preview:
Screenshot 2024-11-11 at 11 20 19 PM

backend/src/dao/CoffeeChatDao.ts
```diff
@@ -172,4 +173,16 @@ export default class CoffeeChatDao extends BaseDao<CoffeeChat, DBCoffeeChat> {
   static async deleteMemberProperties(email: string): Promise<void> {
     memberPropertiesCollection.doc(email).delete();
   }
+
+  /**
+   * Gets a mapping of all categories to corresponding members, given the semester
```
andrew032011, line 178: Maybe add a note here that this data could be up to X number of days stale? So changes made to member info or category info aren't immediately reflected here.

Also, how is this data populated initially? It might be good to have a script that refreshes this data once a day.

andrew032011:
Screenshot 2024-11-24 at 7 40 04 PM
This is what I see in the deploy preview.

andrew032011:
> Screenshot 2024-11-24 at 7 40 04 PM This is what I see in the deploy preview. Is this because the prod DB doesn't have the data populated yet? Maybe get a script together that writes the data, test it on the dev DB, and if you're sure it works (should be safe since it's modifying a collection that's not in use yet), then run that script on the prod DB (let me know if you need help with this last part)?

Also, I think even after you populate the data in the DB, we still might want to change the error handling here. In particular, how about we just don't show any suggestions at all (instead of the loud error) instead of putting up an error modal each time the loading fails?

frontend/src/API/CoffeeChatAPI.ts
```diff
@@ -71,4 +71,18 @@ export default class CoffeeChatAPI {
       return result;
     });
   }
+
+  public static async getCategoryToMembers(semester: string): Promise<BingoBoard> {
+    const res = APIWrapper.get(`${backendURL}/bingo-board/${semester}`).then((res) => res.data);
+    return res.then((val) => {
+      if (val.error) {
```
andrew032011, line 78: We could move this error handling to wherever its used in the UI. I think it should be the UI who gets to decide how to handle this. `getCategoryToMembers` should just return the promise (could be pending/fulfilled/rejected).

The I might format this function the same way the other ones above are, and just return the promise.

andrew032011:
Screenshot 2024-11-26 at 6 12 45 PM
Hm, noticed that "a lead (not your role)" isn't showing any suggestions, even though I'd expect at least a few

patriciaahuang:
> Screenshot 2024-11-26 at 6 12 45 PM Hm, noticed that "a lead (not your role)" isn't showing any suggestions, even though I'd expect at least a few

Fixed!

andrew032011 (approved):

PR #691, opened 2024-11-14
IDOL/New Website: Add Pull from IDOL Images Job

andrew032011:
### Summary 
- Create a script that downloads all images from Storage API (under `images/` folder)
- Attach it to a daily GitHub Actions job
- I didn't think it was practical to set up firebase access from the new-dti-website workspace, so I'm trying a setup with downloading everything from the `backend/` directory, since it already interacts with the storage bucket, then moving the files over to `new-dti-website/public/team`.

### Notion/Figma Link 

https://www.notion.so/IDOL-New-Website-Add-Pull-from-IDOL-Images-Job-13e0ad723ce18054abb0d6a24a99dbf7?pvs=4

### Test Plan 

Locally I've ran the same commands, and it works.
Also, a test PR was generated using the action: https://github.com/cornell-dti/idol/pull/693


cchrischen (approved):
Looks good! Discussed offline:

* There is no current way for information about alumni/inactive members to be retrieved from IDOL. 
  * Do we include all alumni/inactive members or limit it to a certain number of semesters before the current one?
* GitHub does not diff check images that are pulled from this job, meaning it may get noisy. If not, we may need another collection to keep track of dirty images and have logic to check this in the script.
* We should consider changing this job to run on a weekly schedule.

.github/workflows/pull-from-idol-images.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: mkdir backend/temp
+      - name: Pull Images From IDOL
+        env:
+          FIREBASE_PRIVATE_KEY: ${{ secrets.FIREBASE_PRIVATE_KEY }}
+          FIREBASE_PRIVATE_KEY_ID: ${{ secrets.FIREBASE_PRIVATE_KEY_ID }}
+        run: yarn workspace backend run pull-from-idol-images
+      - name: Move Images from backend/temp to new-dti-website/public/team
+        run: mv backend/temp/** new-dti-website/public/team
+      - name: Delete Temp Directory
+        run: rm -r backend/temp
```
henry-li-06, line 42: is there a reason this isn't in the node script? makes it incredibly difficult to follow what's happening :(

backend/scripts/pull-from-idol-images.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable no-console */
+import { writeFile } from 'fs/promises';
+import { join } from 'path';
+import { allMemberImages } from '../src/API/imageAPI';
+
+const main = async () => {
+  console.log('Downloading all member images...');
+  const processedImages = await allMemberImages();
```
henry-li-06, line 8: what about profile images for non-active members? aren't they also gonna get pulled in?

backend/scripts/create-idol-images-pr.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+  const prBody = `
+  ## Summary
+
+  This is a PR auto-generated from running \`yarn workspace backend run pull-from-idol-images\` and
+  \`yarn workspace backend run create-idol-images-pr\`.
+
+  It updates the new-dti-website with latest images from IDOL.
+
+  Please verify that the images look fine on the website and that the image sizes aren't too large.`;
```
henry-li-06, line 44: this in and of itself makes this automation seem not too useful - there's about 0% chance a headshot that someone uploads is the right size (iPhone photo size is way larger than 20KB which is expected). in which case the reviewer has to pull the branch, run the image-cli, push the commit  

and expecting the user to compress the headshot first just seems like a bad UX imo

henry-li-06:
> Do we include all alumni/inactive members or limit it to a certain number of semesters before the current one?

members archive!

andrew032011:
> > Do we include all alumni/inactive members or limit it to a certain number of semesters before the current one?
> 
> members archive!

Yeah we'll probably use that in the automation once we automate this step.

backend/scripts/pull-from-idol-images.ts
andrew032011: Yep!

.github/workflows/pull-from-idol-images.yml
andrew032011: Yeah, I ran into some permissions issues from trying to access the file system outside of `backend/` directory within the node script. So I used this hack as a workaround.

backend/scripts/create-idol-images-pr.ts
andrew032011: We have a follow up item for image compression here!

backend/scripts/pull-from-idol-images.ts
henry-li-06: all of them? riya's image is probably in there still and i'm guessing it isn't planned for hers to be on there

.github/workflows/pull-from-idol-images.yml
henry-li-06: even with a `chmod` for the script?

backend/scripts/pull-from-idol-images.ts
andrew032011: I did a clean-up recently so they're just the same images. I'll add a follow up item to autocleanup that bucket if we no longer need an image, or at the very least doing filtering here.

.github/workflows/pull-from-idol-images.yml
andrew032011: I'll try that!

andrew032011: Filed a ticket: https://www.notion.so/Rewrite-pull-from-IDOL-images-job-with-shell-commands-in-node-script-1470ad723ce1808d92a1c9449ee57b8c?pvs=4

backend/scripts/pull-from-idol-images.ts
andrew032011: https://www.notion.so/Make-Pull-from-IDOL-Images-Script-Only-Download-Necessary-Images-for-Website-1470ad723ce1807ea3fcd11cabb49138?pvs=4

PR #710, opened 2024-11-17
prevent non-admin users from accessing admin pages

kevinmram:
Currently, non-admin IDOL users are able to access admin page (ex. <url>/admin/edit) if they visit the correct URL... 

However, we should redirect non-admins to the home page if this happens!

I tested this by removing myself as an Admin in Firebase, and I could no longer access /admin pages.


frontend/src/pages/_app.tsx
```diff
@@ -135,10 +135,16 @@ const MenuContent: React.FC<{ hasAdminPermission: boolean }> = ({ hasAdminPermis
...
-    router.push('/applicant');
-  }
-
+  useEffect(() => {
+    if (!router.pathname.startsWith('/applicant') && !hasMemberPermissions) {
+      router.push('/applicant');
+      return;
+    }
+    if (router.pathname.startsWith('/admin') && !hasAdminPermission) {
+      router.push('/');
```
JasonMun7, line 146:
```suggestion
    if (router.pathname !== '/') router.push('/');
```
Also maybe adding the return as well to keep consistency.

frontend/src/pages/_app.tsx
```diff
@@ -135,10 +135,16 @@ const MenuContent: React.FC<{ hasAdminPermission: boolean }> = ({ hasAdminPermis
...
   const hasMemberPermissions = useHasMemberPermission();
+  const hasAdminPermission = useHasAdminPermission();
 
-  if (!router.pathname.startsWith('/applicant') && !hasMemberPermissions) {
-    router.push('/applicant');
-  }
-
+  useEffect(() => {
+    if (!router.pathname.startsWith('/applicant') && !hasMemberPermissions) {
+      router.push('/applicant');
```
JasonMun7, line 142:
```suggestion
    if (router.pathname !== '/applicant') router.push('/applicant');
```
This is just to check if there already on the page to prevent the push.

JasonMun7 (approved):
Hey Kevin! This looks great haha! I just added some suggestions! Lmk what you think! Other than that LGTM!

PR #711, opened 2024-11-17
delete shoutout images w/ shoutout

kevinmram:
Previously, if an IDOL user deleted a shoutout that had an image attached to it, the image wouldn't get deleted (taking up space in our storage).

I tested this by submitting shoutouts with images, seeing them get added to the Firebase, and removed when I deleted the shoutout from IDOL.

### DIFF
--- a/backend/src/API/shoutoutAPI.ts
+++ b/backend/src/API/shoutoutAPI.ts
@@ -1,7 +1,7 @@
 import PermissionsManager from '../utils/permissionsManager';
 import { NotFoundError, PermissionError } from '../utils/errors';
 import ShoutoutsDao from '../dao/ShoutoutsDao';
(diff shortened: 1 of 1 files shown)


backend/src/API/shoutoutAPI.ts
```diff
@@ -1,6 +1,7 @@
 import PermissionsManager from '../utils/permissionsManager';
 import { NotFoundError, PermissionError } from '../utils/errors';
 import ShoutoutsDao from '../dao/ShoutoutsDao';
+import { deleteImage } from 'backend/src/API/imageAPI';
```
andrew032011, line 4:
```suggestion
import { deleteImage } from './imageAPI';
```

PR #722, opened 2024-11-20
Website: Update Alumni Section

andrew032011:
### Summary 
Update Alumni section by handpicking all `May 2024` graduates from the archives.

In the future, we would like to automate this process through a semesterly cron job: https://www.notion.so/Website-Automate-Populating-Alumni-4e8ff89fee4e482dacd33acad5f0ed5f?pvs=4

### Notion/Figma Link 

https://www.notion.so/Website-Update-Alumni-Section-1440ad723ce180f0b0eecd683eb123ce?pvs=4

### Test Plan 

https://github.com/user-attachments/assets/87b3ef95-e1f4-4640-aaa7-5c176874e4b7

### Notes 
- The order to determined by `roles.json`, except the roles from each section are concatenated together, and removing duplicate roles (i.e. when role leads show up at the beginning of a role section)

### Breaking Changes


new-dti-website/components/team/MemberDisplay.tsx
```diff
@@ -109,21 +115,17 @@ const MemberDisplay: React.FC = () => {
...
-                        image={`team/${member.netid}.jpg`}
-                        key={member.netid}
-                      />
-                    </a>
-                  ) : (
-                    <></>
-                  )
-                )}
+                {orderedAlumni.map((member, index) => (
+                  <a href={member.linkedin ?? undefined} key={index}>
```
JasonMun7, line 119: I think it would be better to use a 'next/Link' here if we're keeping it consistent! Also I can't remember if there was a reason we didn't want the link to create a new tab? If we did then We could wrap the a tag within a Link as well like: 

`<Link href={member.linkedin ?? '#'}>
  <a target="_blank" rel="noopener noreferrer">
    <MemberCard
      {...member}
      roleDescription={member.roleDescription as RoleDescription}
      cardState={undefined}
      image={`team/${member.netid}.jpg`}
    />
  </a>
</Link>`

Again wondering if @Bookie0 has thoughts on this!

new-dti-website/components/team/MemberDisplay.tsx
```diff
@@ -6,28 +6,34 @@ import teamRoles from './data/roles.json';
...
-      roleName: string;
-      description: string;
-      members: IdolMember[];
-      roles: string[];
-      color: string;
-    };
-  },
-  allMembers
-);
+type RoleEntry = {
```
JasonMun7, line 13: Awesome job making a reusable type for this!

new-dti-website/components/team/MemberDisplay.tsx
```diff
@@ -6,28 +6,34 @@ import teamRoles from './data/roles.json';
...
+
+const roles = populateMembers(teamRoles as RoleEntry, allMembers);
 
 const MemberDisplay: React.FC = () => {
   const [selectedRole, setSelectedRole] = useState<string>('Full Team');
   const [selectedMember, setSelectedMember] = useState<IdolMember | undefined>(undefined);
   const [clickedSection, setClickedSection] = useState<string | undefined>(undefined);
 
+  // Make roles in alumni section the same order as full team
+  const alumniRoles = populateMembers(teamRoles as RoleEntry, alumniMembers as IdolMember[]);
```
JasonMun7, line 31: What are your thoughts on maybe using a **Set** to track members for the duplicate checks instead of using '.includes'? 
 
`const orderedAlumni: IdolMember[] = [];`
`const seenMembers = new Set<IdolMember>();`

JasonMun7 (approved):
This Alumni Section is coming together amazingly haha! Thanks for the manual handpicking! Hopefully we can get that cron job out soon for that! I left some comments below but overall this is starting to look so good! Thanks Andrew!

new-dti-website/components/team/MemberDisplay.tsx
andrew032011: I didn't change this one because I couldn't find a way to "disable" the link behavior for the case that `member.linkedin` is `null` or `undefined`. Using `#` I believe just tries to construct a link `https://new.cornelldti.org/team#`.

So I _think_ (please correct me if I'm wrong) that I'd be forced to write something like this? Is there a better way?
```
                  member.linkedin ? (
                    <Link
                      href={member.linkedin}
                      key={index}
                      target="_blank"
                      rel="noopener noreferrer"
                    >
                      <MemberCard
                        {...member}
                        roleDescription={member.roleDescription as RoleDescription}
                        cardState={undefined}
                        image={`team/${member.netid}.jpg`}
                        key={member.netid}
                      />
                    </Link>
                  ) : (
                    <MemberCard
                      {...member}
                      roleDescription={member.roleDescription as RoleDescription}
                      cardState={undefined}
                      image={`team/${member.netid}.jpg`}
                      key={member.netid}
                    />
                  )
```

andrew032011: I'm on board with opening up a new page in a new tab though!

JasonMun7: You're absolutely right — checking for null or undefined before rendering the link would be necessary! 
Thank you for catching this! You can disregard this comment and sorry for the confusion!

new-dti-website/components/team/MemberDisplay.tsx
andrew032011: I thought about that. My opinion is that the imperative approach plus having to maintain another data structure might be harder to read and also overkill for a section that will only have ~20-40 elements. I usually save such optimizations for larger data sets/backend logic.

PR #726, opened 2024-11-21
Website: Fix Linkedin Links and Add Falsy Check for Links

andrew032011:
### Summary 
Noticed that there's a number of profiles in Alumni page that don't have Linkedin's or have malformed URL's. I fixed the malformed URL's and add all the missing links.

Also fixed how we handle empty links. Right now, a link of `""` would lead to a redirect back to the current page (instead of a no-op).

### Notion/Figma Link 
https://www.notion.so/Website-Fix-Linkedin-Links-and-Fix-Check-for-Empty-Links-1450ad723ce1801a8058eeeaa05ebfd3?pvs=4

### Test Plan 
Manually click through all links on alumni page, and observe that it redirects to their LinkedIn profiles.


new-dti-website/components/team/MemberDisplay.tsx
```diff
@@ -117,7 +117,7 @@ const MemberDisplay: React.FC = () => {
               >
                 {orderedAlumni.map((member, index) => (
                   <a
-                    href={member.linkedin ?? undefined}
+                    href={!member.linkedin ? undefined : member.linkedin}
```
andrew032011, line 120: Previously this was just a nullish check, which handles the case when Linkedin profile field is "null" but not when it's an empty string.

JasonMun7 (approved):
This looks good to me!

PR #728, opened 2024-11-23
created function to fix links in candidate decider

jujulcrane:
### Summary 

I created a function that checks links in Candidate Decider and edits them so that they are proper links. For example, it adds to https:// to a link such as www.linkedin.com/in/andrew-x-chen/ and adds http://www. to linkedin.com/in/andrew-x-chen/. It will also transfrom a linked in username into a link, for example, andrew-x-chen to http://www.linkedin.com/in/andrew-x-chen/. It does the same for github links. It also parses links out of badly formatted text such as "This is not my portfolio but it is close! https://jason-mun-fallback-website.vercel.app/." 

### Notion/Figma Link 
[https://www.notion.so/Candidate-Decider-Make-Link-Reading-More-Robust-1430ad723ce18005b836d0e6adf5732c](url)

### Test Plan 

Here is a screen recording of the testing that I did of various links. There was a strange error at the end of the recording but I don't think its related to the links and it went away on the second try... 

https://github.com/user-attachments/assets/dbd75f37-25e7-4def-8e78-8b7d9107908e

### DIFF
--- a/frontend/src/components/Candidate-Decider/ApplicantCredentials.tsx
+++ b/frontend/src/components/Candidate-Decider/ApplicantCredentials.tsx
@@ -12,6 +12,55 @@ type Props = {
   preferredName?: string;
 };
 
+const fixLink = (link: string, git: boolean, linkedIn: boolean): string | undefined => {
+  if (!link) {
+    return undefined;
+  }
+
+  const urlRegex =
+    /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g;
+  const matches = link.match(urlRegex);
+
+  let extractedLink = matches ? matches[0].trim() : link.trim();
+
+  let fixedLink: string = '';
+
+  if (!extractedLink.startsWith('https://') && extractedLink.startsWith('http://')) {
+    extractedLink = extractedLink.replace('http://', 'https://');
+  }
+
+  if (extractedLink.includes('https://')) {
+    if (!extractedLink.startsWith('https://')) {
+      const index = extractedLink.indexOf('https://');
+      fixedLink = extractedLink.substring(index);
+    } else {
+      fixedLink = extractedLink;
+    }
+  } else if (linkedIn) {
+    if (extractedLink.startsWith('www.')) {
+      const index = extractedLink.indexOf('https://');
+      fixedLink = 'https://' + extractedLink.substring(index);
+    } else if (extractedLink.startsWith('linkedin.com')) {
+      const index = extractedLink.indexOf('https://');
+      fixedLink = 'https://www.' + extractedLink.substring(index);
+    } else if (!extractedLink.includes('linkedin.com/in')) {
+      fixedLink = `https://www.linkedin.com/in/${extractedLink}/`;
+    } else {
+      fixedLink = extractedLink;
+    }
+  } else if (git) {
+    if (extractedLink.startsWith('github.com')) {
+      fixedLink = 'https://' + extractedLink;
+    } else if (!extractedLink.includes('github.com')) {
+      fixedLink = 'https://github.com/' + extractedLink;
+    }
+  } else {
+    fixedLink = extractedLink;
+  }
+
+  return fixedLink;
+};
+
 const ApplicantCredentials: React.FC<Props> = ({
   name,
   email,
@@ -29,21 +78,21 @@ const ApplicantCredentials: React.FC<Props> = ({
     <p>{email}</p>
     <p>Class of {gradYear}</p>
     <div className={styles.iconsContainer}>
-      <a className={styles.icon} href={resumeURL}>
+      <a className={styles.icon} href={fixLink(resumeURL, false, false)}>
         <FileIcon />
       </a>
       {githubURL && (
-        <a className={styles.icon} href={githubURL}>
+        <a className={styles.icon} href={fixLink(githubURL, true, false)}>
           <GithubIcon />
         </a>
       )}
       {linkedinURL && (
-        <a className={styles.icon} href={linkedinURL}>
+        <a className={styles.icon} href={fixLink(linkedinURL, false, true)}>
           <LinkedinIcon />
         </a>
       )}
       {portfolioURL && (
-        <a className={styles.icon} href={portfolioURL}>
+        <a className={styles.icon} href={fixLink(portfolioURL, false, false)}>
           <GlobeIcon />
         </a>
       )}


frontend/src/components/Candidate-Decider/ApplicantCredentials.tsx
```diff
@@ -12,6 +12,55 @@ type Props = {
...
+
+  const urlRegex =
+    /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_+.~#?&//=]*)/g;
+  const matches = link.match(urlRegex);
+
+  let extractedLink = matches ? matches[0].trim() : link.trim();
+
+  let fixedLink: string = '';
+
+  if (!extractedLink.startsWith('https://') && extractedLink.startsWith('http://')) {
```
andrew032011, line 28: I think `https://` vs. `http://` is not an issue we need to fix, so we could remove this

frontend/src/components/Candidate-Decider/ApplicantCredentials.tsx
```diff
@@ -12,6 +12,55 @@ type Props = {
...
+    extractedLink = extractedLink.replace('http://', 'https://');
+  }
+
+  if (extractedLink.includes('https://')) {
```
andrew032011, line 32: I would expect that at this point, I know I'm working with some type of link. The regex + trim should've extracted out the link already, so I don't see a scenario where there would be "https://" in the string but not at the beginning. Let me know if I'm missing something though!

frontend/src/components/Candidate-Decider/ApplicantCredentials.tsx
```diff
@@ -12,6 +12,55 @@ type Props = {
...
+    if (!extractedLink.startsWith('https://')) {
+      const index = extractedLink.indexOf('https://');
+      fixedLink = extractedLink.substring(index);
+    } else {
+      fixedLink = extractedLink;
+    }
+  } else if (linkedIn) {
+    if (extractedLink.startsWith('www.')) {
+      const index = extractedLink.indexOf('https://');
```
andrew032011, line 41: If it starts with `www.`, then I wouldn't expect the string to have `https://`. Shouldn't we just prefix `extractedLink` with `https://`? Similar questions with the below logic

andrew032011 (commented):
I would move this function into `frontend/src/utils.ts`. We will probably use it in a lot of places.

It feels to me that there are a few extra cases and a few cases unhandled here (i.e. `www.github.com/...`). But it'll be easier to convince yourself of this if you wrote some tests showing all the different/exhaustive scenarios!

I would also put a docstring for this function too, especially since other people will want to use this a lot.

frontend/src/components/Candidate-Decider/ApplicantCredentials.tsx
jujulcrane: oh wait yeah ill fix that

andrew032011:
> I would move this function into `frontend/src/utils.ts`. We will probably use it in a lot of places.
> 
> It feels to me that there are a few extra cases and a few cases unhandled here (i.e. `www.github.com/...`). But it'll be easier to convince yourself of this if you wrote some tests showing all the different/exhaustive scenarios!
> 
> I would also put a docstring for this function too, especially since other people will want to use this a lot.

Feel free to re-request review from me once you've addressed these comments too! ^^

henry-li-06:
> > I would move this function into `frontend/src/utils.ts`. We will probably use it in a lot of places.
> > It feels to me that there are a few extra cases and a few cases unhandled here (i.e. `www.github.com/...`). But it'll be easier to convince yourself of this if you wrote some tests showing all the different/exhaustive scenarios!
> > I would also put a docstring for this function too, especially since other people will want to use this a lot.
> 
> Feel free to re-request review from me once you've addressed these comments too! ^^

what if this was just done in parsing the csv to begin with 👀 either FE or BE

andrew032011:
> > > I would move this function into `frontend/src/utils.ts`. We will probably use it in a lot of places.
> > > It feels to me that there are a few extra cases and a few cases unhandled here (i.e. `www.github.com/...`). But it'll be easier to convince yourself of this if you wrote some tests showing all the different/exhaustive scenarios!
> > > I would also put a docstring for this function too, especially since other people will want to use this a lot.
> > 
> > 
> > Feel free to re-request review from me once you've addressed these comments too! ^^
> 
> what if this was just done in parsing the csv to begin with 👀 either FE or BE
I still want to show the original string here just in case something goes wrong with the string parsing! So I'd rather we keep it and just format it for the profile card.

I would put it in the FE so we have the BE as a source of truth of what the original value was.

jujulcrane:
I added some more cases that it should handle now. Let me know if there are any more cases that I'm missing! Here are the current tests: 
Screenshot 2024-11-23 at 3 38 03 PM

andrew032011:
> I added some more cases that it should handle now. Let me know if there are any more cases that I'm missing! Here are the current tests: Screenshot 2024-11-23 at 3 38 03 PM

I'd move those into a test file! In the same directory, you can create a file `utils.tests.ts` and use Jest to write your unit tests! See any of our tests under the [tests folder](https://github.com/cornell-dti/idol/tree/e6b082a57fab1e1e4b414934c169fb5ce5d2caf6/backend/tests) for examples. You won't need to do any mocking here or really anything too fancy. Just checking inputs and expected outputs. 

And you can run 
```
yarn test <path to testing file, e.g. backend/tests/MembersDao.test.ts>
```
to run the tests in the file.

Feel free to Slack me if you need help setting it up!

frontend/src/utils.ts
```diff
@@ -116,3 +116,85 @@ export class Emitters {
...
+ *
+ * @example
+ * fixLink("Visit my profiles: https://linkedin.com/in/jujulcrane and github.com/jujulcrane", false, true);
+ * // Returns: undefined (multiple links in the input)
+ *
+ * @example
+ * fixLink("jujulcrane", false, true);
+ * // Returns: "https://www.linkedin.com/in/jujulcrane/"
+ */
+export const fixLink = (link: string, git: boolean, linkedIn: boolean): string | undefined => {
```
andrew032011, line 145: Maybe instead of booleans, we could change this into just a parameter ("linkType")? That way if we want to support more links, we won't have to add an extra boolean for each one (also I think it'll make it easier to read). You can just make a `type LinkType = "git" | "linkedIn"` and use that type here.

Also super small/opinionated nit but I think `formatLink` might be a better name here. "fix" feels like it's fixing a bug.

frontend/src/utils.ts
```diff
@@ -116,3 +116,85 @@ export class Emitters {
...
+    }
+    if (git) {
+      return `https://github.com/${link.trim().toLowerCase()}`;
+    }
+    return undefined;
+  }
+
+  const extractedLink = matches[0]
+    .trim()
+    .replace(/[.,)]+$/, '')
```
andrew032011, line 170: out of curiosity what does this do? Maybe add a comment.

frontend/src/utils.ts
```diff
@@ -116,3 +116,85 @@ export class Emitters {
...
+    .toLowerCase();
+
+  if (linkedIn) {
+    if (extractedLink.startsWith('www.')) {
+      return `https://${extractedLink}`;
+    }
+    if (extractedLink.startsWith('linkedin.com')) {
+      return `https://www.${extractedLink}`;
+    }
+    if (!extractedLink.includes('linkedin.com/in')) {
```
andrew032011, line 180: This feels `https://www.linkedin.com/in/${extractedLink}` case feels like a potential duplicate with the logic on lines 158-166 same with the portion below. Correct me if I'm wrong though! If it's a duplicate I think we could just remove it here.

frontend/src/utils.ts
```diff
@@ -116,3 +116,85 @@ export class Emitters {
...
+    }
+    return undefined;
+  }
+
+  const extractedLink = matches[0]
+    .trim()
+    .replace(/[.,)]+$/, '')
+    .toLowerCase();
+
+  if (linkedIn) {
```
andrew032011, line 173: This whole bottom section feels like it doesn't need to split into the LinkedIn and github links tbh. Since the only case where it matters which link type it is is if they just give us a username, which is handled above. Otherwise, I think the logic here can be as simple as checking if it's prefixed by `https://`, and if not, adding that. Do let me know if I'm missing a case though!

andrew032011 (commented):
Feel free to re-request review once you're done making the changes we talked about!

frontend/src/utils.ts
```diff
@@ -116,3 +116,69 @@ export class Emitters {
...
+/**
+ * Parses links from english text and fixes a provided link to ensure it adheres to a proper format for LinkedIn or GitHub profiles.
+ * Also handles raw usernames (e.g., "jujulcrane") by constructing a proper URL.
+ *
+ * @param {string} link - The input link, which could be a full URL, partial URL, or raw username.
+ * @param {LinkType} linkType - The type of link if either GitHub or LinkedIn.
+ * @returns {string | undefined} - The fixed and properly formatted link or undefined if the input is invalid.
+ *
+ * @example
+ * formatLink("https://linkedin.com/in/jujulcrane", linkedin);
```
andrew032011, line 130:
```suggestion
 * formatLink("https://linkedin.com/in/jujulcrane", "linkedin");
```

frontend/src/utils.ts
```diff
@@ -116,3 +116,69 @@ export class Emitters {
...
+ * // Returns: "https://www.linkedin.com/in/jujulcrane"
+ *
+ * @example
+ * formatLink("github.com/jujulcrane", github);
```
andrew032011, line 134:
```suggestion
 * formatLink("github.com/jujulcrane", "github");
```

frontend/src/utils.ts
```diff
@@ -116,3 +116,69 @@ export class Emitters {
...
+ * // Returns: "https://github.com/jujulcrane"
+ *
+ * @example
+ * foramtLink("Visit my profiles: https://linkedin.com/in/jujulcrane and https://github.com/jujulcrane", linkedin);
```
andrew032011, line 138:
```suggestion
 * formatLink("Visit my profiles: https://linkedin.com/in/jujulcrane and https://github.com/jujulcrane", "linkedin");
```

frontend/src/utils.ts
```diff
@@ -116,3 +116,69 @@ export class Emitters {
...
+ * // Returns: undefined (multiple links in the input)
+ *
+ * @example
+ * formatLink("jujulcrane", linkedin);
```
andrew032011, line 142:
```suggestion
 * formatLink("jujulcrane", "linkedin");
```

andrew032011 (approved):
Excellent work!!

PR #731, opened 2024-11-23
Website: Make All Interactive Elements Buttons

cchrischen:
### Summary 
Change all buttons that are interactive, i.e. some functionality that does not redirect the user, to use button elements. 
* All icons: Home carousel, Team carousel, Team roles, Sponsors medals
* Apply tabs
* Navbar hamburger and close icons
* Member cards 
* Accordions on course and apply

Also fixed z-index issues with the apply banner being over the navbar on small viewports.

### Notion/Figma Link 
[Notion](https://www.notion.so/Website-Make-Anything-Clickable-a-Button-1440ad723ce180a29baaf8ad535f6caf)

### Test Plan 
Ensuring that tabbing through the page highlights all clickable elements.

### DIFF
--- a/new-dti-website/components/apply/ApplicationTimeline.tsx
+++ b/new-dti-website/components/apply/ApplicationTimeline.tsx
@@ -16,14 +16,14 @@ type TabProps = {
 };
 
 const Tab: React.FC<TabProps> = ({ isSelected, text, onClick }) => (
-  <div
+  <button
     className={`flex items-center lg:px-5 lg:py-4 md:px-4 md:py-3 xs:px-2 md:rounded-t-xl xs:rounded-t-lg ${
       isSelected ? 'bg-[#FEFEFE] text-[#A52424]' : 'bg-[#7E2222CC] text-[#FEFEFE]'
     } hover:cursor-pointer md:h-min xs:h-full`}
     onClick={onClick}
   >
     <p className="font-bold lg:text-lg md:text-[13px] xs:text-[10px]">{text}</p>
-  </div>
+  </button>
 );
 
 type DateTime = {
--- a/new-dti-website/components/apply/ApplicationTimeline.tsx
+++ b/new-dti-website/components/apply/ApplicationTimeline.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import { KeyboardEvent, useEffect, useRef, useState } from 'react';
 import Image from 'next/image';
 import Link from 'next/link';
 import config from '../../config.json';
@@ -12,17 +12,23 @@ import { extractEndDate, extractEndTime, parseDate } from '../../src/utils/dateU
 type TabProps = {
   isSelected: boolean;
   text: string;
+  tabIndex?: number;
   onClick?: () => void;
 };
 
-const Tab: React.FC<TabProps> = ({ isSelected, text, onClick }) => (
+const Tab: React.FC<TabProps> = ({ isSelected, text, onClick, tabIndex }) => (
   <button
     className={`flex items-center lg:px-5 lg:py-4 md:px-4 md:py-3 xs:px-2 md:rounded-t-xl xs:rounded-t-lg ${
       isSelected ? 'bg-[#FEFEFE] text-[#A52424]' : 'bg-[#7E2222CC] text-[#FEFEFE]'
     } hover:cursor-pointer md:h-min xs:h-full`}
     onClick={onClick}
+    role="tab"
+    tabIndex={isSelected ? 0 : -1}
+    aria-selected={isSelected}
   >
-    <p className="font-bold lg:text-lg md:text-[13px] xs:text-[10px]">{text}</p>
+    <p className="font-bold lg:text-lg md:text-[13px] xs:text-[10px]" role="tabpanel">
+      {text}
+    </p>
   </button>
 );
 
@@ -221,6 +227,12 @@ const ApplicationTimeline = () => {
   const scrollToIndex =
     nextEventIndex === sortedEvents.length ? nextEventIndex - 1 : nextEventIndex;
 
+  const handleKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
+    if (e.key === 'ArrowLeft' || e.key === 'ArrowRight') {
+      setCycle(cycle === 'freshmen' ? 'upperclassmen' : 'freshmen');
+    }
+  };
+
   useEffect(() => {
     if (timelineRef.current && selectedNodeRef.current && width >= TABLET_BREAKPOINT) {
       const innerDiv = selectedNodeRef.current.getBoundingClientRect().top;
@@ -247,7 +259,7 @@ const ApplicationTimeline = () => {
             >
               cornell-dti/timeline
             </p>
-            <div className="flex items-end">
+            <div className="flex items-end" role="tablist" onKeyDown={handleKeyDown}>
               {isFall ? (
                 <>
                   <Tab
--- a/new-dti-website/components/course/DDProjects.tsx
+++ b/new-dti-website/components/course/DDProjects.tsx
@@ -30,8 +30,8 @@ export default function DDProjects({ title, description, imageSrc }: DDProjectsP
   };
 
   return (
-    <div
-      className={`transition-all duration-300 ease-in-out ${
+    <button
+      className={`transition-all duration-300 ease-in-out text-left ${
         isOpen ? 'bg-red-500' : 'bg-white'
       } w-full max-w-8xl rounded-xl drop-shadow-sm px-10 py-8 border-1 border-[#E4E4E4]`}
       onClick={toggleCard}
@@ -58,6 +58,6 @@ export default function DDProjects({ title, description, imageSrc }: DDProjectsP
           <img src={imageSrc} alt={title} className="mt-4 w-full h-48 object-cover rounded-lg" />
         </div>
       </div>
-    </div>
+    </button>
   );
 }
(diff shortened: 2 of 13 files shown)


andrew032011 (commented):
The current changes look good. I pulled the branch and looked for instances of `onClick` and still found some components (One of them being on the Apply page, Role descriptions) that are using `onClick` on non-button components (i.e. a `div`).

new-dti-website/components/team/TeamHero.tsx
```diff
@@ -161,7 +159,7 @@ const TeamHero = () => {
                     index === carouselIndex % carouselLength ? '' : 'opacity-50'
                   }`}
                 >
-                  <div
+                  <button
```
Bookie0, line 162: So one issue i found when using `tab` to test these changes is that when you open the image, you cannot use `tab` to navigate through the X close icon and left/right chevron icons. I think when the image is open in this larger view, we should treat it as a modal and so lock the focus to only those 3 elements (X close btn, left chevron, and right chevron). We could also add the ability to use `esc` to close the image, too.

Currently, as you can see in the video, once I open the image in the large view, pressing tab changes the to the next image. But I think we should be reserving the focus state for those 3 elements


https://github.com/user-attachments/assets/84b13a2a-d899-4347-881c-641976ceb685

new-dti-website/components/team/MemberDisplay.tsx
```diff
@@ -67,22 +67,25 @@ const MemberDisplay: React.FC = () => {
...
-                  width={role.width}
-                  height={role.height}
-                  className={`lg:h-[66px] xs:h-[50px] w-auto ${
-                    selectedRole === role.altText ? 'scale-125' : ''
-                  } hover:scale-125`}
-                />
+                >
+                  <Icon
+                    icon={`${role.src}_base.svg`}
+                    hoverIcon={`${role.src}_sticker.svg`}
```
Bookie0, line 78: I don't see any interactivity on the table FYI

sponsors medlas

new-dti-website/components/course/DDProjects.tsx
```diff
@@ -30,8 +30,8 @@ export default function DDProjects({ title, description, imageSrc }: DDProjectsP
   };
 
   return (
-    <div
-      className={`transition-all duration-300 ease-in-out ${
+    <button
```
Bookie0, line 33: I've noticed that these FAQ accordion elements actually have two focus states: one for the whole element and one for the icon. I think the icon should be purely visual (ie: you shouldn't be able to interact/focus on it). 


 
accordgion focus state
inner focus state


Nested buttons is also typically bad practice because it's not semantic HTML and it can mess up screen readers, so i think these are other reasons we should remove the inner `<button>` element containing the icon.

new-dti-website/components/team/MemberGroup.tsx
```diff
@@ -62,17 +64,18 @@ type MemberCardProps = {
...
+      id="memberCard"
+      className={`w-fit md:p-3 md:pb-4 xs:p-2 xs:pb-3 h-fit justify-self-center ${
+        props.cardState ? 'opacity-70 hover:opacity-100' : 'opacity-100'
+      } ${
+        props.cardState === 0 && 'shadow-[0_4px_4px_0_#00000040]'
+      } hover:shadow-[0_4px_4px_0_#00000040] cursor-pointer`}
+    >
+      <MemberSummary {...props} enlarged={false} />
+    </Card>
+  </button>
```
Bookie0, line 78: I can use `tab` to navigate between the different team member cards, but I cannot use my keyboard (`enter` or `space`) to open the card. Instead I have to use my mouse. The cards should be keyboard accessible

new-dti-website/components/apply/ApplicationTimeline.tsx
```diff
@@ -16,14 +16,14 @@ type TabProps = {
...
-  <div
+  <button
     className={`flex items-center lg:px-5 lg:py-4 md:px-4 md:py-3 xs:px-2 md:rounded-t-xl xs:rounded-t-lg ${
       isSelected ? 'bg-[#FEFEFE] text-[#A52424]' : 'bg-[#7E2222CC] text-[#FEFEFE]'
     } hover:cursor-pointer md:h-min xs:h-full`}
     onClick={onClick}
   >
     <p className="font-bold lg:text-lg md:text-[13px] xs:text-[10px]">{text}</p>
-  </div>
+  </button>
```
Bookie0, line 26: since these are tabs, I think we should mention their role using the ARIA `tab` role. It helps screen readers clearly label what this element is.

So you can add `role="tab"` to the `<button>` elements, and then `role="tabpanel"` to the actual tab panel contents.

[MDN has good documentation for this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role). And make sure to check out the ["Keyboard interactions" section](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#keyboard_interactions) for the correct behavior for the keyboard :)

new-dti-website/components/navbar.tsx
```diff
@@ -70,14 +70,15 @@ const Navbar: React.FC = () => {
...
-            />
+            <button onClick={handleMenuClick}>
+              <Image
+                className="h-12 w-auto md:h-14"
+                src="/icons/hamburger_icon.svg"
+                width={56.5}
+                height={56.5}
+                alt="Hamburger Menu Icon"
+              />
+            </button>
```
Bookie0, line 81: For icon buttons like this one, we should be adding `aria-label` attribute for accessibility. It helps screen reader technologies announce to the user what this icon button does (since it doesn't have a text label).

[More on `aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label)

When you add `aria-label`, I think we won't need the alt text on `<Image>` anymore since that would be redundant. However, make sure to still include the empty `alt=""` attribute to explicitly tell screen readers that this image is decorative and can be safely skipped. (If we forget to add the attribute, even if we don't technically need it, the accessibility score for the website is lowered and can be flagged on automated testing tools)

Bookie0 (changes requested):
Hi Chris! Thank you for taking the time to address these accessibility issues!! 

This is a great start! I've left some comments with some additional feedback and some ways to move forward. Please Slack me if you have any questions! :)

new-dti-website/components/icons.tsx
```diff
@@ -9,7 +9,7 @@ interface IconProps {
   activeIcon: string;
   altText: string;
   isActive: boolean;
-  onClick: () => void;
+  onClick?: () => void;
```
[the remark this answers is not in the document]
Bookie0: In general, I think we should be using `<button>` for all interactive content.

However, I believe sometimes there might be exceptions. 

Off the top of my head, for example, imagine we have an interactive card with a link inside. From a user's standpoint, they would expect that clicking anywhere on the card (except on the link) triggers an action, and then clicking the link triggers another action. 

We might want to avoid using `<button>` to wrap the card a since there would be a nested `<a>` link inside. Using `<button>` with `<a>` inside of it would be confusing to screenreader technologies and not very accessible.

In this case, I see 2 solutions.

The first would involve quite some CSS. So we'd use something like a regular `<div>` or `<article>`. And then we could have a hidden `<button>` element that we'd style using CSS (with `::after`) to cover the entire card. Then with some `z-index` trickery, we'd move `<a>` on top. The reason we do all of this is to try and keep the `onClick` on the `<button>` element because it is semantic, natively focusable to keyboard keys, and helps screenreader technologies. But this is quite complicated.

Another solution I'd propose would be to just use that regular `<div>` tag, and then apply the `role="button"` to that `<div>`. Then you could combine it with `tabindex="0"` and the `onClick` attribute and basically have yourself a `<button>`. However, we should be wary of this approach because `role="button"` tells the screenreader that the element is a button, but it doesn't provide any actual functionality. Again, this goes back to my point that using a `<button>` is almost always the best case.

So yeah TL;DR, let's try and avoid using `onClick` on an element that isn't a `<button>`. It's simpler, more accessible, and easier to maintain.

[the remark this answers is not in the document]
Bookie0: hope this helps :)

new-dti-website/components/team/MemberDisplay.tsx
cchrischen: The medals become clickable on small viewports.

JasonMun7 (changes requested):
Hey Chris! The changes are looking great so far! Honestly, now that I’m looking at it, why don’t we just extend these updates to the `components/ui/button` as well? We could add the accessibility improvements like `aria-labels` and any necessary `roles` that @Bookie0 is mentioning about directly into the button component? This would future-proof it and make sure all buttons across the app are consistent and so we can directly add the button classes in`globalcss` as variants as well! 

The button component already provides a great skeleton with its variant and size options, while also supporting React children, so it seems like a natural place to centralize these enhancements!

What do you think? Happy to help explore this if you think it’s a good direction! I'm also completely fine with keeping it as is but open to hear what other's think about that as well.

new-dti-website/components/ui/button.tsx
```diff
@@ -35,14 +35,20 @@ const buttonVariants = cva(
...
   asChild?: boolean;
 }
 
 const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
-  ({ className, variant, size, asChild = false, ...props }, ref) => {
+  ({ className, variant, size, asChild = false, ariaLabel, ...props }, ref) => {
     const Comp = asChild ? Slot : 'button';
     return (
-      <Comp className={cn(buttonVariants({ variant, size, className }))} ref={ref} {...props} />
+      <Comp
```
JasonMun7, line 46: Awesome this looks great! Thanks for doing this! After we get this in I can add the variants to this so we can omit the button styles in the global css! Thanks again Chris!

Also just for future reference I think it'd be best to just use this component now instead of the button tag! I don't think we have to do that for this PR though!

JasonMun7 (approved):

Bookie0 (approved):
nice! thanks for the changes. just a few little nits and we should be good to go

PR #732, opened 2024-11-25
Coffee Chat: Remove Live Checking and Update Chats After Running Autochecker

patriciaahuang:
### Summary 

On the admin side, the color of each chat was being fetched live, which caused some lagging. The chats were also not being updated automatically after clicking "Re-run autochecker."


andrew032011:
As a sanity check, I did some manual testing, and I found that the state still lags incorrectly after I run the autochecker. Can you check if you're potentially using the old values in the state after the autochecker is run? I have to refresh before I see the results of the autochecker updated.

https://github.com/user-attachments/assets/276b027a-df88-4832-883d-8b232dee3b70

PR #733, opened 2024-11-26
IDOL: Add Semester Joined and College to User Profile

cchrischen:
### Summary 
This PR adds semester joined and college to the user profile. Semester joined is used to keep track of who is a newbie, for coffee chat requirement purposes. College is used on the new website on the pie chart. Semester joined is visible but not editable on the user side -- it should be configured by admin. College is editable on the user side. 

### Notion/Figma Link 
[Notion](https://www.notion.so/Website-IDOL-Add-Semester-Joined-and-College-Fields-to-IdolMember-1440ad723ce18056b66ec5436eb56959)

### Test Plan 
Editing the fields, verifying values persisted in database.

https://github.com/user-attachments/assets/70723fef-cb8e-48c4-9ba8-e6373739b185

### Notes 
* Semester joined and college have not been populated in the database.
* Though I have marked both fields as required, they can be undefined for now for backward compatibility.


common-types/index.d.ts
```diff
@@ -39,14 +39,19 @@ type RoleDescription =
...
+type College = 'Eng' | 'CAS' | 'CALS' | 'Dyson' | 'HumEc' | 'Hotel' | 'ILR' | 'Brooks';
+
 /** The data type used by IDOL to represent a DTI member. */
 interface IdolMember {
   readonly netid: string;
   readonly email: string;
   readonly firstName: string;
   readonly lastName: string;
   readonly pronouns: string;
+  readonly joined?: string;
```
andrew032011, line 52: how about `semesterJoined`?

common-types/index.d.ts
```diff
@@ -39,14 +39,19 @@ type RoleDescription =
   | 'Design Advisor'
   | 'Business Advisor';
 
+/** The possible colleges an IDOL member could be. */
+type College = 'Eng' | 'CAS' | 'CALS' | 'Dyson' | 'HumEc' | 'Hotel' | 'ILR' | 'Brooks';
```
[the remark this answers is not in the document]
andrew032011: How about `semesterJoined`?

andrew032011 (approved):

andrew032011:
> Semester joined and college have not been populated in the database.

@cchrischen The order here (to prevent downtime on the edit profile page) is to populate `semesterJoined` field before you merge this in. Either through a script or it might actually be easier to use the "Upload CSV" tool (I can help with this since the tool is a little buggy right now and not super easy to work with: https://www.notion.so/Edit-User-Improvements-1180ad723ce180a88ed2e562ca0ee4ad?pvs=4).

frontend/src/components/Admin/AddUser/AddUser.tsx
```diff
@@ -152,6 +154,14 @@ export default function AddUser(): JSX.Element {
       return;
     }
 
+    // Check that college is selected
+    if (!member.college) {
+      Emitters.generalError.emit({
+        headerMsg: 'College not selected!',
+        contentMsg: 'Please select college from the dropdown.'
+      });
```
andrew032011, line 162: Make sure to `return` here, right now it still lets me submit even though I'm missing this field. That said we will likely need to take this away for now since no one has their college populated yet. So it's going to show an error every time an admin tries to save the user, if the college isn't populated yet

andrew032011: also, if we're going to require `Semester Joined`, we should require it here too so admins don't forget. In fact, if you don't, members aren't going to be able to save profiles because the semester joined field won't be populated (and they can't fix it because the field is disabled on the member side)

andrew032011 (commented):
actually i found a few items we have to address

PR #735, opened 2024-11-26
Website: Remove animations and add Media Reduce Checks

JasonMun7:
### Summary 

Removed the animations from the Courses/Apply page and added reduce motion checks throughout the website. 

- [x] Removed fade-in animations from the course/Apply page (Ticket 2)
- [x] Implemented `prefers-reduced-motion` adjustments to hide animations when users enable reduced motion settings (Ticket 1)
  - Utilized Tailwind's `reduce-motion` and `motion-safe` utilities for CSS-based animations
  - Created a `useMediaReduce` hook for handling JavaScript-based animations
  - Added pause and play buttons for horizontal carousels on the Team and Product pages
  - Developed a reusable `useCarouselControls` hook for future carousel implementations
- [x] Added documentation for the Banner component

### Test Plan 

1. **Course Page**:
   - Verified fade-in animations are removed, and content loads directly.
   - Also just made the opening animations of the past projects section open a bit slower on isOpen
2. **Reduced Motion Settings**:
   - Tested with `prefers-reduced-motion` enabled and confirmed animations are hidden.
   - Verified animations play as expected when the setting is disabled.
   - On screen resize, it sets the carousel to pause until it's played again by using the pause/play button
   
### Example 
https://github.com/user-attachments/assets/4e4d2d70-fcca-4d9e-bc32-454d014e63ce


new-dti-website/components/apply/Banner.tsx
```diff
@@ -14,6 +14,23 @@ const variantStyles: Record<NonNullable<BannerProps['variant']>, string> = {
...
+ *
+ * @remarks
+ * The `Banner` component shows a message in a banner that can be styled and positioned
+ * based on the user's scroll position. It supports three variants (`info`, `alert`, and `success`), which can be modified
+ * (I just used these variant names because they're the most common banner usages),
+ * and allows additional customizations via className.
+ *
+ * @param props - Contains:
+ *   - `message` (string): The message to be displayed in the banner. *(Required)*
+ *   - `variant` (`'info' | 'alert' | 'success'`, optional): Specifies the banner's style. Defaults to `'info'`.
```
Bookie0, line 28: i would rename the variants because i think they're a bit vague and don't exactly correspond to the actual visual style of the banner.

I'd do this:

|Image|Current|Proposed|
|-|-|-|
|Screenshot 2024-11-26 at 10 09 22 AM|`alert`|`accent`|
|Screenshot 2024-11-26 at 10 08 38 AM|`info`|`default`|
|Screenshot 2024-11-26 at 10 08 58 AM|`success`|`outlined`|

new-dti-website/components/products/imageCarousel.tsx
```diff
@@ -63,6 +64,20 @@ const ImageCarousel = (props: { items: carouselItem[] }) => {
...
+          aria-label={isPlaying ? 'Pause carousel' : 'Play carousel'}
+        >
+          <Image
+            src={isPlaying ? '/icons/pause.svg' : '/icons/play.svg'}
+            alt={isPlaying ? 'Pause' : 'Play'}
+            width={24}
+            height={24}
+          />
+        </button>
+      )}
```
Bookie0, line 80: honestly i think we should show this play/pause icon button regardless of whether you have Reduce Motion settings toggled on or off. 

I think we should do what [apple.com](https://www.apple.com/) does at the bottom of their home page with the Apple TV/Apple Arcade/etc. carousels:
- They always show the play/pause icon button
- If you have Reduce Motion on, the carousel does _not_ animate by default
- If you have Reduce Motion off, the carousel does animate by default (but you can easily pause it)

new-dti-website/components/team/TeamHero.tsx
```diff
@@ -184,6 +190,20 @@ const TeamHero = () => {
...
+          aria-label={isPlaying ? 'Pause carousel' : 'Play carousel'}
+        >
+          <Image
+            src={isPlaying ? '/icons/pause.svg' : '/icons/play.svg'}
+            alt={isPlaying ? 'Pause' : 'Play'}
+            width={24}
+            height={24}
+          />
+        </button>
+      )}
```
Bookie0, line 206: Same comment as this.

new-dti-website/components/ui/carousel/useCarouselControls.ts
```diff
@@ -0,0 +1,56 @@
+import { useEffect, useRef, useState } from 'react';
+import Autoplay from 'embla-carousel-autoplay';
+import type { CarouselApi } from './carousel';
+
```
Bookie0, line 4: i also agree with @andrew032011 that the implementation of the carousels on the Product and About pages are a bit choppy.

Here's what it looks like for me:

https://github.com/user-attachments/assets/619207cf-baf6-466a-97e1-00a26e480ac3

The animation is very slow, and the image doesn't completely fit inside the white border before it scrolls again

Bookie0 (changes requested):
Thank you for these changes! It's definitely a step in the good direction.

Removing the fade-in transition and adding the play/pause icon button is great.

However it seems like the behavior with the carousels (specifically the timing and animations) has been slightly broken.

Bookie0 (approved):
Cool! i think the carousel on the Team page is smoother now. 

I still do think the carousel on the Products page is still slightly rough though (see @andrew032011's [comment here](https://github.com/cornell-dti/idol/pull/735#issuecomment-2499797971))

JasonMun7:
> Cool! i think the carousel on the Team page is smoother now.
> 
> I still do think the carousel on the Products page is still slightly rough though (see @andrew032011's [comment here](https://github.com/cornell-dti/idol/pull/735#issuecomment-2499797971))
> 
> In the context of everything, it's a small bug, which is why I'm going to go ahead and approve the PR so we can move forward. Definitely something we can revisit in the future tho! :)

Awesome! Thank you so much for the review! Just a quick clarifying question on how the product carousel should be! Should there be more delay in between the switch between each logo/should there the transition animation be slower or both? Thanks!

Bookie0:
> Just a quick clarifying question on how the product carousel should be! Should there be more delay in between the switch between each logo/should there the transition animation be slower or both? Thanks! @Bookie0

I feel like we could make it a bit more snappier and responsive. The transition between each logo switch is a bit unnatural because it accelerates and then slows down a lot. You can it barely inching towards the other logo, but i think the animation should end before that.


I know i keep referencing apple lol, but i think they do it quite well. Maybe we can try an effect like this?
https://github.com/user-attachments/assets/fa74f6b7-7b8b-4c60-aa43-f4938eea8592

andrew032011 (approved):

PR #739, opened 2024-11-28
Pull from IDOL Images: Compress Images to < 300 KB

andrew032011:
### Summary 
We previously didn't have any image compression logic, which required us to manually compress images when they exceed the limit.

Make the Pull from IDOL Images script compress images to < 300 KB. We can also revisit this if it makes the images too low quality/blurry.

We may consider overwriting the bucket's images with the compressed version, but only if we verify that the quality is not significantly diminished. For now, we should keep the images in the bucket as a source of truth for what the original image looked like.

### Notion/Figma Link 
https://www.notion.so/Website-Compress-Images-in-Pull-from-IDOL-Images-Job-1410ad723ce18056a2f1f224afae14d1?pvs=4

### Test Plan 
Pull from IDOL Images PR: https://github.com/cornell-dti/idol/pull/729

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -44,6 +44,7 @@
     "netlify-lambda": "^2.0.14",
     "nodemailer": "^6.7.3",
     "serverless-http": "^2.5.0",
+    "sharp": "^0.33.5",
     "typescript": "^4.4.3",
     "uuid": "^8.3.2",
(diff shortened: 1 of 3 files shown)


cchrischen (approved):
Thanks for getting around to this!

PR #743, opened 2024-11-28
Coffee Chat Reminders

patriciaahuang:
### Summary 

Remind members to submit coffee chats, if they don't have a blackout.

### Test Plan 

Screenshot 2024-11-28 at 3 35 08 PM


andrew032011:
I see the use-case for this being more to notify members who don't have a bingo yet to get a bingo/submit their coffee chats. And more specifically, since people who just joined this semester (aka newbies) need to get a bingo, Kevin probably wants a button to just send a reminder to the newbies without bingos.

And I'd display the "exclamation point indicator" for just those folks, newbies without a bingo.

frontend/src/components/Modals/NotifyMemberModal.tsx
```diff
@@ -9,32 +9,35 @@ const NotifyMemberModal = (props: {
   member?: Member;
   members?: Member[];
   trigger: JSX.Element;
-  endOfSemesterReminder: boolean;
```
andrew032011, line 12: Thanks for re-using the TEC components here!

backend/src/API/mailAPI.ts
```diff
@@ -194,3 +195,39 @@ export const sendTECReminder = async (
...
+      approvedCount += 1;
+    }
+    if (chat.status === 'pending') {
+      pendingCount += 1;
+    }
+  });
+
+  const text = `Hey! You currently have ${approvedCount} coffee ${
+    approvedCount !== 1 ? 'chats' : 'chat'
+  } approved and ${pendingCount} coffee ${
```
andrew032011, line 229: If we're only sending a reminder to newbies without bingos it might be good to customize this message so the policy is clear. If not, maybe adding a note like "If you're a newbie, remember you must submit at least one bingo"

backend/src/API/mailAPI.ts
```diff
@@ -194,3 +195,26 @@ export const sendTECReminder = async (
...
+ * @param member - The member being sent the email
+ * @returns - The response body containing information of the member being sent the email
+ */
+export const sendCoffeeChatReminder = async (
+  req: Request,
+  member: IdolMember
+): Promise<AxiosResponse> => {
+  const subject = 'Coffee Chat Reminder';
+
+  const isNewbie = await checkMemberMeetsCategory(member.email, member.email, 'a newbie').then(
```
andrew032011, line 211: Oh actually, we can use `member.semesterJoined` now to figure this out! I think this would involve adding some util function to figure out the current semester and compare it with the field.

backend/src/API/mailAPI.ts
```diff
@@ -194,3 +195,26 @@ export const sendTECReminder = async (
...
+    (result) => result.status === 'pass'
+  );
+
+  const text = `Hey! You currently have no coffee chat bingo this semester
```
andrew032011, line 215: It turns out the way that you format the message in this string is exactly how it will appear in the reminder email. I just tested it out in the deploy preview.
Screenshot 2024-12-01 at 4 48 08 PM

backend/src/API/mailAPI.ts
```diff
@@ -194,3 +195,26 @@ export const sendTECReminder = async (
...
+  ${isNewbie ? ', and newbies are required to get at least 1 bingo' : ''}
+  .\nThis is a reminder to submit your coffee chats by the end of the semester.\n
+  To submit your coffee chats, please visit https://idol.cornelldti.org/forms/coffeeChats.`;
```
andrew032011, line 218: nit: There's also a little indent in the message here, maybe we can remove and try to align the text on the left?

backend/src/API/mailAPI.ts
andrew032011: Also, I'm checking with Kevin about when the deadline should be. I think instead of end of semester we might want to set the deadline as last day of classes.

frontend/src/components/Admin/CoffeeChats/CoffeeChatDashboard.tsx
```diff
@@ -13,23 +14,51 @@ type CoffeeChatStats = {
...
+
+      const fetchNewbieStatus = async () => {
+        const newbieMap: { [key: string]: boolean } = {};
+
+        await Promise.all(
+          allMembers.map(async (member) => {
+            const { status } = await CoffeeChatAPI.checkMemberMeetsCategory(
+              member,
+              member,
+              'a newbie'
```
andrew032011, line 44: instead of this check, can we use `member.semesterJoined`? One thing we should be very careful about is that we shouldn't use the bingo board categories anywhere except for the validation logic `checkMemberMeetsCategory`. We shouldn't even hardcode it anywhere except for the validation logic because the categories are going to change next semester, and there's no guarantee that we'll continue to store any of the data.

andrew032011 (approved):

Issue #734, opened 2024-11-26
Code audit


Bookie0:
Hi team, here are a few issues i've found while poking through the code that we can fix to improve accessibility and maintainability of our code:

- We should use more semantic

andrew032011:
> Same [here](https://github.com/cornell-dti/idol/blob/e6b082a57fab1e1e4b414934c169fb5ce5d2caf6/new-dti-website/components/team/MemberGroup.tsx#L41); we use a `<h3>` even if the page is missing a `<h2>`

For this one, the `MemberGroup` is a component on the team page, which has an `<h2>` in the subheader above, so the `<h3>` wouldn't be considered skipping headers on the entire page.

Bookie0:
> For this one, the `MemberGroup` is a component on the team page, which has an `<h2>` in the subheader above, so the `<h3>` wouldn't be considered skipping headers on the entire page.

Ah I missed that, you're right! we can disregard that one then :)

ladriennel:
@andrew032011 It's all good now

PR #746, opened 2024-11-29
New DTI Website: Use Details/Summary For FAQ Accordion

andrew032011:
### Summary 
For the FAQ accordion [here](https://github.com/cornell-dti/idol/blob/e6b082a57fab1e1e4b414934c169fb5ce5d2caf6/new-dti-website/components/course/DDProjects.tsx#L33), we should use the `<details>` and `<summary>` tags. See this [MDN article for details](https://developer.mozilla.org/en-US/blog/html-details-exclusive-accordions/).

### Notion/Figma Link 
Part of https://github.com/cornell-dti/idol/issues/734

### Test Plan 
Verify FAQ Accordion looks and behaves the same as before.

### DIFF
--- a/new-dti-website/src/app/globals.css
+++ b/new-dti-website/src/app/globals.css
@@ -215,11 +215,3 @@ summary::after {
 details[open] > summary::after {
   rotate: 180deg;
 }
-
-details {
-  user-select: none;
-}
-
-details > *:not(summary) {
-  user-select: text;
-}
--- a/new-dti-website/src/app/globals.css
+++ b/new-dti-website/src/app/globals.css
@@ -192,3 +192,34 @@
   color: black;
   background-color: #fefefe;
 }
+
+summary {
+  padding-top: 1rem;
+  padding-bottom: 1rem;
+  list-style: none;
+  display: flex;
+  align-items: center;
+  font-weight: bold;
+  justify-content: space-between;
+  padding-right: 1rem;
+}
+
+summary::after {
+  content: '';
+  width: 13px;
+  height: 7px;
+  background: url('/icons/dropdown.svg');
+  background-size: cover;
+}
+
+details[open] > summary::after {
+  rotate: 180deg;
+}
+
+details {
+  user-select: none;
+}
+
+details > *:not(summary) {
+  user-select: text;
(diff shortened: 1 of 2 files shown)


new-dti-website/src/app/globals.css
```diff
@@ -192,3 +192,34 @@
...
+  background: url('/icons/dropdown.svg');
+  background-size: cover;
+}
+
+details[open] > summary::after {
+  rotate: 180deg;
+}
+
+details {
+  user-select: none;
```
andrew032011, line 220: Clicking on the accordian a few times in a row leads to some of the summary text getting highlighted, which is a regression from previous behavior. So I made the details not selectable.

cchrischen (approved):
Looks good! Since we are standardizing the accordions between the apply page and the course page, we should check to see which design to settle on before merging this one, in case we settle on the courses one.

andrew032011:
> > Looks good! Since we are standardizing the accordions between the apply page and the course page, we should check to see which design to settle on before merging this one, in case we settle on the courses one.
> 
> I'll surface this in #idol-design.

Vannessa brought up a good point that the Application and Courses Project Showcase Accordians are for different purposes, so it makes sense to make them different. So we decided not to standardize them.

new-dti-website/src/app/globals.css
Bookie0: I do get ur point tho andrew and it is a shame that using `<details>` leads to this regression, so if you feel like that adding `user-select: none;` is worth it in this case, I'll defer to you here.

but let's also make sure that this is an exception bc i dont think we want to make this behavior standard

Bookie0 (approved):
THanksss!

new-dti-website/src/app/globals.css
andrew032011: Yeah, I had a feeling it was bad practice. I reverted that part! I agree with you that rapidly clicking might be something we do when testing it, but in practice, we wouldn't do that so I wouldn't consider it to be a "primary task purpose of a control".

Bookie0: ok, sounds good then!

PR #748, opened 2024-11-29
New DTI Website: Make Icon Order in 'Applications' section Consistent

andrew032011:
### Summary 
For the Product/Design/Development/Business icons, we should use the same order as on the “About” page.

Use the order based on number of ppl on each team:

1. Development
2. Design
3. Business
4. Product

### Notion/Figma Link 
https://www.notion.so/New-Website-APPLY-Icons-in-Applications-section-14d0ad723ce180f5be8ed66e0bda6355?pvs=4

### Test Plan 
Screenshot 2024-11-28 at 8 56 29 PM
Screenshot 2024-11-28 at 8 56 16 PM

### DIFF
--- a/new-dti-website/components/apply/data/applications.json
+++ b/new-dti-website/components/apply/data/applications.json
@@ -1,18 +1,16 @@
 {
-  "product": {
-    "roleName": "Product Manager",
-    "icon": { "src": "/icons/product_sticker.svg", "width": 90, "height": 90 },
+  "development": {
+    "roleName": "Developer",
+    "icon": { "src": "/icons/dev_sticker.svg", "width": 96, "height": 90 },
     "skills": [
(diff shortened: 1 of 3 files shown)


cchrischen (approved):
I think there still is a discrepancy between the order between these two and the order of the chart section. See this [comment](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz?node-id=7-2#1038251939). Maybe we change the order of the chart section?

andrew032011:
> I think there still is a discrepancy between the order between these two and the order of the chart section. See this [comment](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz?node-id=7-2#1038251939). Maybe we change the order of the chart section?

Great catch! Updated.

PR #770, opened 2024-12-05
Website Redesign: Slideshow fix

JasonMun7:
### Summary 

This PR fixes the Slideshow delay bug on the Home page. Instead of relying on the dynamic calculation for the width from using the useScreenSize() hook, I'm simply hiding the element using the tailwind css utilities so we don't have to calculate it on the mount. Flex-col wasn't needed since I didn't realize that the order of the columns wasn't the same haha! 

- [x] Fixed weird delay issue so it should be good! 
- [x] Changed the Red blob as well so it didn't rely on width and so is also present on mount! 
- [x] Adjusted Icon widths and utilized scale for styling instead of width

### Notion/Figma Link 

[Slideshow Delay bug](https://www.notion.so/New-Website-Slideshow-has-a-delay-when-loading-1400ad723ce1801d99a4e02db3b9d760?pvs=4)

### Test Plan 

https://github.com/user-attachments/assets/cbbd1cfb-f6df-4f81-80d7-a0cbdf2fe3fa


andrew032011 (approved):
Literally a lifesaver 🙏🙏

Bookie0 (commented):
wait idk if this is related but i still see this layout shift bug on my side?

https://github.com/user-attachments/assets/c5d0d421-33e9-44ef-b0d6-879dc01f0e25

andrew032011:
> wait idk if this is related but i still see this layout shift bug on my side?
> 
>  Screen.Recording.2024-12-05.at.10.20.53.AM.mov

Source of truth should be https://deploy-preview-770--new-cornell-dti.netlify.app/

Bookie0:
> 
> Source of truth should be https://deploy-preview-770--new-cornell-dti.netlify.app/


ah ok that should be good then thanks!!

Bookie0 (approved):

PR #776, opened 2024-12-06
Coffee Chat: Bingos Aren't Showing Up Correctly on Bingo Dashboard

andrew032011:
### Summary 
Previous logic was overwriting the value of bingo for every "bingo line" that it checked, so the final value of `bingo` would only be the last one checked.

Change this to show bingo as true if at least one line evaluates to true.


cchrischen (approved):
thanks for finding the bug!

PR #799, opened 2024-12-14
Created a "You need to save!" Modal on the Candidate Decider

jujulcrane:
### Summary 

Created a "You Need to Save!" modal if user tries to navigate away from a candidate without saving in the candidate decider. It allows them to either cancel, discard and proceed, or save and proceed.

### Notion/Figma Link 

[https://www.notion.so/Candidate-Decider-Show-a-You-Need-to-Save-Modal-if-User-Tries-to-Navigate-Away-from-Candidate-W-1440ad723ce1806caae2c09118cacceb](url)

### Test Plan 

https://github.com/user-attachments/assets/3d73ba4e-d0a4-4d21-82ba-462a10c1ea9a

### DIFF
--- a/frontend/src/components/Candidate-Decider/CandidateDecider.tsx
+++ b/frontend/src/components/Candidate-Decider/CandidateDecider.tsx
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react';
-import { Button, Dropdown, Checkbox } from 'semantic-ui-react';
+import { Button, Dropdown, Checkbox, Modal } from 'semantic-ui-react';
 import CandidateDeciderAPI from '../../API/CandidateDeciderAPI';
 import ResponsesPanel from './ResponsesPanel';
 import LocalProgressPanel from './LocalProgressPanel';
@@ -17,6 +17,10 @@ type CandidateDeciderProps = {
 };
 
 const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
+  const [isSaved, setIsSaved] = useState(true);
+  const [isOpen, setIsOpen] = useState(false);
+  const [isBypassingModal, setIsBypassingModal] = useState(false);
+  const [navigationDirection, setNavigationDirection] = useState<'next' | 'previous' | null>(null);
   const [currentCandidate, setCurrentCandidate] = useState<number>(0);
   const [showOtherVotes, setShowOtherVotes] = useState<boolean>(false);
 
@@ -68,7 +72,13 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [currentCandidate, instance.candidates, reviews]);
 
-  const next = () => {
+  const next = (force = false) => {
+    if (!force && !isSaved && !isBypassingModal) {
+      setNavigationDirection('next');
+      setIsOpen(true);
+      return;
+    }
+    setIsBypassingModal(false);
     if (currentCandidate === instance.candidates.length - 1) return;
     setCurrentCandidate((prev) => {
       const nextCandidate = prev + 1;
@@ -77,7 +87,13 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
     });
   };
 
-  const previous = () => {
+  const previous = (force = false) => {
+    if (!force && !isBypassingModal) {
+      setNavigationDirection('previous');
+      setIsOpen(true);
+      return;
+    }
+    setIsBypassingModal(false);
     if (currentCandidate === 0) return;
     setCurrentCandidate((prev) => {
       const prevCandidate = prev - 1;
@@ -85,6 +101,15 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
       return prevCandidate;
     });
   };
+
+  const confirmNavigation = (direction: 'next' | 'previous') => {
+    setIsOpen(false);
+    setIsBypassingModal(true);
+    setTimeout(() => {
+      if (direction === 'next') next(true);
+      else previous(true);
+    }, 0);
+  }
 
   const handleRatingAndCommentChange = (id: number, rating: Rating, comment: string) => {
     CandidateDeciderAPI.updateRatingAndComment(instance.uuid, id, rating, comment);
@@ -106,13 +131,52 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
           uuid: ''
         }
       ]);
+      setIsSaved(true);
     }
   };
+
+  useEffect(() => {
+    setIsSaved(
+      currentComment === defaultCurrentComment &&
+      currentRating === defaultCurrentRating
+    );
+  }, [currentComment, currentRating, defaultCurrentComment, defaultCurrentRating]);
 
   return instance.candidates.length === 0 ? (
     <div></div>
   ) : (
+
     <div className={styles.candidateDeciderContainer}>
+      <Modal
+        open={isOpen}
+        onClose={() => setIsOpen(false)}
+        size="small"
+      >
+        <Modal.Header>Don't Forget To Save!</Modal.Header>
+        <Modal.Content>
+          <p>You have unsaved changes. Do you want to save them before navigating?</p>
+        </Modal.Content>
+        <Modal.Actions>
+          <Button onClick={() => setIsOpen(false)}>Clancel</Button>
+          <Button
+            onClick={() => {
+              handleRatingAndCommentChange(
+                currentCandidate,
+                currentRating ?? 0,
+                currentComment ?? ''
+              );
+              confirmNavigation(navigationDirection!);
+            }}>
+            Save and {navigationDirection === 'next' ? 'Proceed' : 'Go Back'}
+          </Button>
+          <Button
+            primary
+            onClick={() => confirmNavigation(navigationDirection!)}
+          >
+            Discard and {navigationDirection === 'next' ? 'Proceed' : 'Go Back'}
+          </Button>
+        </Modal.Actions>
+      </Modal>
       <div className={styles.applicationContainer}>
         <div className={styles.searchBar}>
           <SearchBar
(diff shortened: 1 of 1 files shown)


frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -106,13 +131,42 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
           uuid: ''
         }
       ]);
+      setIsSaved(true);
     }
   };
 
+  useEffect(() => {
+    setIsSaved(currentComment === defaultCurrentComment && currentRating === defaultCurrentRating);
```
andrew032011, line 139: instead of storing this value in a state, what if we just had a variable that stored it? `currentComment, currentRating, defaultCurrentComment, defaultCurrentRating` are all state variables so they'd trigger a re-render anyways when they're changed.

In general it isn't the best pattern to store state that's derived from other states, since it requires you to keep them in sync, and can get lead to buggy behavior if they ever became out of sync

andrew032011:
Can we also handle the two other methods of navigation? (Candidate ID dropdown and search bar)

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -106,13 +131,42 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
     <div></div>
   ) : (
     <div className={styles.candidateDeciderContainer}>
+      <Modal open={isOpen} onClose={() => setIsOpen(false)} size="small">
+        <Modal.Header>Don't Forget To Save!</Modal.Header>
+        <Modal.Content>
+          <p>You have unsaved changes. Do you want to save them before navigating?</p>
+        </Modal.Content>
+        <Modal.Actions>
+          <Button onClick={() => setIsOpen(false)}>Clancel</Button>
```
andrew032011, line 152:
```suggestion
          <Button onClick={() => setIsOpen(false)}>Cancel</Button>
```

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -113,13 +151,43 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
+        </Modal.Actions>
+      </Modal>
       <div className={styles.applicationContainer}>
         <div className={styles.searchBar}>
           <SearchBar
             instance={instance}
             setCurrentCandidate={(candidate) => {
-              setCurrentCandidate(candidate);
-              populateReviewForCandidate(candidate);
+              if (!isSaved) {
```
andrew032011, line 183: can you abstract this function and the below into a helper function? for cleanliness ideally we can put all the handling for this into a single function, as opposed to spreading the logic around.

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -113,13 +151,43 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
+            onClick={() => {
+              handleRatingAndCommentChange(
+                currentCandidate,
+                currentRating ?? 0,
+                currentComment ?? ''
+              );
+              confirmNavigation(navigationDirection!);
+            }}
+          >
+            Save and {navigationDirection === 'previous' ? 'Go Back' : 'Proceed'}
```
andrew032011, line 171: I'm worried that this language is confusing "Go Back" might be interpreted as going back to the previous state (someone might think this is the same as "Discard")

maybe we don't need to show the navigation direction? Just "Save", "Discard", and "Cancel" seem good with me

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -86,6 +109,21 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
     });
   };
 
+  const confirmNavigation = (direction: 'next' | 'previous' | 'other') => {
+    setIsOpen(false);
+    setIsBypassingModal(true);
+    setTimeout(() => {
```
andrew032011, line 115: hm, wondering what the point of `setTimeout` with 0 milliseconds is?

jujulcrane: Oh oops I forgot to get rid of that - I was trying to fix a problem I was running into where it was making the user click save twice before it worked but I ended up fixing it a different way

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -17,6 +17,14 @@ type CandidateDeciderProps = {
 };
 
 const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
+  const [isOpen, setIsOpen] = useState(false);
```
andrew032011, line 20: rename to `isModalOpen`

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -17,6 +17,14 @@ type CandidateDeciderProps = {
...
+  const [isBypassingModal, setIsBypassingModal] = useState(false);
```
andrew032011, line 21: I'm pretty sure we can get rid of this `isBypassingModal` entirely. It feels like a combination of `force` and `isSaved` is enough.

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -109,17 +145,51 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
     }
   };
 
+  const handleCandiateChange = (candidate: number) => {
```
andrew032011, line 148:
```suggestion
  const handleCandidateChange = (candidate: number) => {
```

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -68,7 +79,13 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [currentCandidate, instance.candidates, reviews]);
 
-  const next = () => {
+  const next = (force = false) => {
+    if (!force && !isSaved && !isBypassingModal) {
```
andrew032011, line 83: Now that we're not using `navigationDirection` anymore for the text, it feels like `prev`, `next`, and `handleCandidateChange` can all be condensed into just one function. Can you try deleting `prev` and `next` and just using `handleCandiateChange`? (you might have to check if `candidate` number is within `0...instance.candidates.length - 1` but otherwise the looks the same to me)

and then we can get rid of `navigationDirection `!

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -17,6 +17,14 @@ type CandidateDeciderProps = {
 };
 
 const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const [isBypassingModal, setIsBypassingModal] = useState(false);
+  const [navigationDirection, setNavigationDirection] = useState<
+    'next' | 'previous' | 'other' | null
+  >(null);
+  const [navigationDirectionCandidate, setNavigationDirectionCandidate] = useState<number | null>(
```
andrew032011, line 25: rename? "navigationDirection" doesn't really apply anymore if we're not tracking navigation direction

andrew032011 (commented):
It's really coming together! Just some comments to help reduce code duplication

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -109,17 +98,64 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
+        <Modal.Actions>
+          <Button onClick={() => setIsModalOpen(false)}>Cancel</Button>
+          <Button
+            onClick={() => {
+              handleRatingAndCommentChange(
+                currentCandidate,
+                currentRating ?? 0,
+                currentComment ?? ''
+              );
+              if (nextCandidate !== null) {
```
andrew032011, line 128: How about a function `navigateToNextCandidate` for this and the code that's being duplicated below?

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -109,17 +98,64 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
+        </Modal.Actions>
+      </Modal>
       <div className={styles.applicationContainer}>
         <div className={styles.searchBar}>
           <SearchBar
             instance={instance}
             setCurrentCandidate={(candidate) => {
-              setCurrentCandidate(candidate);
-              populateReviewForCandidate(candidate);
+              handleCandidateChange(candidate);
```
andrew032011, line 158: Actually you can just pass in `handleCandidateChange` here
i.e.
```
(candidate) => { handleCandidateChange(candidate); }
```
is the same as
```
handleCandidateChange
```

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -136,29 +172,39 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
           <span className={styles.ofNum}>of {instance.candidates.length}</span>
           <Button.Group className={styles.previousNextButtonContainer}>
-            <Button basic color="blue" disabled={currentCandidate === 0} onClick={previous}>
+            <Button
+              basic
+              color="blue"
+              disabled={currentCandidate === 0}
+              onClick={() => {
+                if (currentCandidate === 0) return;
+                setNextCandidate(currentCandidate - 1);
```
andrew032011, line 186: How about let's move `setNextCandidate` into `handleCandidateChange`? And move the bound checks there as well (==0 and == instance.candidates.length - 1)?

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -109,17 +98,64 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
+              setIsModalOpen(false);
+            }}
+          >
+            Save
+          </Button>
+          <Button
+            primary
+            onClick={() => {
+              if (nextCandidate !== null) {
+                setCurrentCandidate(nextCandidate);
```
andrew032011, line 142: See above, move this into a separate function `navigateToNextCandidate` to reduce code duplication

andrew032011 (approved):

PR #817, opened 2025-01-05
Add slightly larger favicon

andrew032011:
### Summary 
Add favicon to website that's larger than the one used here: https://github.com/cornell-dti/idol/pull/816.

@Bookie0 provided an .ico file.


Bookie0 (approved):
ty!!

PR #832, opened 2025-01-31
fix candidate decider google docs links, other improvements requested from leads

oscarwang20:
### Summary 

After setting up the candidate decider instance for SP25 I noticed that all the resumes were inaccessible due to a broken URL. After some testing, it looks like Google docs links are case sensitive, hence this small PR.

I've also included some small changes that were requested last semester by leads:
- 1-indexed dropdowns on the candidate search bar and candidate ID selector
- github, linkedin, resume, and portfolio links now open in a new tab 

### Notion/Figma Link 
N/A

### Test Plan 

Testing was done manually to:
- ensure that parsed resume URLs now work properly
- check that other credential headers that contain URLs still work
- verify that the correct applicants are still rated despite new 1-indexed display text


cchrischen (approved):
looks good :) thanks for the qol updates
it just looks like the test case is failing, can you update that

PR #833, opened 2025-02-01
IDOL: Interview Scheduler

cchrischen:
### Summary 
This PR implements an MVP of the interview scheduler. Admins are allowed to create new scheduler instances at `/admin/interview-scheduler`. There is currently no way to modify an existing instance, other than opening/closing the scheduler. The rest of the feature happens at `/interview-scheduler/[uuid]`. 

Applicants are only allowed to access schedulers in which they are applicants of. They can sign themselves up for a time slot, after which they can view the time slot they signed up for. If needed, they can request a time slot change via email, and a lead will change it on the scheduler. They cannot see which lead/member(s) signed up for this slot.

Members can navigate to the same page and sign up for interview slot(s). They cannot see which applicant, if any, signed up for this slot.

Admins can also follow the same process, but can see every parties' identity. Admins can also edit each party via a dropdown menu. They can add availabilities through a button on the top right. Deleting time slots was made intentionally tedious due to its implications. 

The backend should prevent any sensitive information from being sent back to the client, if that client is not a DTI member. 

Bug Fixes:
* There was a bug that allowed anyone to log into IDOL introduced in the last PR.

### Notion/Figma Link 
[Figma](https://www.figma.com/design/G5m5N7c1opQFGBVMM1Uvg4/IDOL-Recruitment-Tool-(Copy)?node-id=1-4&p=f&t=fveBttyqguXzKz8b-0)

### Test Plan 
Playtesting at 2/1 12 - 1!

### Notes 

Sorry for the big PR.

### DIFF
--- a/backend/src/API/interviewSchedulerAPI.ts
+++ b/backend/src/API/interviewSchedulerAPI.ts
@@ -6,6 +6,12 @@ import PermissionsManager from '../utils/permissionsManager';
 const interviewSchedulerDao = new InterviewSchedulerDao();
 const interviewSlotDao = new InterviewSlotDao();
 
+const censoredApplicant: Applicant = {
+  email: '',
+  firstName: '',
+  lastName: ''
+};
+
 export const getAllApplicants = async (): Promise<string[]> => {
   const instances = await interviewSchedulerDao.getAllInstances();
   const applicants = new Set<string>();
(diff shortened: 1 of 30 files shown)


cchrischen:
Closing to break up into multiple PRs.

PR #835, opened 2025-02-05
IDOL: Update coffee chat bingo board for SP25

cchrischen:
### Summary 
This PR updates the bingo board for spring 2025. 
* Updated the bingo board. The board is still saved as a constant, but it will change with an upcoming PR.
* Updated autochecker to check "is an advisor" and "is a newbie". Everything else must be manually checked by the C&I lead until we obtain the data from the category form.
* Added an isArchived field for all past coffee chats. New coffee chats should have it set to false.

### Notion/Figma Link 
[Notion](https://www.notion.so/Idol-SP25-1890ad723ce180a19f3ff99268d34e53?p=1900ad723ce1809da802c767149da395&pm=s)

### Test Plan 
Tested the new bingo board such that it only displays unarchived chats, members could submit, and autocheck works.

https://github.com/user-attachments/assets/436c75da-dab1-49f5-9e14-c910de43ce02

### Notes 
`CoffeeChat` and `DBCoffeeChat` types have been updated and have been reflected in both databases.

### DIFF
--- a/backend/src/API/coffeeChatAPI.ts
+++ b/backend/src/API/coffeeChatAPI.ts
@@ -65,15 +65,12 @@ export const createCoffeeChat = async (
     'approved',
     coffeeChat.otherMember
   );
-  const prevChats = [
-    ...pendingChats.filter((chat) => !chat.isArchived),
-    ...approvedChats.filter((chat) => !chat.isArchived)
-  ];
+  const prevChats = [...pendingChats, ...approvedChats];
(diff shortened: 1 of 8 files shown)


JasonMun7 (approved):
Hey Chris! The changes make it look a lot cleaner! 
It seems the function in the checks (checkMemberMeetsCategory) is returning no data instead of "pass"/"fail" so I would just check that before merging (Might just need a re-run)!

PR #837, opened 2025-02-07
Website: Add meta titles

cchrischen:
### Summary 
The DTI website lacked titles, so all pages were just the URL.

### DIFF
--- a/new-dti-website/src/hooks/useTitle.ts
+++ b/new-dti-website/src/hooks/useTitle.ts
@@ -1,0 +1,9 @@
+import { useEffect } from 'react';
+
+const useTitle = (title?: string) => {
+  useEffect(() => {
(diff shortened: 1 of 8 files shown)


keverie (approved):
LGTM! 🐐

new-dti-website/src/hooks/useTitle.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect } from 'react';
+
+const useTitle = (title?: string) => {
+  useEffect(() => {
+    document.title = `${title ? `${title} | ` : ''} Cornell DTI`;
+  }, []);
```
JasonMun7, line 6:
```suggestion
  }, [title]);
```
I would include the title within the dependency array so it's not only on mount! If we every decide to change the title in the future and for styling!

JasonMun7 (approved):
Great job creating a hook for this!

PR #841, opened 2025-02-11
Created toggle that removes applicant name by default and replaces wi…

SZCodes:
…th candidate id

### Summary 

Created a toggle button. By default, candidate names and links (GitHub, LinkedIn, Email, and Portfolio) are hidden and the applicant's name is replaced with their Candidate ID. Clicking the toggle button reveals all the hidden information.


cchrischen (approved):
Good work! Your PR is not building because some files are not formatted correctly. You can see which files you should format by clicking on details next to "Build / check (pull_request)".

PR #842, opened 2025-02-12
Filtered out archived coffee chats from square with same name on revi…

SZCodes:
…ew coffee chats page and dashboard page

### Summary 

Fixed bug where coffee chat in categories of the same name in a different semester are still shown. Filtered out archived coffee chats when displaying current semester coffee chats in coffee chat review page and dashboard page.


cchrischen (approved):
Thanks for getting this out so quickly. Can you also fix this other thing on the admin side? The board still calculates the number of pending coffee chats without filtering out archived chats. 
image

cchrischen (approved):

PR #852, opened 2025-02-18
Interview Scheduler Backend

cchrischen:
### Summary 
This PR adds the interview scheduler backend, adding CRUD operations for interview slots and interview schedulers. Some endpoints are duplicated for applicants and IDOL members for permission. Data sent back to applicants should be censored. Updating interview slots should be done transactionally to prevent overwritten data.

### Test Plan 
Unit tests in `InterviewSchedulerDao.test.ts`.

### Breaking Changes 
See the two new Firebase collections

### DIFF
--- a/backend/src/API/interviewSchedulerAPI.ts
+++ b/backend/src/API/interviewSchedulerAPI.ts
@@ -1,13 +1,186 @@
 import InterviewSchedulerDao from '../dao/InterviewSchedulerDao';
+import InterviewSlotDao from '../dao/InterviewSlotDao';
+import { NotFoundError, PermissionError } from '../utils/errors';
+import PermissionsManager from '../utils/permissionsManager';
+import { getMember } from './memberAPI';
 
 const interviewSchedulerDao = new InterviewSchedulerDao();
+const interviewSlotDao = new InterviewSlotDao();
 
-// eslint-disable-next-line import/prefer-default-export
+const censoredApplicant: Applicant = {
+  email: '',
+  firstName: '',
+  lastName: '',
+  netid: ''
+};
+
 export const getAllApplicants = async (): Promise<string[]> => {
   const instances = await interviewSchedulerDao.getAllInstances();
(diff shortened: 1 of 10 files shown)


backend/src/API/interviewSchedulerAPI.ts
```diff
@@ -1,13 +1,186 @@
...
+  netid: ''
+};
 
-// eslint-disable-next-line import/prefer-default-export
 export const getAllApplicants = async (): Promise<string[]> => {
   const instances = await interviewSchedulerDao.getAllInstances();
   const applicants = new Set<string>();
-  instances
-    .filter((inst) => !inst.isArchived)
-    .forEach((inst) => inst.applicants.forEach((app) => applicants.add(app)));
```
JasonMun7, line 11: Is there a reason we're not filtering before we loop through each instance?

backend/tests/InterviewSchedulerDao.test.ts
```diff
@@ -0,0 +1,32 @@
+import InterviewScheulderDao from '../src/dao/InterviewSchedulerDao';
+import { candidateDeciderCollection } from '../src/firebase';
+import { fakeInterviewScheduler } from './data/createData';
+
+const interviewSchedulerDao = new InterviewScheulderDao();
+
```
JasonMun7, line 6: Awesome test cases!!

JasonMun7 (approved):
Hey Chris! This looks like an amazing start for the Interview Scheduler! Also thanks for breaking this PR up!

backend/src/API/interviewSchedulerAPI.ts
cchrischen: Oh yeah, good question. I removed the `isArchived` field since I didn't see a use case for it anymore. `isOpen` would prevent applicants from editing time slots. However, we still want to retrieve closed instances, because applicants may want to check when they scheduled an interview after the sign up deadline.

PR #853, opened 2025-02-22
Giving day: navbar updates to "Sponsor" and "Apply"

Bookie0:
[figma](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1294-1572&p=f&t=51d4oYxL0O06amCX-11)

IDOL site giving day changes

made "Sponsor" link red to bedazzle it
made "Apply" a white button so that it doesn't look weird being a link next to "Sponsor"

made same changes for mobile too (and i left-aligned links for readability)

|X|Before|After|
|-|-|-|
|Desktop|Screenshot 2025-02-22 at 3 54 46 PM|Screenshot 2025-02-22 at 3 54 56 PM|
|Mobile|Screenshot 2025-02-22 at 3 55 24 PM|Screenshot 2025-02-22 at 3 55 14 PM|

### DIFF
--- a/new-dti-website/components/navbar.tsx
+++ b/new-dti-website/components/navbar.tsx
@@ -8,33 +8,27 @@ import SectionWrapper from './hoc/SectionWrapper';
 const navbarItems = [
   {
     name: 'Team',
-    url: '/team',
-    special: false
+    url: '/team'
   },
   {
     name: 'Products',
-    url: '/products',
-    special: false
+    url: '/products'
   },
   {
     name: 'Course',
-    url: '/course',
-    special: false
+    url: '/course'
   },
   {
     name: 'Initiatives',
-    url: '/initiatives',
-    special: false
+    url: '/initiatives'
   },
   {
     name: 'Sponsor',
-    url: '/sponsor',
-    special: true
+    url: '/sponsor'
   },
   {
     name: 'Apply',
-    url: '/apply',
-    special: false
+    url: '/apply'
   }
 ];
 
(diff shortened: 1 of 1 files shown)


new-dti-website/components/navbar.tsx
```diff
@@ -59,13 +59,31 @@ const Navbar: React.FC = () => {
...
               >
-                {item.name}
+                <span
+                  className={`
+                    ${
+                      item.name === 'Sponsor'
+                        ? 'bg-gradient-to-r from-[#F25454] to-[#D63D3D] bg-clip-text text-transparent'
+                        : 'text-white'
+                    }
+                    ${item.name === 'Apply' ? '!text-[#000000]' : ''}
```
Bookie0, line 81: need `!` to override default label color (white on white)

new-dti-website/components/navbar.tsx
```diff
@@ -99,14 +117,36 @@ const Navbar: React.FC = () => {
...
                   >
-                    {item.name}
+                    <span
+                      className={`
+                    ${
+                      item.name === 'Sponsor'
+                        ? 'bg-gradient-to-r from-[#F25454] to-[#D63D3D] bg-clip-text text-transparent'
+                        : 'text-white'
+                    }
+                    ${item.name === 'Apply' ? '!text-[#000000]' : ''}
```
Bookie0, line 144: ditto

new-dti-website/components/navbar.tsx
```diff
@@ -99,14 +117,36 @@ const Navbar: React.FC = () => {
...
                 {navbarItems.map((item) => (
                   <a
                     key={item.name}
-                    className="hover:underline cursor-pointer text-white text-base md:text-2xl font-normal underline-offset-8 decoration-2 decoration-white"
                     href={item.url}
+                    className={`group cursor-pointer p-4 underline-offset-8 decoration-2 h-[40px] flex items-center ${
+                      pathname === item.url ? 'underline' : ''
+                    } ${
+                      item.name === 'Sponsor'
+                        ? 'mr-2 rounded-[128px] [transition:50ms_ease-out] border border-[rgba(227,73,73,0.60)] px-4 py-2 hover:bg-[rgba(229,74,74,0.20)] justify-center h-[52px]'
```
Bookie0, line 129: FYI – ideally, we would use proper variables etc. instead of hard coded values, but in this case i think it's okay to make an exception because of:
1. Giving Day approaching (all these changes are for Giving Day) soon
2. this is just the temporary website and will be revamped later

JasonMun7:
> > Hi @Bookie0!! This is looking great! I really love how the more active CTAs stand out, and the left alignment on mobile enhances readability.
> > One suggestion—I’d recommend leveraging our existing button component in components/ui/button.tsx for these new buttons to maintain consistency across the UI.
> > Great work!
> 
> hi Jason, thanks for the comment and apologies for the late reply!
> 
> I'm not sure if we should be using the `<Button>` component since these are links? I think what we've been doing on other links throughout the website (eg: "Learn more" or "How we give back" on the homepage) is to just use an `<a>` tag, which is what I've been doing in this navbar.
> 
> Also, I also chose not to use the existing `primary-button` class because of several reasons:
> 
> * The styles are pretty different than what we want
> * This is just the temporary website until Giving Day, so I figured we don't need to make a net-new component (just inserting it in the navbar should be fine)
> 
> Let me know what you think! :)

Ohhh ok that's completely fine! I wasn't aware of this being the temporary website! Ignore my comments then and go ahead with the merge!

JasonMun7 (approved):

Bookie0:
perfect thank you Jason!

PR #868, opened 2025-03-01
[Giving day] fixed navbar selected item underline bug

Bookie0:
i notice the navbar didn't show an underline under the currently selected page. this is probably a regression from [my PR here](https://github.com/cornell-dti/idol/pull/853)

- added underline back when the item is selected (only on links, not on the Apply/Sponsor buttons)

|X|Before|After|
|-|-|-|
|Desktop|Screenshot 2025-02-28 at 10 01 30 PM|Screenshot 2025-02-28 at 10 00 50 PM|
|Mobile|Screenshot 2025-02-28 at 10 02 09 PM|Screenshot 2025-02-28 at 10 02 00 PM|

### DIFF
--- a/new-dti-website/components/navbar.tsx
+++ b/new-dti-website/components/navbar.tsx
@@ -139,7 +139,7 @@ const Navbar: React.FC = () => {
                         : 'text-white'
                     }
                     ${pathname === item.url ? 'underline' : ''} 
-                    ${item.name === 'Apply' ? '!text-[#000000] !no-underline' : ''}
+                    ${item.name === 'Apply' ? '!text-[#000000] no-underline' : ''}
                     ${item.name !== 'Apply' ? 'group-hover:underline' : ''}
                   `}
                     >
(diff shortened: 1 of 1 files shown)


new-dti-website/components/navbar.tsx
```diff
@@ -141,7 +138,8 @@ const Navbar: React.FC = () => {
                         ? 'bg-gradient-to-r from-[#F25454] to-[#D63D3D] bg-clip-text text-transparent'
                         : 'text-white'
                     }
-                    ${item.name === 'Apply' ? '!text-[#000000]' : ''}
+                    ${pathname === item.url ? 'underline' : ''}
```
Bookie0, line 141: we need to add it on the `<span>` so it applies to the text label instead of on the parent `<a>`

new-dti-website/components/navbar.tsx
```diff
@@ -141,7 +138,8 @@ const Navbar: React.FC = () => {
                         ? 'bg-gradient-to-r from-[#F25454] to-[#D63D3D] bg-clip-text text-transparent'
                         : 'text-white'
                     }
-                    ${item.name === 'Apply' ? '!text-[#000000]' : ''}
+                    ${pathname === item.url ? 'underline' : ''} 
+                    ${item.name === 'Apply' ? '!text-[#000000] no-underline' : ''}
```
Bookie0, line 142: override underline on buttons (we shouldn't show an underline)

JasonMun7 (approved):

PR #869, opened 2025-03-02
[Giving Day] Update Website Images

cchrischen:
### Summary 
Updated the images on the website to include more up-to-date images.


new-dti-website/public/images/home-3.jpg
Bookie0, line 1: tiny nit – could we make this image square format? Just like the other images in "Teaching the Cornell community" and "Expanding reach to our community" .

Screenshot 2025-03-02 at 11 59 26 AM


Here's a cropped version:

home-3 1

Bookie0 (approved):
Nice, thank you Chris!

And then once the new hero section components are added, we can add [this one](https://drive.google.com/file/d/1e1gVrE5R1v7po5nZzSgdWIXvVuvQIArm/view?usp=sharing) too for the Sponsor page.

JasonMun7 (changes requested):
Hey Chris! Thanks for the updates—loving the images! Haha.

I was wondering if we could optimize them a bit since they’re quite large. Would it be possible to convert them to WebP instead of JPG/JPEG? I usually use [PhotoPea](https://www.photopea.com/) to scale them down by a percentage and reformat them. Ideally, we should aim for around 150 KB or less, especially for the larger images.

Let me know what you think! Thanks again!

JasonMun7 (approved):

PR #886, opened 2025-03-12
CC Autochecker Fail for past sem

JasonMun7:
### Summary 

This PR implements the Autochecker to fail for coffee chats you've chatted with last semester 

- [x] Added the check within `coffeeChatAPI.ts`
- [x] Added in more tests `CoffeeChatAPI.test.ts` for the autochecker which passed
- [x] Removed unecessary typings in `CoffeeChatAPI.test.ts` and `CoffeeChatDao.test.ts` since they're already strongly typed 

### Notion/Figma Link 

- [Notion](https://www.notion.so/Coffee-Chat-Make-Autochecker-fail-for-coffee-chats-with-people-who-you-ve-chatted-with-last-semeste-1890ad723ce18188b997defefc11ca7e?pvs=4)

### Testing
- Ran using Jest 

### Notes
- I currently check to see if there is any archived coffee chats with the same members
- Should I check purely off the Sem? @cchrischen


backend/src/API/coffeeChatAPI.ts
```diff
@@ -193,6 +193,20 @@ export const runAutoChecker = async (uuid: string, user: IdolMember): Promise<Co
     throw new BadRequestError(`Coffee chat with uuid: ${uuid} does not exist`);
   }
 
+  const archivedChats = await coffeeChatDao.getCoffeeChatsByUser(
+    coffeeChat.submitter.email,
+    undefined,
```
cchrischen, line 198: You probably also want to only check accepted archive chats, so I think we should pass in `approved` here.

backend/src/API/coffeeChatAPI.ts
```diff
@@ -193,6 +193,20 @@ export const runAutoChecker = async (uuid: string, user: IdolMember): Promise<Co
...
+    coffeeChat.otherMember
+  );
+
+  const hasArchivedChat = archivedChats.some((chat) => chat.isArchived);
+
+  if (hasArchivedChat) {
+    throw new BadRequestError(
+      `Coffee chat with uuid: ${uuid} has an archived chat with the same submitter and other member`
+    );
+  }
```
cchrischen, line 208: If there is a chat from the previous semester, we should update the coffee chat to have `memberMeetsCategory` be `fail` and an appropriate `errorMessage` isntead of throwing an error. The purpose of this function is to (potentially) update these two fields for autochecking. See the code below for more reference.

cchrischen (commented):
Yay! Thanks Jason for the comprehensive test suite. Could you also add functionality in the function `checkMemberMeetsCategory`? Right now, we only autocheck the two categories that we can check from IDOL information. The other 14 would have to be checked through querying the coffee chat suggestions collection.

backend/tests/CoffeeChatAPI.test.ts
```diff
@@ -188,11 +189,69 @@ describe('User is lead or admin', () => {
   });
 });
 
+describe('runAutoChecker tests', () => {
```
jujulcrane, line 192: Maybe you could add a test that tests that the runAutoChecker throws an error if the user is not an admin? You could also add a test that checks the successful update with a specific category.

jujulcrane (approved):
Hi Jason, this looks really good! I left a comment about some more tests you might think about implementing, although I'm not sure if they are necessary. Great work!

PR #892, opened 2025-03-19
Timeline Bug fix

JasonMun7:
### Summary 

This PR fixes the timeline bug when there is a large gap in dates

- [x] instead of only relying on the first and last date to adjust the progress percentage, restart progress whenever an event date has passed 

### Test Plan 

- Tested with multiple dates and large gaps
- Netlify deploy


cchrischen (approved):
yippee!

PR #915, opened 2025-04-14
[Cand decider] New `<Input>` component

Bookie0:
# Summary

- Created new `<Input>` component
- It renders an `<input>` or `<textarea>` component accordingly
- You can also specify the height and resize properties of the input

|X|Before|After|
|-|-|-|
|Default|def bef|def aft|
|Hover|def bef 2|hov aft|
|Focus|Screenshot 2025-04-14 at 2 31 01 PM|focus aft|
|Disabled|Screenshot 2025-04-14 at 2 40 18 PM|Screenshot 2025-04-14 at 2 39 57 PM|

[From Figma link Input component](https://www.figma.com/design/qHNMOp6rrIFMv1wBXsVtYZ/SP25-IDOL-Candidate-Decider-Tool?node-id=114-804&m=dev)

# Test plan
- Open cand decider
- Interact with the Input, hover, focus, type in it
- Make sure it works well, make sure it saves properly

### DIFF
--- a/frontend/src/components/Common/Input/Input.module.css
+++ b/frontend/src/components/Common/Input/Input.module.css
@@ -1,0 +1,25 @@
+.input {
+  font-size: 16px;
+  border-radius: 6px;
+  border: 1px solid var(--border-default);
+  box-sizing: border-box;
+  width: 100%;
+  height: 256px;
+  padding: 16px;
+  transition: 50ms ease-in;
+  font-family: Lato, 'Helvetica Neue', Arial, Helvetica, sans-serif;
+}
+
+.input:hover {
+  border-color: var(--border-dimmer);
+}
+
+.input:focus {
+  outline: 2px solid var(--accent-focus);
+  outline-offset: -1px;
+}
+
+input.input::placeholder,
+textarea.input::placeholder {
+  color: var(--fg-dimmer);
+}
--- a/frontend/src/components/Common/Input/Input.module.css
+++ b/frontend/src/components/Common/Input/Input.module.css
@@ -23,3 +23,11 @@ input.input::placeholder,
 textarea.input::placeholder {
   color: var(--fg-dimmer);
 }
+
+.input:disabled {
+  opacity: 50;
+}
+
+.input:disabled:hover {
+  border-color: var(--border-default);
+}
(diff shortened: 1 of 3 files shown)


frontend/src/components/Common/Input/Input.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type InputProps = {
+  value: string | undefined;
+  onChange: (e: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+  placeholder?: string;
+  multiline?: boolean;
+  maxHeight?: number;
+  resize?: 'none' | 'both' | 'horizontal' | 'vertical';
+  disabled?: boolean;
+};
```
Bookie0, line 12: Props here to control styling and functionality of the component

frontend/src/components/Common/Input/Input.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (multiline) {
+    return (
+      <textarea
+        value={value}
+        onChange={onChange}
+        placeholder={placeholder}
+        className={styles.input}
+        style={style}
+        disabled={disabled}
+      />
```
Bookie0, line 37: According to [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea) we should be using the `<textarea>` component when you can write multiple lines in the textbox

frontend/src/components/Common/Input/Input.module.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.input:hover {
+  border-color: var(--border-dimmer);
+}
+
+.input:focus {
+  outline: 2px solid var(--accent-focus);
+  outline-offset: -1px;
+}
```
Bookie0, line 20: Using `outline` instead of `border` so that on focus, it doesn't make the component fidget/move other elements around

frontend/src/components/Common/Input/Input.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import styles from './Input.module.css';
+
+type InputProps = {
+  value: string | undefined;
```
cchrischen, line 5: nit: `value?: string`

Bookie0: Done, thx Chris!

PR #916, opened 2025-04-14
Sz266/candidate decider button component

SZCodes:
### Summary 

Created new button component for candidate decider
* Created new files Button.tsx and Button.module.css
* Updated Previous, Next, Save, and Cancel, Save and Discard buttons

Designs based on Figma file: https://www.figma.com/design/qHNMOp6rrIFMv1wBXsVtYZ/SP25-IDOL-Candidate-Decider-Tool?node-id=59-2433&m=dev

### New button components compared to previous
|  | New | Old |
|-|-|-|
| Candidate review and navigation | Screenshot 2025-04-16 at 1 29 22 PM | Screenshot 2025-04-14 at 4 06 48 PM |
| Modal | Screenshot 2025-04-16 at 1 33 55 PM | Screenshot 2025-04-14 at 4 06 59 PM |

### Video demo of new buttons on page

https://github.com/user-attachments/assets/7e269b02-699f-4f86-88ec-39adb14e3fe0


frontend/src/components/Candidate-Decider/CandidateDecider.tsx
SZCodes, line 1: I don't like the way the popup buttons are very close together but could probably be a spacing update later? See the Summary for what the popup box looks like

frontend/src/components/Common/Button/Button.module.css
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.default:hover {
+color: var(--fg-default, #171717);
+background: var(--bg-dimmer, #FAFAFA);
+}
+
+.default:focus {
+border: 2px solid var(--accent-focus, #0E4DD4);
+}
```
Bookie0, line 30: Let's change all the `:focus` psuedo-classes to `:focus-visible`.

[This gives more details](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#focus_vs_focus-visible), but essentially, `:focus` applies the styles whenever the element is focused through any way (mouse click, keyboard tab, or script). And `:focus-visible` only shows the styles when the element is activated via the keyboard.

We want this blue outline to only apply in the latter case.

frontend/src/components/Common/Button/Button.module.css
```diff
@@ -0,0 +1,90 @@
+.custom-button {
```
Bookie0, line 1: I think it would make more sense to call these like `base-styles` or something that is a bit more explicit?

frontend/src/components/Common/Button/Button.module.css
```diff
@@ -0,0 +1,90 @@
...
+background: var(--fg-default, #171717); 
+}
+
+.primary:hover {
+background: var(--fg-dimmer, #333);
+}
+
+.primary:focus {
+border: 2px solid var(--accent-focus, #0E4DD4);
+}
```
Bookie0, line 44: Since the focus state is the same for all of the buttons, I'd define the `:focus-visible` properties only one time on the base style buttons

Bookie0: Also, instead of using `border` (which makes the element shift/change size, see video), I think we should use `outline` instead.

https://github.com/user-attachments/assets/650ffccf-73cf-4814-ba52-0548b5024451

frontend/src/components/Common/Button/Button.module.css
```diff
@@ -0,0 +1,90 @@
...
+
+.negative {
+color: #FC0036;
```
Bookie0, line 47: oops sorry, realized the Figma didn't have up to date color variables – we should use the `--accent-strong-no` variable

frontend/src/components/Common/Button/Button.module.css
```diff
@@ -0,0 +1,90 @@
...
+background: var(--bg-default, #FFF);
+}
+
+/* States */
+.icon {
+width: 16px;
+height: 16px;
+display: flex;
+align-items: center;
+}
```
Bookie0, line 90: also seems like the indentations are a bit off – to make sure linter doesn't get mad, make sure to format the code by saving it (`cmd-S`) or  via the terminal (`yarn format`)

frontend/src/components/Common/Button/Button.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import styles from './Button.module.css'; // Import the CSS file
```
Bookie0, line 2:
```suggestion
import styles from './Button.module.css';
```

(We don't need comments like this since it's pretty obvious what this line does lol)

Bookie0 (changes requested):
Nice!! This is a good start Sophie! Just a few comments regarding code cleanliness and a few nitpicky styles.

Thank you!

Bookie0:
Also I think you have the wrong order for the Before/After images in the PR description's table?
Screenshot 2025-04-14 at 9 12 59 PM

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
Bookie0: Also that modal is a good catch, I'll create a new component for it.

frontend/src/components/Common/Button/Button.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+const Button: React.FC<ButtonProps> = ({ label, disabled = false, onClick, icon, variant='default'}) => (
+  <button
+    disabled={disabled}
+    onClick={onClick}
+    className={`${styles['custom-button']} ${styles[variant]} ${disabled ? styles.disabled : ''}`}
+  >
+    {icon && <span className="icon">{icon}</span>}
+    {label}
```
Bookie0, line 19: also nit but have you tried rendering buttons with icons from [Lucide](https://lucide.dev/icons/) just to test this? 

Could be good to test at least locally.

frontend/src/components/Common/Button/Button.module.css
```diff
@@ -0,0 +1,90 @@
...
+  padding: 16px 24px;
+  border-radius: 6px;
+  border: 1px solid var(--border-default, #EBEBEB);
+  background: var(--bg-default, #FFF);
+  font-size: 14px;
+  cursor: pointer;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  gap: 10px;
```
Bookie0, line 11:
```suggestion
  gap: 8px;
```

frontend/src/components/Common/Button/Button.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import styles from './Button.module.css'; // Import the CSS file
+
+type ButtonProps = {
+  label?: string;
+  disabled?: boolean;
+  onClick?: () => void;
+  icon?: React.ReactNode;
```
Bookie0, line 8: Also if you're up to it (if not, no worries, I can follow up in a subsequent PR), but I think it'd be nice to allow for more granular control regarding the icon position.

In the [design system](https://www.figma.com/design/qHNMOp6rrIFMv1wBXsVtYZ/SP25-IDOL-Candidate-Decider-Tool?node-id=122-763&m=dev), we're able to set the icon on the left or the right (or not at all). 

You could do this similar to how you already have the optional `icon` prop set up, but instead, you'd have `iconLeft` and `iconRight` as props.

Bookie0 (commented):
A few other comments! :)

Also I see you have merge conflicts – we should keep both the existing changes, and then add on your new changes from your branch (essentially we don't want to lose anyone's work when we merge this). But lmk if you need help on fixing this.

frontend/src/components/Common/Button/Button.module.css
Bookie0: @SZCodes did u catch this?

frontend/src/components/Common/Button/Button.tsx
Bookie0: hi @SZCodes do u feel like tackling this?

frontend/src/components/Common/Button/Button.tsx
Bookie0: @SZCodes just checking if you tested this?

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -165,19 +173,17 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
             onClick={() => {
               if (nextCandidate !== null) {
                 navigateToNextCandidate(nextCandidate);
               }
             }}
-          >
-            Discard
-          </Button>
+            variant="primary"
+          />
```
Bookie0, line 186: Nice! Modal buttons look well spaced out now. Can we change the variants of the button to make "Save" `primary`, "Cancel" `default`, and "Discard" `negative`? I think this makes more sense conceptually.

Screenshot 2025-04-15 at 10 15 17 PM

Bookie0 (changes requested):
Just a few more tiny changes (make sure to check some of my comments from my first review)!

Feel free to re-request when you're ready.

frontend/src/components/Common/Button/Button.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import styles from './Button.module.css';
+
+type ButtonProps = {
+  label?: string;
+  disabled?: boolean;
+  onClick?: () => void;
+  icon?: React.ReactNode;
+  iconLeft?: boolean;
+  iconRight?: boolean;
```
Bookie0, line 10: @SZCodes 

I think this logic is a bit flawed because ideally we should be able to use the button like this:


```
<Button iconLeft={<PlusIcon />} />
```

But currently `iconLeft` and `iconRight` only accept booleans.

I think for the purpose of shipping this, let's disregard it for now, but we should probably follow up afterwards.

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -153,32 +154,37 @@ const CandidateDecider: React.FC<CandidateDeciderProps> = ({ uuid }) => {
...
+              );
               if (nextCandidate !== null) {
                 navigateToNextCandidate(nextCandidate);
               }
             }}
-          >
-            Discard
-          </Button>
+            variant="primary"
+          />
```
Bookie0, line 187: nice!

Bookie0 (approved):
OK looks good! Thank you for taking the time to tackle the changes requested, Sophie!

Let's ship it!!

PR #927, opened 2025-04-21
Sz266/candidate decider selector

SZCodes:
### Summary 

Created Selector component for Candidate Decider based on Figma file [here](https://www.figma.com/design/qHNMOp6rrIFMv1wBXsVtYZ/SP25-IDOL-Candidate-Decider-Tool?node-id=76-558&m=dev)
* Created new files Selector.tsx, Selector.module.css
* Updated Selector in CandidateDecider.tsx
* Note that left column has increased width to fix the length of the selector component

### New button components compared to previous
|  | New | Old |
|-|-|-|
| Selector | Screenshot 2025-04-20 at 11 22 45 PM | Screenshot 2025-04-20 at 11 23 29 PM |

### Video demo of new component

https://github.com/user-attachments/assets/ed491018-2e14-48a6-87ba-5c89d86121e7


frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -13,18 +13,19 @@ import {
 import Button from '../Common/Button/Button';
 import Input from '../Common/Input/Input';
 import Switch from '../Common/Switch/Switch';
+import Selector from '../Common/Selector/Selector'; // adjust path as needed
```
Bookie0, line 16:
```suggestion
import Selector from '../Common/Selector/Selector';
```

not sure we need that comment?

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -13,18 +13,19 @@ import {
...
-  { value: 3, text: 'Maybe', color: 'yellow' },
-  { value: 4, text: 'Yes', color: 'green' },
-  { value: 5, text: 'Strong Yes', color: 'green ' },
-  { value: 0, text: 'Undecided', color: 'grey' }
+  { value: 1, text: 'Strong No', color: 'red' as const },
+  { value: 2, text: 'No', color: 'orange' as const },
+  { value: 3, text: 'Maybe', color: 'yellow' as const },
+  { value: 4, text: 'Yes', color: 'green' as const },
+  { value: 5, text: 'Strong Yes', color: 'darkgreen' as const },
+  { value: 0, text: 'Undecided', color: 'grey' as const }
```
Bookie0, line 28: wait why do we need `as const` here

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,99 @@
...
+.red {
+  color: var(--accent-dark-red, #d8001b);
+}
+
+.orange {
+  color: var(--accent-light-red, #ff676c);
+}
+
+.yellow {
+  color: var(--accent-yellow, #ff9300);
```
Bookie0, line 48: I noted this in your Button component PR, but we I don't think we need to add these fallback hex values when using CSS variables.

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,99 @@
...
+
+.grey {
+  color: var(--fg-dimmest, #666);
+}
+
+/* Selected states */
+.selected {
+  display: flex;
+  /* width: 450/6px; */
+  /* padding: 16px; */
```
Bookie0, line 67:
```suggestion
```

its good practice to remove commented out code like this unless there's a reason

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,99 @@
...
+  border-radius: 6px;
+  justify-content: center;
+  align-items: center;
+  gap: 10px;
+  /* flex: 1 0 0; */
```
Bookie0, line 72:
```suggestion
```

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,99 @@
...
+}
+
+.darkgreenSelected {
+  background: var(--accent-dark-green, #0a6e29);
+}
+
+.greySelected {
+  background: var(--border-default, #ebebeb);
+  color: var(--fg-default, #171717);
+}
```
Bookie0, line 99: Could you also add the custom focus states? It should be the `2px` solid blue outline (`accent-focus`)

Screenshot 2025-04-21 at 12 25 57 AM

frontend/src/components/Common/Selector/Selector.tsx
```diff
@@ -0,0 +1,37 @@
...
+      const selectedClass = isSelected ? `${styles.selected} ${styles[`${color}Selected`]}` : '';
+
+      return (
+        <button
+          key={value}
+          className={`${styles.option} ${baseClass} ${selectedClass}`}
+          onClick={() => onChange(value)}
+        >
+          {text}
+        </button>
```
Bookie0, line 31: So the problem with rendering a list of buttons is that they are all unique elements with no relation to each other. Notice how on the current Candidate decider in production, you can use `tab` to navigate to the choices, and then use left/right arrow keys to choose between the options.

That's because they're not using `<button>` but `<input type="radio"`, which allows this kind of keyboard functionality.

Screenshot 2025-04-21 at 12 28 49 AM


For accessibility and usability purposes, I think you should revamp this `<Selector>` component to use this type of `<input>`. You would then use CSS to customize it to make it "look" like a button.

Bookie0 (changes requested):
Thank you Sophie! This is a good start, just a few nits regarding the CSS, and the biggest change would be using `<input type="radio">` instead of the regular `<button>`

Thx! :)

frontend/src/components/Common/Selector/Selector.module.css
SZCodes: This time if I don't add them the color doesn't show up, not sure why

Bookie0: Oh that's because the variable names are wrong, check out the proper color names in `frontend/src/pages/App.css`

Eg `--accent-dark-red` should be `--accent-strong-no`

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,116 @@
+.selectorContainer {
+  display: flex;
+  width: 540px;
+  height: 54px;
+  padding: 4px;
+  gap: 2px;
```
Bookie0, line 6:
```suggestion
```

We don't need gap between the options according to the [spec](https://www.figma.com/design/qHNMOp6rrIFMv1wBXsVtYZ/SP25-IDOL-Candidate-Decider-Tool?node-id=122-938&m=dev). Plus, it gives a tiny bit more room for the options to grow! :)

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,116 @@
...
+
+.option {
+  display: flex;
+  padding: 16px;
+  justify-content: center;
+  align-items: center;
+  position: relative;
+  text-align: center;
+  gap: 10px;
+  flex: 1 0 0;
```
Bookie0, line 24: I think u copied them from Dev Mode in Figma, but I don't think we need these properties

```suggestion
```

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,116 @@
...
+  cursor: pointer;
+  background-color: transparent;
+  transition:
+    background-color 50ms,
+    color 50ms;
+  font-family: Inter, sans-serif;
+  font-size: 14px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 100%;
```
Bookie0, line 38:
```suggestion
  border-radius: 6px;
  cursor: pointer;
  font-family: Inter, sans-serif;
  font-size: 14px;
  font-weight: 500;
  line-height: 100%;
```

I think these styles might be from previously where you used `<button>` but i don't think you need all of them anymore.

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,116 @@
...
+  background: var(--accent-strong-yes);
+}
+
+.grey {
+  color: var(--fg-dimmest);
+}
+.greySelected {
+  background: var(--border-default);
+  color: var(--fg-default);
+}
```
Bookie0, line 83: This is a bit hard to read, can we make the color `--fg-default`?
Screenshot 2025-04-23 at 12 17 21 AM

frontend/src/components/Common/Selector/Selector.module.css
```diff
@@ -0,0 +1,116 @@
...
+  color: var(--accent-strong-yes);
+}
+.darkgreenSelected {
+  background: var(--accent-strong-yes);
+}
+
+.grey {
+  color: var(--fg-dimmest);
+}
+.greySelected {
```
Bookie0, line 80: Here too:

```suggestion
.gray {
  color: var(--fg-dimmest);
}
.graySelected {
```

frontend/src/components/Common/Selector/Selector.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import styles from './Selector.module.css';
+
+export interface RatingOptions {
+  value: number;
+  text: string;
+  color: 'red' | 'orange' | 'yellow' | 'green' | 'darkgreen' | 'grey';
```
Bookie0, line 7: And here :)

```suggestion
  color: 'red' | 'orange' | 'yellow' | 'green' | 'darkgreen' | 'gray';
```

frontend/src/components/Candidate-Decider/CandidateDecider.tsx
```diff
@@ -12,17 +12,18 @@ import {
...
 
-const ratings = [
+const ratings: RatingOptions[] = [
   { value: 1, text: 'Strong No', color: 'red' },
   { value: 2, text: 'No', color: 'orange' },
   { value: 3, text: 'Maybe', color: 'yellow' },
   { value: 4, text: 'Yes', color: 'green' },
-  { value: 5, text: 'Strong Yes', color: 'green ' },
+  { value: 5, text: 'Strong Yes', color: 'darkgreen' },
   { value: 0, text: 'Undecided', color: 'grey' }
```
Bookie0, line 27: Nit: since we're in an American setting, we should use American naming conventions lol

```suggestion
  { value: 0, text: 'Undecided', color: 'gray' }
```

Bookie0 (changes requested):
Thank you for your dedication, @ZhaoSophie! Just a few more tiny nits regarding accessibility (the color contrast for "Undecided") and code cleanliness (spelling, unused CSS properties), and we should be good to go! :)

frontend/src/components/Common/Selector/Selector.module.css
SZCodes: Screenshot 2025-04-29 at 11 52 23 AM

Do you mean the background like this?

frontend/src/components/Common/Selector/Selector.module.css
SZCodes: I removed all except line-height to keep the border around the buttons but lemme know if you want me to change this.

Screenshot 2025-04-29 at 12 12 07 PM
Screenshot 2025-04-29 at 12 12 23 PM

Bookie0: sounds good

Bookie0 (approved):
WOOH! All done, great job! :)

PR #931, opened 2025-04-22
[New Website] New `<ScrollingMarquee>` component 

Bookie0:
# Changes

- Used [React Fast Marquee](https://www.react-fast-marquee.com/) to implement a ScrollingMarquee component
- Added logos [according to Figma](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1325-12126&m=dev)
- Added component to test page
- Included gradients on edges of the page
- For accessibility, made the marquee not play by default if reduce motion user preferences are set to on

# Video

https://github.com/user-attachments/assets/e9917b46-1ba0-46b4-bda4-a4f3bdebfcd1

# Test plan

- Go to `/test-page`
- Verify that it works and scrolls like normal
- Hover on the carousel and ensure it stops
- Then head to System Settings > Accessibility > Display > Reduce motion and toggle it on
- Reload page, make sure the marquee doesn't scroll by default


new-dti-website-redesign/src/app/globals.css
```diff
@@ -179,3 +179,9 @@ section {
   outline: 2px solid var(--foreground-1);
   outline-offset: 3px;
 }
+
+/* Hacky way to avoid the gap between different
+subsections of the carousel */
+.controlChildDivWidth > div > div {
+  min-width: fit-content;
+}
```
Bookie0, line 187: This is basically to fix the bug that causes the subsections of the marquee to be too long (and empty):

Screenshot 2025-04-22 at 12 48 12 AM

new-dti-website-redesign/src/components/ScrolllingMarquee.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const logos: Logo[] = [
+  { src: '/products/logos/cuapts.svg', alt: 'CU Apartments logo', width: 110, height: 80 },
+  { src: '/products/logos/queuemein.svg', alt: 'Queue Me In logo', width: 80, height: 80 },
+  { src: '/products/logos/zing.svg', alt: 'Zing logo', width: 111, height: 100 },
+  { src: '/products/logos/cureviews.svg', alt: 'CU Reviews logo', width: 80, height: 80 },
+  { src: '/products/logos/cornellgo.svg', alt: 'CornellGo logo', width: 80, height: 80 },
+  { src: '/products/logos/courseplan.svg', alt: 'Courseplan logo', width: 60, height: 60 },
+  { src: '/products/logos/carriage.svg', alt: 'Carriage logo', width: 70, height: 70 },
```
Bookie0, line 19: we need custom dimensions since every icon has a different aspect ratio, so make it look visually/optically balanced, we need to set custom widths and heights

new-dti-website-redesign/src/components/ScrolllingMarquee.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const ScrollingMarquee = () => {
+  const [shouldPlay, setShouldPlay] = useState(true);
+
+  useEffect(() => {
+    const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
+    if (mediaQuery.matches) {
+      setShouldPlay(false);
+    }
+  }, []);
```
Bookie0, line 49: If this [reduce motion setting](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) is on, the marquee shouldn't autoplay by default.

This is an accessibility issue – some users prefer an interface that doesn't have motion.

new-dti-website-redesign/src/components/ScrolllingMarquee.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <div className="h-24">
+      <div className="w-full z-5 bg-background-1 border-t border-b border-border-1 absolute left-0 controlChildDivWidth">
```
Bookie0, line 53: the `<div className="h-24">` is to make set the element inside the main layout and take up place. 

Then the inner `<div>` has `absolute` position so that it can overlap on the left and right side. 

Then there's also the `controlChildDivWidth` for that width bug.

cchrischen (commented):
looks sick! the marquee (didn't even know it was called that) has a scrollbar for me.
image

The marquee also freezes when I try testing on iPad.

Bookie0:
> looks sick! the marquee (didn't even know it was called that) has a scrollbar for me. image


Hi @cchrischen! 

Ah oops, I didn't catch that because i didn't have my scrollbars set to "Always visible" in settings. Seems like it was an issue with one of the logos being too tall, so I reduced its size. 

Should be fixed now, thank you!

Screenshot 2025-04-22 at 2 59 40 AM

cchrischen (approved):

Bookie0:
Thanks so much Chris!

PR #932, opened 2025-04-22
Interview Scheduler Bug Bashing Fixes

cchrischen:
### Summary 
Bug Fixes:
*  Reinforced permission checking for update interview slot endpoint. Members and applicants should not be able to override each other, and only one user may get an approval message when signing up for the same slot on the frontend.
   * Leads are able to bypass this checking. This means that although IDOL subteam members are able to edit the time slot to whomever they want, the checks will still be done. Direct database edits will need to be done to make these changes if not a lead.
*  Fixed an issue where the backend would request the redacted scheduler information. 
*  Fixed an issue where applicants cannot sign up because the returned scheduler from the endpoint redacts all applicants.

### Test Plan 
Testing during worksesh!

### Notes 
Applicants should have much more robust permission checking, especially since this is open-source. This will be one of the only endpoints that allow non-IDOL members to interact with IDOL. It is imperative to make sure that outside users may not make unauthorized endpoint requests.

### DIFF
--- a/backend/src/API/interviewSchedulerAPI.ts
+++ b/backend/src/API/interviewSchedulerAPI.ts
@@ -1,6 +1,7 @@
+import { LEAD_ROLES } from '../consts';
 import InterviewSchedulerDao from '../dao/InterviewSchedulerDao';
 import InterviewSlotDao from '../dao/InterviewSlotDao';
-import { NotFoundError, PermissionError } from '../utils/errors';
+import { BadRequestError, NotFoundError, PermissionError } from '../utils/errors';
 import PermissionsManager from '../utils/permissionsManager';
 import { getMember } from './memberAPI';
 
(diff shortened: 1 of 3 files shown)


jujulcrane (approved):
Everything looks really good! I'm sure the leads will be very happy to have this next semester! 

Some minor things I noticed while testing out the create an interview scheduler instance:
- The inputted duration and members per slot doesn't clear to being empty after creating an interview scheduler instance.
- After creating an instance you have to reload for it to actually appear under All Interview Scheduler Instances (which might be annoying since you have to manually open it after creating it before you can use it)

For the actual Interview Scheduler 
- when you try to create a room and click save without actually adding a room, it doesn't notify you why save isn't working, instead nothing happens. It might be helpful to have some user feedback that "they need to add a room" before saving
- Theres also the bug you saw in work session where you can sign up multiple times in the same room and it duplicates 

https://github.com/user-attachments/assets/2fd81f33-0ad9-48bb-b085-a9348cf96ffe


- Sometimes the time slot doesn't turn green when you sign up for it

https://github.com/user-attachments/assets/e56f37d6-4067-4c1e-9484-1acfb6adcbb0

PR #934, opened 2025-04-23
[New Website] Added Feature Card Component

ladriennel:
### Summary 

Implemented FeatureCard for new website and added to test-components page

### Notion/Figma Link 

Figma: https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1336-13027&t=LggOx0DB0r0RzBaQ-0
Notion: https://www.notion.so/Team-Website-Task-Force-1be0ad723ce18020aad2fea7c24a58ef?p=1d30ad723ce18078ba84e6ebaf50d63b&pm=s

Screenshot 2025-04-22 at 6 10 30 PM


new-dti-website-redesign/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type FeatureCardProps = {
+  title: string;
+  body: string;
+  children: React.ReactNode;
+};
+
+export default function FeatureCard({ title, body, children }: FeatureCardProps) {
+  return (
+    <div className="p-[32px] max-w-[395px] border border-border-1 flex flex-col">
```
Bookie0, line 13: So the way the card works is that we'd have 3 of them on the same row, eg:

Screenshot 2025-04-22 at 11 57 33 PM


They would each fill 1/3rd of the container's width, so i think we should try avoid using hardcoded values for the width. Instead, we can probably use the Tailwind `w-1/3` [property](https://tailwindcss.com/docs/width). Then when using the card, the card's width would automatically be handled by the parent container.

(For the purpose of testing the component on the `/test-components` page, you can set a fixed value for this parent container, eg `118px).

Bookie0, line 13: Let's also try avoid using custom values when we have round numbers such as `32px`. 
```suggestion
    <div className="p-8 max-w-[395px] border border-border-1 flex flex-col">
```

new-dti-website-redesign/src/components/FeatureCard.tsx
```diff
@@ -0,0 +1,19 @@
...
+      <IconWrapper className="icon-wrapper">{children}</IconWrapper>
+      <h5 className="text-foreground-1 pt-[16px]">{title}</h5>
```
Bookie0, line 15: Since we're using `flex` on the card, we should leverage the `gap` property to space out elements instead of using `pt-[16px]`. 

According to the [specs](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1336-13026&m=dev), there should be `4px` of gap between the title and the body, and then `16px` between the icon and the text.

So you'll probably have to create another `<div>` that only contains the `<h5>` and the `<p>`.

new-dti-website-redesign/src/components/FeatureCard.tsx
Bookie0, line 13: For semantic reasons, we should use `<article>` instead of `<div>`. [Here's a better answer](https://teamtreehouse.com/community/when-to-use-div-section-and-article) that goes more into detail if ur interested! :)

new-dti-website-redesign/src/components/FeatureCard.tsx
Bookie0, line 15: Nit: we don't need to explicitly set `text-foreground-1` since we're already setting it in `globals.css` [here](https://github.com/cornell-dti/idol/blob/258edcebf9682e8bc360c95d09e29978cd1ef55e/new-dti-website-redesign/src/app/globals.css#L68) in `body`, so it transmits to the `<h5>` as well.

```suggestion
      <h5 className="pt-[16px]">{title}</h5>
```

new-dti-website-redesign/src/app/test-components/page.tsx
```diff
@@ -253,6 +254,78 @@ export default function TestComponents() {
           </IconWrapper>
         </div>
       </div>
+
+      <div className="flex flex-col gap-6">
+        <h4>Feature Card</h4>
+        <div className="flex gap-0">
```
Bookie0, line 260: Nit: by default, `flex` doesn't add any gap between child elements, so we don't need to to explicitly have `gap-0`
```suggestion
        <div className="flex">
```

Bookie0 (changes requested):
Awesome, this looks great @ladriennel! Just a few nits and comments regarding semantics and styling, but otherwise we can ship this fast! :)

ladriennel:
> Awesome, this looks great @ladriennel! Just a few nits and comments regarding semantics and styling, but otherwise we can ship this fast! :)

Thanks for the reviews Clément! I made changes based on the suggestions.

Bookie0:
> Thanks for the reviews Clément! I made changes based on the suggestions.

Great work @ladriennel! Component is looking perfect now

Bookie0 (approved):
nice nice nice

:shipit: !!! :)

PR #937, opened 2025-04-23
[New website] Updates to `<Button>`, `<Tabs>` component, heading semantic fixes 

Bookie0:
# Changes

- Made changes to `<Button>`:
  - Refactored into `forwardRef` to support native props better
  - Added `role` so that we can differentiate `role="tab"` and the regular `role="button"`
- `<Hero>` and `<CtaSection>` components: made the subheadings more semantic by using `<p>` instead of `<h6>`
- Created `<Tabs>` component that uses `<Button>`
  - `label` (str) and `content` (`React.ReactNode`) for the tab label and for the actual children of the tab panel
  - Added some logic to support proper keyboard navigation between tabs for accessibility
  - Used instances of `<Tabs>` on the `/test-components` page

# Video

https://github.com/user-attachments/assets/07dc0386-8d21-476d-8187-7a8a6e2c1ec7

# Test plan

- Head to `/test-components`
- Click the tabs and make sure they change the panel
- Use `tab` on your keyboard, and the left/right arrows to switch between tabs
- For tab panel 1 and tab panel 2, hitting `tab` should focus the panel itself. For tab panel 2, `tab` should focus the `<Input>`

### DIFF
--- a/new-dti-website-redesign/src/components/Tabs.tsx
+++ b/new-dti-website-redesign/src/components/Tabs.tsx
@@ -56,6 +56,23 @@ export default function Tabs({ tabs, className = '' }: TabsProps) {
         ))}
       </div>
 
+      {/* IMPORTANT: according to the W3C (official accessibility documentation, https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)
+       * When a tabpanel doesn't contain any focusable elements, the rendered should have tabindex="0" so that it's included in the tab sequence of the page. Make sure to also include focusState for the proper outline styling too!
+
+       Example implementation:
+        <Tabs
+          tabs={[
+            {
+              label: 'Tab 1',
+              content: (
+                <div className="focusState" tabIndex={0}>
+                  <p>No interactive content in this tab panel, so need tabIndex</p>
+                </div>
+              )
+            }
+          ]}
+        />
+       */}
       <div role="tabpanel" id={`panel-${activeIndex}`} aria-labelledby={`tab-${activeIndex}`}>
         {tabs[activeIndex]?.content}
       </div>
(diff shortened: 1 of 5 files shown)


new-dti-website-redesign/src/components/Tabs.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const tabsRef = useRef<(HTMLButtonElement | HTMLAnchorElement | null)[]>([]);
+
+  // You should be able to use the left/right arrow keys to navigate between tabs
+  // This piece of code handles keyboard navigation so that the tabs are accessible
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    if (e.key === 'ArrowRight' || e.key === 'ArrowLeft') {
+      e.preventDefault();
+      // +1 for going to right, -1 for going to left
+      const dir = e.key === 'ArrowRight' ? 1 : -1;
+      const nextIndex = (activeIndex + dir + tabs.length) % tabs.length;
```
cchrischen, line 27:
```suggestion
      const nextIndex = (activeIndex + dir) % tabs.length;
```

cchrischen (approved):
great! tabbing works well

new-dti-website-redesign/src/components/Tabs.tsx
Bookie0: thx!

Bookie0: oh wait @cchrischen this actually some functionality of using the arrow keys to change tabs. In the previous code, if you were on the first tab, pressing the left arrow would move you back to 3rd on the right (since it should wrap around)


|Before (good)|After|
|-|-|
|https://github.com/user-attachments/assets/baa295f0-38d7-4c6a-9e75-6e51831ecba7|https://github.com/user-attachments/assets/c8f21e53-cc8e-4eda-b8f2-64d90f437fcb|

cchrischen: oh good catch, negative mod smh

PR #946, opened 2025-04-30
[New Website] Added Footer Component

ladriennel:
### Summary 

Created Footer Component for new website

### Notion/Figma Link 

Figma: https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1322-2354&t=p9NPEHhcog8jBj64-0

Notion: https://www.notion.so/Team-Website-Task-Force-1be0ad723ce18020aad2fea7c24a58ef?p=1d30ad723ce1809d9188fab2b146a9ba&pm=s

Screenshot 2025-04-29 at 11 32 28 PM


new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+                label="Back to top"
+                variant="secondary"
+                badge={<Image src="/arrowLogo.svg" alt="Arrow Logo" width={25} height={25} />}
+                onClick={() => {
+                  window.scrollTo({
+                    top: 0,
+                    behavior: 'smooth'
+                  });
+                }}
+              />
```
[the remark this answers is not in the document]
Bookie0: I think ideally, this `<Button>` component should be flexible enough to accept any child (badge, image, icon, etc.) on the left or on the right. But that would probably require some extra configuration of the component itself, which isn't the point of the PR.

I do think we should avoid using `badge` prop on `<Button>` for now though! For a temporary fix, maybe the solution for now is to add single custom new prop on the component (eg: `backToTopArrow`). This will make the usecase pretty specific, but it should work for now.

Let me know what you think @ladriennel!

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+    </div>
+  );
+
+  return (
+    <div className="bg-background-1">
+      <div className="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-4 gap-0">
+        {/* 1) Logo */}
+        <div className="p-8 sm:pb-0 md:row-start-1 md:col-start-1 lg:row-auto lg:col-auto lg:pb-8">
+          <DTILogoSection />
+        </div>
```
Bookie0, line 192: For consistency, the padding should be `16px` to match the other sections and navbar at `breakpoint-md`



image

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+            label="Back to top"
+            variant="secondary"
+            badge={<Image src="/arrowLogo.svg" alt="Arrow Logo" width={25} height={25} />}
+            onClick={() => {
+              window.scrollTo({
+                top: 0,
+                behavior: 'smooth'
+              });
+            }}
+          />
```
Bookie0, line 231: There's a bug on smaller screens where the icon becomes super small:

image

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+        </div>
+      </div>
+    </div>
+  );
```
Bookie0, line 235: Also since the button is taking up spacing in the container, it's causing the other links to wrap on several lines which we don't want.

I think either you could make the button position `absolute` and then align it to the bottom right of the footer. And then in that case, you actually wouldn't 2 buttons to show/hide – you'd just have 1 single button where you'd change its `position` property.



Screenshot 2025-04-30 at 1 26 43 AM

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+      </ul>
+    </div>
+  );
+  const ProductLinksSection = () => (
+    <div>
+      <h6 className="text-foreground-1 pb-[8px]">Products</h6>
+      <ul className="flex flex-col list-none">
+        {ProductLinks.map(({ href, label }) => (
+          <li key={href}>
+            <Link href={href} className="text-foreground-3 hover:text-foreground-1">
```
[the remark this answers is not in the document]
Bookie0: And let's also make the link open in new tab (`target="_blank"`)

Bookie0 (changes requested):
Thank you, this is looking great Adrienne! Most of my comments are literally quite nits so they should be quick and ez to polish! :)

Biggest thing is probably regarding the SVG file, some responsiveness, and potentially some code refactoring.

dti-website/next-env.d.ts
```diff
@@ -1,5 +1,4 @@
 /// <reference types="next" />
-/// <reference types="next/types/global" />
```
Bookie0, line 2: also is this intentional? it says int he `NOTE` that this file shouldn't be edited

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+              {label}
+            </Link>
+          </li>
+        ))}
+        <li>
+          <span className="text-foreground-3 cursor-default">Carriage</span>
+        </li>
+        <li>
+          <span className="text-foreground-3 cursor-default">Cornell Go</span>
+        </li>
```
Bookie0, line 156: wait also shouldn't these be links too? In the future we'll have URLs, for now feel free to leave the `href` empty

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+        height={50}
+        className="min-w-[100px]"
+      />
+      <ul className="flex max-w-[220px] justify-between list-none pt-4">
+        {SocialLinks.map(({ href, label, icon }) => (
+          <li key={href}>
+            <Link
+              href={href}
+              className="text-foreground-3 hover:text-foreground-1"
+              aria-label={label}
```
Bookie0, line 176: also yay for `aria-label`!!

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+        {/* 4) Products */}
+        <div className="p-8 sm:pt-0 md:row-start-2 md:col-start-2 md:pt-8 lg:row-auto lg:col-auto">
+          <div className="flex justify-between items-end">
+            <ProductLinksSection />
+            <div className="md:hidden">
+              <Button
+                label="Back to top"
+                variant="secondary"
+                badge={<Image src="/arrowLogo.svg" alt="Arrow Logo" width={25} height={25} />}
```
Bookie0, line 207: Since this icon is purely decorative, we don't need the alt text here (just leave the `alt` attribute but make the value empty). Plus we already have the "Back to top" label on the button itself, which should convey sufficient information.

Another reason we don't want the alt text for this icon is that when using the a screen reader software, it'll pause twice on this element – once on the button, and then once again on the image describe it. But in this case, pausing twice is redundant.

[For reference](https://designsystem.digital.gov/components/icon/accessibility-tests/#:~:text=Icons%20provide%20a%20text%20alternative,should%20not%20hear%20a%20description.) :)

Bookie0:
```suggestion
                badge={<Image src="/arrowLogo.svg" alt="" width={25} height={25} />}
```

dti-website/next-env.d.ts
ladriennel: No I'm not sure what changed that, but I think it was a command I ran to fix one of the build/fix errors.

Bookie0: oh hmm.. i think we should avoid doing that? What kind of error was it? If it's a lint or format thing, ideally we should try and fix the error instead of doing a workaround thing like this?

not sure tho

new-dti-website-redesign/src/components/Footer.tsx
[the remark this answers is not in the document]
ladriennel: Sounds good! For now I just added a backToTop prop which renders the icon on left if its passed through.

new-dti-website-redesign/src/components/Footer.tsx
```diff
@@ -0,0 +1,232 @@
...
+          />
+        }
+      </Link>
+
+      <ul className="flex max-w-[220px] justify-between list-none pt-4">
+        {SocialLinks.map(({ href, label, icon }) => (
+          <li key={href}>
+            <Link
+              href={href}
+              className="text-foreground-3 hover:text-foreground-1 inline-flex rounded-sm focusState"
```
Bookie0, line 190:
```suggestion
              className="text-foreground-3 hover:text-foreground-1 inline-flex rounded-sm focusState transition-[color] duration-[120ms]"
```

just for a smoother transition! :)

Bookie0 (approved):
Hi @ladriennel! Just a tiny transition duration, but otherwise everything looks really good!
Thank you for taking the time to fix the issue. 

LGTM!

PR #957, opened 2025-05-13
New Instance and Delete Instance Modals

jujulcrane:
### Summary 

* Added a modal to replace window alert for creating a new Interview Status Instance
* Added a modal that confirms deletion of an instance

### Test Plan 

https://github.com/user-attachments/assets/f128917e-fbc5-45e4-b520-43bb0e9f8397

* Click '+ New Empty Instance': Verify modal appears
* Enter new instance name and click create: Verify new empty instance is created with the correct name
* Delete an instance: Verify that when clicking "No" the instance does not delete, and when clicking "Yes, delete" the instance is deleted


frontend/src/components/Modals/InterviewStatusNewInstanceModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Modal open={open} onClose={onClose} size="small" closeIcon>
+      <Modal.Header>Create New Instance</Modal.Header>
+      <Modal.Content>
+        <Input
+          fluid
+          placeholder="Enter instance name..."
+          value={name}
+          onChange={(e) => setName(e.currentTarget.value)}
+        />
```
clementroze, line 46: If possible, could we start using some of the new components that i made for Cand Decider?

Eg `<Input>` in components/Common/Input/Input.tsx`

That way we can have the fresh, brand new components that are more aesthetically up to date and also more accessible!

Here's an example of how the new Input component looks. This is the multiline variant (i.e: `<textarea>`), but the default variant should work fine for this use case.

Screenshot 2025-05-12 at 11 34 29 PM

frontend/src/components/Modals/InterviewStatusNewInstanceModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+      </Modal.Content>
+      <Modal.Actions>
+        <Button onClick={onClose}>Cancel</Button>
+        <Button positive onClick={handleCreate}>
+          Create
+        </Button>
```
clementroze, line 52: Same here! :) 

if it isn't too much, I think it could be good to using the new buttons in `components/Common/Button/Button.tsx`

I'd also love to hear your reasoning behind using the `positive` prop which makes the button green. I feel like since this button doesn't necessarily indicate a positive/successful action (since it just creates a new instance), we should just use the default accent color. 

In Semantic UI, it seems like the default accent color is blue ("emphasis"):

Screenshot 2025-05-12 at 11 37 35 PM

In the new `<Button>` component, we use a black color:

Screenshot 2025-05-12 at 11 37 52 PM

frontend/src/components/Modals/InterviewStatusDeleteInstanceModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <Modal open={open} onClose={onClose} size="small" closeIcon>
+      <Modal.Header>Confirm Delete</Modal.Header>
+      <Modal.Content>
+        <p>Are you sure you want to delete instance “{instanceName}” and all its statuses?</p>
+      </Modal.Content>
+      <Modal.Actions>
+        <Button onClick={onClose} disabled={isDeleting}>
+          No
```
clementroze, line 36:
```suggestion
          Cancel
```


I think using "Cancel" matches more common UI terminology for dismissive actions? It also communicates the intent of the button more clearly (ie: exiting the modal without doing anything)


Lmk what you think!

frontend/src/components/Modals/InterviewStatusDeleteInstanceModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </Button>
+        <Button negative onClick={handleDelete} loading={isDeleting}>
+          Yes, delete
```
clementroze, line 39:
```suggestion
          Delete instance
```

Here too, I'd avoid using words like "Yes" because they rely on users reading and processing the entire modal text (especially the "Are you sure you want to..." question) to understand what they're agreeing to. 

Using action-specific button labels like "Delete instance" make the choice clearer and make it explicit what exactly you're deleting.

clementroze (commented):
Hmm i think there was an issue with git or smth because this PR has all the changes from [your initial PR that you already merged](https://github.com/cornell-dti/idol/pull/955/files)? I think build the PR off of the old branch or smth.


Lmk when u get that resolved and i'll take another look at the code + test it out.


For now, just a few preliminary thoughts regarding using the new design system components + some of the button labels! :)

frontend/src/components/Modals/InterviewStatusNewInstanceModal.tsx
jujulcrane: Thanks for pointing that out Clément! I usually associate green with creation as it contrasts with deletion which is usually red but I see your point on how it doesn't really match our color scheme and will replace with the new button instead to stay consistent!

clementroze: Ah ok gotcha, that’s fair and I see where ur coming from. That said I do usually see green more commonly tied to success states like confirmations or completed actions, so I worry it might send the wrong signal. Eg for some of your confirmation modals here, I think using green for the button makes sense!

IMG_4559
IMG_4560


That’s why i think sticking with the default accent color would keep things more neutral + aligned with the rest of the UI.

but I appreciate your flexibility on this! :)

cchrischen (approved):
Great work! There is currently no indication on which dashboard you're currently on. It would be nice to know somewhere on the header when you have a group selected.
image

clementroze (approved):
nice ty for the button + input component changes!

clementroze (approved):
good to go Juju!

jujulcrane:
> Great work! There is currently no indication on which dashboard you're currently on. It would be nice to know somewhere on the header when you have a group selected. image

Ah yes good point, I'll make sure the instance name is visible!

PR #958, opened 2025-05-13
[New Website] Team Card 

cchrischen:
### Summary 
Implemented the team card component, along with the member details component. Added components on test page.

### Notion/Figma Link 
[Figma](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1336-13034&m=dev)

### Test Plan 

https://github.com/user-attachments/assets/424c87c5-7532-40de-8e26-bb79f9cc2d32

### DIFF
--- a/new-dti-website-redesign/src/components/TeamCard.tsx
+++ b/new-dti-website-redesign/src/components/TeamCard.tsx
@@ -1,0 +1,132 @@
+import { getColorClass, productLinks } from '@/utils/memberUtils';
+import Image from 'next/image';
+import Link from 'next/link';
+
+const MemberSummary = ({
+  user,
+  image,
+  enlarged = false
+}: {
+  user: IdolMember;
+  image: string;
+  enlarged?: boolean;
+}) => (
+  <>
+    <Image
+      src={image}
+      alt={`${user.firstName} ${user.lastName}`}
+      className="rounded-lg w-auto h-auto"
+      width={232}
+      height={232}
+    />
+    <div className="text-left">
+      {enlarged ? (
+        <h3>{`${user.firstName} ${user.lastName}`}</h3>
+      ) : (
+        <h6>{`${user.firstName} ${user.lastName}`}</h6>
+      )}
+      <p className={`text-${getColorClass(user.role)}`}>{user.roleDescription}</p>
+    </div>
+  </>
+);
+
+type MemberCardProps = {
+  user: IdolMember;
+  image: string;
+  selected: boolean;
+  onClick?: () => void;
+};
+
+export const MemberCard = ({ user, image, selected, onClick }: MemberCardProps) => {
+  const baseStyles =
+    'p-8 border border-border-1 flex flex-col gap-4 focusState hover:shadow-white hover:shadow-[0_-4px_8px_0]';
+
+  return (
+    <button
+      className={`${baseStyles} ${selected ? 'border-b-foreground-1' : ''}`}
+      onClick={onClick}
+    >
+      <MemberSummary user={user} image={image} />
+    </button>
+  );
+};
+
+type MemberDetailsProps = {
+  user: IdolMember;
+  image: string;
+};
+export const MemberDetailsCard = ({ user, image }: MemberDetailsProps) => {
+  const baseStyles = 'w-1/2 p-8 border border-border-1 flex flex-col';
+  return (
+    <div className="flex w-full">
+      <div className={`${baseStyles} gap-4`}>
+        <MemberSummary user={user} image={image} enlarged />
+      </div>
+      <div className={`${baseStyles} gap-8`}>
+        <div className="flex gap-8">
+          <div className="w-1/2">
+            <p className="text-foreground-3">Graduating</p>
+            <p>{user.graduation}</p>
+          </div>
+          <div className="w-1/2">
+            <p className="text-foreground-3">Major</p>
+            <p>{`${user.major}${user.doubleMajor ? ` & ${user.doubleMajor}` : ''}`}</p>
+          </div>
+        </div>
+        <div className="flex gap-8">
+          <div className="w-1/2">
+            <p className="text-foreground-3">Hometown</p>
+            <p>{user.hometown}</p>
+          </div>
+          <div className="w-1/2">
+            <p className="text-foreground-3">Subteam</p>
+            {user.subteams[0] in productLinks ? (
+              <Link className="flex gap-2 items-center" href={productLinks[user.subteams[0]].link}>
+                <p className="underline">{productLinks[user.subteams[0]].name}</p>
+                <svg
+                  xmlns="http://www.w3.org/2000/svg"
+                  width="16"
+                  height="16"
+                  viewBox="0 0 16 16"
+                  fill="none"
+                >
+                  <path
+                    d="M10 2H14V6"
+                    stroke="white"
+                    stroke-width="1.33333"
+                    stroke-linecap="round"
+                    stroke-linejoin="round"
+                  />
+                  <path
+                    d="M6.66602 9.33333L13.9993 2"
+                    stroke="white"
+                    stroke-width="1.33333"
+                    stroke-linecap="round"
+                    stroke-linejoin="round"
+                  />
+                  <path
+                    d="M12 8.66667V12.6667C12 13.0203 11.8595 13.3594 11.6095 13.6095C11.3594 13.8595 11.0203 14 10.6667 14H3.33333C2.97971 14 2.64057 13.8595 2.39052 13.6095C2.14048 13.3594 2 13.0203 2 12.6667V5.33333C2 4.97971 2.14048 4.64057 2.39052 4.39052C2.64057 4.14048 2.97971 4 3.33333 4H7.33333"
+                    stroke="white"
+                    stroke-width="1.33333"
+                    stroke-linecap="round"
+                    stroke-linejoin="round"
+                  />
+                </svg>
+              </Link>
+            ) : (
+              <div>{productLinks[user.subteams[0]].name}</div>
+            )}
+          </div>
+        </div>
+        <div>
+          <p className="text-foreground-3">About</p>
+          <div className="flex flex-col gap-2">
+            {user.about.split('\n').map((para, i) => (
+              <p key={i}>{para}</p>
+            ))}
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
--- a/new-dti-website-redesign/src/utils/memberUtils.ts
+++ b/new-dti-website-redesign/src/utils/memberUtils.ts
@@ -1,0 +1,91 @@
+export const getGeneralRole = (role: Role): GeneralRole => {
+  switch (role) {
+    case 'ops-lead':
+      return 'lead';
+    case 'product-lead':
+      return 'lead';
+    case 'dev-lead':
+      return 'lead';
+    case 'design-lead':
+      return 'lead';
+    case 'business-lead':
+      return 'lead';
+    case 'tpm':
+      return 'developer';
+    case 'apm':
+      return 'pm';
+    case 'pm-advisor':
+      return 'pm';
+    case 'dev-advisor':
+      return 'developer';
+    case 'design-advisor':
+      return 'designer';
+    case 'business-advisor':
+      return 'business';
+    default:
+      return role;
+  }
+};
+
+export const getColorClass = (role: Role): string => {
+  const generalRole = getGeneralRole(role);
+  const colors: { [key in GeneralRole]: string } = {
+    lead: 'red',
+    pm: 'purple',
+    developer: 'green',
+    designer: 'blue',
+    business: 'yellow'
+  };
+  return `accent-${colors[generalRole]}`;
+};
+
+export const productLinks: { [key: string]: { name: string; link: string } } = {
+  courseplan: {
+    name: "CoursePlan",
+    link: "https://courseplan.io/"
+  },
+  reviews: {
+    name: "CUReviews",
+    link: "https://www.cureviews.org/"
+  },
+  queuemein: {
+    name: "Queue Me In",
+    link: "https://queueme.in/"
+  },
+  cuapts: {
+    name: "CU Apts",
+    link: "https://www.cuapts.org/"
+  },
+  idol: {
+    name: "IDOL",
+    link: "https://www.cornelldti.org"
+  },
+  cornellgo: {
+    name: "Cornell GO",
+    link: ""
+  },
+  carriage: {
+    name: "Carriage",
+    link: ""
+  },
+  leads: {
+    name: "Lead",
+    link: ""
+  },
+  business: {
+    name: "Business",
+    link: ""
+  },
+  curaise: {
+    name: "CU Raise",
+    link: ""
+  },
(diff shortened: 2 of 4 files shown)


new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+      width={232}
+      height={232}
+    />
+    <div className="text-left">
+      {enlarged ? (
+        <h3>{`${user.firstName} ${user.lastName}`}</h3>
+      ) : (
+        <h6>{`${user.firstName} ${user.lastName}`}</h6>
+      )}
+      <p className={`text-${getColorClass(user.role)}`}>{user.roleDescription}</p>
```
clementroze, line 28: If its the enlarged version can we make the description a `<h5>` element

Screenshot 2025-05-13 at 4 03 09 PM

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+  enlarged = false
+}: {
+  user: IdolMember;
+  image: string;
+  enlarged?: boolean;
+}) => (
+  <>
+    <Image
+      src={image}
+      alt={`${user.firstName} ${user.lastName}`}
```
clementroze, line 17:
```suggestion
      alt={`${user.firstName} ${user.lastName}'s profile picture`}
```

(just to make the alt text a bit more descriptive)

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+type MemberCardProps = {
+  user: IdolMember;
+  image: string;
+  selected: boolean;
+  onClick?: () => void;
+};
+
+export const MemberCard = ({ user, image, selected, onClick }: MemberCardProps) => {
+  const baseStyles =
+    'p-8 border border-border-1 flex flex-col gap-4 focusState hover:shadow-white hover:shadow-[0_-4px_8px_0]';
```
clementroze, line 42: Ah sorry if this wasn't clear enough in the [specs](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1336-13031&m=dev) – I was thinking that on hover the card has a background color change (the shadow only appears when the card is selected/open)


```suggestion
    'p-8 border border-border-1 flex flex-col gap-4 focusState hover:bg-background-2 transition-[background-color] duration-[120ms]';
```

also added a transition that is consistent with the button hover states

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  return (
+    <button
+      className={`${baseStyles} ${selected ? 'border-b-foreground-1' : ''}`}
+      onClick={onClick}
+    >
+      <MemberSummary user={user} image={image} />
+    </button>
```
clementroze, line 50: Hmm I'm not sure if it's valid to put content like `<p>` or `<div>` inside of a `<button>` tag.

[According to docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#technical_summary), `<button>` can only accept phrasing content inside as children.


[Phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#phrasing_content) doesn't include things like `<p>` or `<div>` (those are [Flow content](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#flow_content) tags)



I wonder if you know any best practices around making cards interactive and clickable without putting wrapping it all in a `<button>` tag?

If it's not too much effort, here's what I would do:

- Use `<article>` for the main parent card ([specs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/article)).
- Have an invisible `<button>` element somewhere inside. Use CSS to overlay a psuedo class (`::after`) on top of the card for a bigger tap target, and keep the interactive states (hover/focus) on that button
- Have the rest of the inner elements stay where they are


The reason we should keep `<button>` is because it comes with built-in accessibility and behavior, like focusable by default, triggers with keyboard, and it's also semantic for screen readers.

clementroze: At Replit we ran into a similar problem concerning cards that had to be clickable but had inner children.

It's not exactly the same thing because we used `<a>` instead of `<button>`, but it works the same way. 

As you can see, the `<a>` is wrapped around the name of the card. Then there's a `::after` with some specific CSS styles that make it cover the entire card


Screenshot 2025-05-13 at 4 21 36 PM

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+        <div className="flex gap-8">
+          <div className="w-1/2">
+            <p className="text-foreground-3">Hometown</p>
+            <p>{user.hometown}</p>
+          </div>
+          <div className="w-1/2">
+            <p className="text-foreground-3">Subteam</p>
+            {user.subteams[0] in productLinks ? (
+              <Link
+                className="flex gap-2 items-center"
```
clementroze, line 85:
```suggestion
                className="flex gap-2 items-center focusState w-fit"
```

for custom focus state and also to make the link not span the entire width of the col

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+                target="_blank"
+                href={productLinks[user.subteams[0]].link}
+              >
+                <p className="underline">{productLinks[user.subteams[0]].name}</p>
```
clementroze, line 89:
```suggestion
                <p className="underline underline-offset-3">{productLinks[user.subteams[0]].name}</p>
```

just to increase space between underline + text so its not too tight! :)

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+          </div>
+        </div>
+        <div>
+          <p className="text-foreground-3">About</p>
+          <div className="flex flex-col gap-2">
+            {user.about.split('\n').map((para, i) => (
+              <p key={i}>{para}</p>
+            ))}
+          </div>
+        </div>
```
clementroze, line 132: For the contact icons + "Chat with me", do you plan on adding that later?

Screenshot 2025-05-13 at 4 26 13 PM

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+};
+
+type MemberDetailsProps = {
+  user: IdolMember;
+  image: string;
+};
+export const MemberDetailsCard = ({ user, image }: MemberDetailsProps) => {
+  const baseStyles = 'w-1/2 p-8 border border-border-1 flex flex-col';
+  return (
+    <div className="flex w-full">
```
clementroze, line 61: Also for the row of grids above and below the enlarged card, as well as the Close button, do you plan on implementing those within this component, or do you think we should have it directly on the `/team` page?

image

new-dti-website-redesign/src/utils/memberUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const getColorClass = (role: Role): string => {
+  const generalRole = getGeneralRole(role);
+  const colors: { [key in GeneralRole]: string } = {
+    lead: 'red',
+    pm: 'purple',
+    developer: 'green',
+    designer: 'blue',
+    business: 'yellow'
+  };
```
clementroze, line 38: nice!

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+  onClick?: () => void;
+};
+
+export const MemberCard = ({ user, image, selected, onClick }: MemberCardProps) => {
+  const baseStyles =
+    'p-8 border border-border-1 flex flex-col gap-4 focusState hover:shadow-white hover:shadow-[0_-4px_8px_0]';
+
+  return (
+    <button
+      className={`${baseStyles} ${selected ? 'border-b-foreground-1' : ''}`}
```
clementroze, line 46:
```suggestion
      className={`${baseStyles} ${selected ? 'bg-background-2 border-b-foreground-1' : ''}`}
```

selected state should have slightly elevated background color

clementroze: Also can we add a shadow like the one on the navbar for the currently selected card:

Screenshot 2025-05-13 at 4 34 54 PM


I think the way [I implemented it for the navbar](https://vscode.dev/github/cornell-dti/idol/blob/cc/team-card/new-dti-website-redesign/src/components/Navbar.tsx#L49-L50) was by using an `::after` pseudo-element and then some custom shadow properties

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,136 @@
...
+};
+
+type MemberDetailsProps = {
+  user: IdolMember;
+  image: string;
+};
+export const MemberDetailsCard = ({ user, image }: MemberDetailsProps) => {
+  const baseStyles = 'w-1/2 p-8 border border-border-1 flex flex-col';
+  return (
+    <div className="flex w-full">
```
clementroze, line 61: Could you also make it responsive too pls?

Screenshot 2025-05-13 at 4 38 57 PM


[Here's](https://www.figma.com/design/ttAGEX3pHmzuhMzp8uAyhz/SP25-Cornelldti.org-Website-Revamp?node-id=1336-20185&m=dev) what it should look like ideally:
Screenshot 2025-05-13 at 4 37 58 PM

clementroze (commented):
Hi Chris! Thanks for your work on this! 

This is a great first pass at the component. Just left a few comments about small CSS nits (i added some code suggestions if it's helpful!), about the use of `<button>` for the card, and couple questions on the social media icons + close functionality.

Lmk if you wanna talk about any of these, thanks! :)

new-dti-website-redesign/src/components/TeamCard.tsx
cchrischen: I think the client should be responsible for implementing the open/close functionality. This is coupled with how the grid is laid out and where we want to render the details card, so it should be best left for the `/team` page.

new-dti-website-redesign/src/components/TeamCard.tsx
cchrischen: I did some weird stuff to get the focus state to work, let me know if it works lol

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,243 @@
...
+type MemberCardProps = {
+  user: IdolMember;
+  image: string;
+  selected: boolean;
+  onClick?: () => void;
+};
+
+export const MemberCard = ({ user, image, selected, onClick }: MemberCardProps) => {
+  const baseStyles =
+    'relative p-8 border border-border-1 flex flex-col gap-4 hover:bg-background-2 transition-[background-color] duration-[120ms] has-[:focus]:outline-2 has-[:focus]:outline-offset-3 has-[:focus]:z-10';
```
clementroze, line 48:
```suggestion
    'relative p-8 border border-border-1 flex flex-col gap-4 hover:bg-background-2 transition-[background-color] duration-[120ms] has-[:focus-visible]:outline-2 has-[:focus-visible]:outline-offset-3 has-[:focus-visible]:z-10';
```


So that the white outline appears the element is focused via `tab` only, not on click.

[See `:focus` vs `:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#focus_vs_focus-visible)

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  return (
+    <article
+      className={`${baseStyles} ${selected ? `bg-background-2 border-b-foreground-1 after:content-[''] after:absolute after:bottom-[-1px] after:left-0 after:w-full after:h-[1px] after:bg-foreground-1 after:shadow-[0_-4px_8px_0_var(--foreground-1)] after:rounded-full` : ''}`}
+    >
+      <MemberSummary user={user} image={image} />
+      <button
+        className="opacity-0 cursor-pointer after:content-[''] after:absolute after:top-0 after:left-0 after:w-full after:h-full focus:outline-4"
```
clementroze, line 56:
```suggestion
        className="opacity-0 cursor-pointer after:content-[''] after:absolute after:top-0 after:left-0 after:w-full after:h-full"
```

Not sure we need the focus here since it's being applied on `<article>` right?

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,243 @@
...
+              {user.about.split('\n').map((para, i) => (
+                <p key={i}>{para}</p>
+              ))}
+            </div>
+          </div>
+        </div>
+        <div className="flex justify-between p-8 border border-transparent border-t-border-1 ">
+          <div className="flex gap-4 items-center">
+            {user.website && (
+              <Link href={user.website}>
```
clementroze, line 149: I noticed links don't have a focus state by default, so instead of adding the `focusState` class for each of the links, I added a global style for all `<a>` in [this commit](https://github.com/cornell-dti/idol/pull/958/commits/b020fa45f9b0ad23e976e0126d83de616a053944) :)

clementroze: |Before|After|
|-|-|
|Screenshot 2025-05-17 at 1 37 03 AM|Screenshot 2025-05-17 at 1 36 53 AM|

new-dti-website-redesign/src/components/TeamCard.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+type MemberDetailsProps = {
+  user: IdolMember;
+  image: string;
+};
+export const MemberDetailsCard = ({ user, image }: MemberDetailsProps) => {
+  const baseStyles = 'md:w-1/2 border border-border-1 flex flex-col';
+  return (
+    <div className="flex w-full flex-col md:flex-row">
+      <div className={`${baseStyles} p-8 gap-4`}>
```
clementroze, line 71:
```suggestion
      <div className={`${baseStyles} p-8 gap-4 border-r-0`}>
```

To remove the double border on the right side of the left column

clementroze (approved):
Everything looks great! Just a few last nits and then good to go!

Thank you!

new-dti-website-redesign/src/components/TeamCard.tsx
clementroze: Ok sounds good!

new-dti-website-redesign/src/components/TeamCard.tsx
clementroze: Works + looks good to me!

clementroze (approved):
thanks for addressing the nits!

### :shipit: :shipit: :shipit:

PR #964, opened 2025-05-17
[New Website] Smoothly animation page links in `<Navbar>`

clementroze:
# Summary

- The current way of showing which page you were currently on in the navbar involved just a white underline, which wasn't very visible. it also didn't animate when you changed page, which made it kinda choppy

Screenshot 2025-05-17 at 1 23 09 PM

- This new design uses a pill `<span>` rendered independently of the link. It also uses `useLayoutEffect` to smoothly animate between each page
- For stylistic reasons, the pill highlight should be the same size as the "Apply" button. So I customized the `<Button>` component a bit to add a `small` prop, which is used in `<Navbar>`
- I also had to move `<Navbar>` to the root `layout.tsx` page so that the highlight animation persists across pages and doesn't reset

# Video

https://github.com/user-attachments/assets/0c357822-6909-46d5-a0ad-091915ea00f1

# Test plan
- Navigate to each page in navbar
- For all page that aren't Home or Apply, make sure the pill highlight correctly animates as seen in the video
- When switching from one of those links to Home/Apply, make sure the pill highlight disappears
- When switching from Home/Apply to any of those other pages, make sure the pill highlight comes back
- Ensure no breaking changes in mobile nav (resize window to check – the hamburger menu logic shouldn't be impacted by this)

# Note
- I added lots of inline code comments to try and explain the logic on the animation – it's a bit complex to ensure we cover edge cases and to even make this animation smooth

### DIFF
--- a/new-dti-website-redesign/src/app/layout.tsx
+++ b/new-dti-website-redesign/src/app/layout.tsx
@@ -1,6 +1,6 @@
 import type { Metadata } from 'next';
 import './globals.css';
-import Navbar from '@/components/Navbar';
+import Navbar from '../components/Navbar';
 
 export const metadata: Metadata = {
   title: 'Create Next App',
--- a/new-dti-website-redesign/src/app/layout.tsx
+++ b/new-dti-website-redesign/src/app/layout.tsx
@@ -1,5 +1,6 @@
 import type { Metadata } from 'next';
 import './globals.css';
+import Navbar from '@/components/Navbar';
 
 export const metadata: Metadata = {
   title: 'Create Next App',
@@ -13,7 +14,10 @@ export default function RootLayout({
 }>) {
   return (
     <html lang="en">
-      <body>{children}</body>
+      <body>
+        <Navbar />
+        {children}
+      </body>
     </html>
   );
 }
--- a/new-dti-website-redesign/src/components/Button.tsx
+++ b/new-dti-website-redesign/src/components/Button.tsx
@@ -8,6 +8,7 @@ type ButtonProps = {
   label: string;
   href?: string;
   variant?: 'primary' | 'secondary' | 'tertiary';
+  size?: 'default' | 'small';
   badge?: React.ReactNode;
   backToTop?: React.ReactNode;
   newTab?: boolean;
@@ -23,6 +24,7 @@ const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonProps>(
       href,
       className = '',
       variant = 'primary',
+      size = 'default',
       badge,
       backToTop,
       newTab = false,
@@ -30,8 +32,7 @@ const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonProps>(
     },
     ref
   ) => {
-    const baseStyles = `
-        px-6 h-12 w-fit rounded-full cursor-pointer inline-flex items-center justify-center gap-2
+    const baseStyles = `w-fit rounded-full cursor-pointer inline-flex items-center justify-center gap-2
         transition-[background-color] duration-[120ms] focusState text-nowrap`;
 
     const variantStyles = {
@@ -42,7 +43,12 @@ const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonProps>(
       tertiary: `bg-transparent border border-border-1 text-foreground-1 hover:bg-background-2`
     }[variant];
 
-    const sharedClasses = `${baseStyles} ${className} ${variantStyles}`;
+    const sizeStyles = {
+      default: 'px-6 h-12',
+      small: 'px-4 h-10'
+    }[size];
+
+    const sharedClasses = `${baseStyles} ${className} ${variantStyles} ${sizeStyles}`;
 
     const content = (
       <>
--- a/new-dti-website-redesign/src/components/Layout.tsx
+++ b/new-dti-website-redesign/src/components/Layout.tsx
@@ -1,7 +1,6 @@
 'use client';
 
 import React, { ReactNode } from 'react';
-import Navbar from './Navbar';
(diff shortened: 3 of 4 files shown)


new-dti-website-redesign/src/app/layout.tsx
```diff
@@ -13,7 +14,10 @@ export default function RootLayout({
 }>) {
   return (
     <html lang="en">
-      <body>{children}</body>
+      <body>
+        <Navbar />
+        {children}
+      </body>
```
clementroze, line 20: So essentially we're moving `<Navbar>` from `src/components/Layout.tsx` into the root `layout.tsx`  in `src/app`.

This is because we wanna avoid unnecessary re-renders. If `<Navbar>` is defined in a page-specific or in a layout component used per page, it'll get remounted on every route change. This breaks transitions and we won't smooth highlight animation.

new-dti-website-redesign/src/components/Button.tsx
```diff
@@ -42,7 +43,12 @@ const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonProps>(
       tertiary: `bg-transparent border border-border-1 text-foreground-1 hover:bg-background-2`
     }[variant];
 
-    const sharedClasses = `${baseStyles} ${className} ${variantStyles}`;
+    const sizeStyles = {
+      default: 'px-6 h-12',
+      small: 'px-4 h-10'
+    }[size];
```
clementroze, line 49: just some new props/styling for a smaller button variant to fit inside the nav

new-dti-website-redesign/src/components/Layout.tsx
```diff
@@ -11,8 +10,6 @@ type Props = {
 
 const Layout = ({ children }: Props): React.ReactElement => (
   <>
-    <Navbar />
-
```
clementroze, line 15: See here

new-dti-website-redesign/src/components/Navbar.tsx
```diff
@@ -19,6 +19,69 @@ export default function Navbar() {
     { href: '/sponsor', label: 'Sponsor' }
   ];
 
+  // to only show the highlight on the links above (not on Home or Apply)
```
clementroze, line 22: I hope the inline comments make sense here, but at a high-level, all of this handles the animated highlight pill that tracks the selected navigation link. It aims to provide smooth transitions while managing edge cases like moving in/out of the non-nav pages (Home and Apply)

new-dti-website-redesign/src/components/Navbar.tsx
```diff
@@ -19,6 +19,69 @@ export default function Navbar() {
...
+    if (linkEl) {
+      // measure position + width of actual link
+      // and then create new style for positioning the new highlight
+      const { offsetLeft, offsetWidth } = linkEl;
+      const newStyle = { left: offsetLeft, width: offsetWidth };
+
+      const cameFromHomeOrApply = prevPathname.current === '/' || prevPathname.current === '/apply';
+      const nowIsNavLink = navLinks.some((link) => link.href === pathname);
+
+      if (cameFromHomeOrApply && nowIsNavLink) {
```
clementroze, line 57: We want the highlight pill to animate between the Team, Products, Course, Initiatives, and Sponsor links. We don't want a pill behind Home (it's the logo) or Apply (it already has a white pill for the button shape).

So what was happening was that if you were on Products for example, and then went to Home, and then went to Sponsor for example, the pill would animate from the left side. This is because the logic tells the pill to transition smoothly from its previous position. In this case, the previous position was from Products, since the Home page doesn't have a corresponding nav link and thus didn't set a position for the highlight pill.

As a result, the highlight pill would animate in from the incorrect position, which created an unintended slide-in effect.

I fixed this by adding a check to see where the user is coming from. Specifically, if the previous page was either `/` or `/apply`, and the new page is one of the nav links (Team, Products, etc.), we skip the sliding animation entirely. Instead, we directly set the highlight pill to its new position with no transition.

This way, we avoid animating from an irrelevant position when the highlight wasn’t visible on the previous page. The logic also ensures that if you're moving between valid nav links, the highlight pill animates smoothly as expected.

new-dti-website-redesign/src/components/Navbar.tsx
```diff
@@ -38,26 +102,49 @@ export default function Navbar() {
...
+                className="bottom-0 absolute -z-10 rounded-full h-10 bg-background-2"
+                ref={highlightRef}
+                style={{
+                  left: highlightStyle.left,
+                  width: highlightStyle.width,
+                  opacity: isNavLink ? 1 : 0,
+                  pointerEvents: isNavLink ? 'auto' : 'none',
+                  transition:
+                    'left 0.2s cubic-bezier(.4,0,.2,1), width 0.2s cubic-bezier(.4,0,.2,1), opacity 0.3s ease'
+                }}
```
clementroze, line 135: These CSS props are in the `style` attribute instead of `className` since we're dynamically updating the styles, so they have to be applied inline at runtime. 

Also the transition is also in style bc it's too messy to have custom transition states for multiple properties in Tailwind lol

clementroze (commented):
Just leaving some additional context in these review comments

new-dti-website-redesign/src/components/Navbar.tsx
jujulcrane: Woah I didn't know useLayoutEffect was a thing 😮

jujulcrane (approved):
Animation looking fireeeee. Seems to be working as intended on my end :)

new-dti-website-redesign/src/components/Navbar.tsx
clementroze: lol yeah i also just discovered it

clementroze:
awesome thanks juju!

PR #972, opened 2025-05-22
[New Website] Mobile nav & banner polish, added "Skip to main" link

clementroze:
# Changes

- Made `<Banner>` flushed top to the navbar
- Fixed bug where you could scroll on rest of page when mobile navbar was open
- Added `id`s to `<main>`
- added "Skip to main content" links

# Video

https://github.com/user-attachments/assets/9a52fa39-2e42-4320-8920-1019bfac71a5

# Test plan

### Skip to main content link

It's a little technical but here are some instructions

- Open `localhost:3000`
- Press `tab`
  - Ensure that the "Skip to main content" link is visible
- Press `tab` a few times 
  - Ensure that you're able to navigate around the links **in the navbar** (eg: logo, Team, Products, Course, etc.)
- Reload the page and press `tab` once again
  - Now click `enter`
  - Ensure that when you press `tab` again, you skip over all the links in the navbar. 
  - The first focused element should be the white CTA button in the hero section (eg: "Apply to DTI")
- Then, using `tab`, navigate to some other page using any of the links 
- Press `tab` again, and ensure that "Skip to main content" is still the first element that you focus on

### Banner

- To test the banner, open the website and resize the window
- Head to `/apply` 
- Make sure there isn't an awkward gap between the banner and the navbar (like [this](https://github.com/cornell-dti/idol/pull/972#discussion_r2101676524) – it should be completely flushed)

### DIFF
--- a/new-dti-website-redesign/src/app/layout.tsx
+++ b/new-dti-website-redesign/src/app/layout.tsx
@@ -35,7 +35,7 @@ export default function RootLayout({
           href="#main-content"
           className={`${baseStyles} absolute -inset-4 opacity-0 z-70
           focus-visible:opacity-100 ${
-            skipToMainOffset ? 'focus-visible:top-16' : 'focus-visible:top-4'
+            skipToMainOffset ? 'focus-visible:top-20' : 'focus-visible:top-4'
           } focus-visible:left-4 transition-all duration-[120ms] px-6 h-12 bg-foreground-1 text-background-1 hover:bg-foreground-2 w-fit`}
         >
           Skip to main content
--- a/new-dti-website-redesign/src/app/layout.tsx
+++ b/new-dti-website-redesign/src/app/layout.tsx
@@ -28,7 +28,7 @@ export default function RootLayout({
         <a
           id="skip-to-main"
           href="#main-content"
-          className={`${baseStyles} absolute -inset-4 opacity-0 z-50 p-2
+          className={`${baseStyles} absolute -inset-4 opacity-0 z-50
           focus-visible:opacity-100 focus-visible:top-4 focus-visible:left-4 transition-all duration-[120ms] px-6 h-12 bg-foreground-1 text-background-1 hover:bg-foreground-2 w-fit`}
         >
           Skip to main content
--- a/new-dti-website-redesign/src/app/layout.tsx
+++ b/new-dti-website-redesign/src/app/layout.tsx
@@ -12,7 +12,12 @@ export default function RootLayout({
   children: React.ReactNode;
 }>) {
   const pathname = usePathname();
+
+  // on design system pages, we don't want to show the navbar
   const shouldShowNavbar = !pathname.startsWith('/design-system');
+
+  // on apply page, we want the "Skip to main content" link to be more offset (because of the banner)
+  const skipToMainOffset = pathname.startsWith('/apply');
 
   // Focus <body> on route change to reset tab order
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


new-dti-website-redesign/src/app/layout.tsx
```diff
@@ -12,9 +14,26 @@ export default function RootLayout({
...
   const shouldShowNavbar = !pathname.startsWith('/design-system');
 
+  // Focus <body> on route change to reset tab order
+  useEffect(() => {
+    const body = document.body;
+    if (body) {
+      body.focus();
+    }
+  }, [pathname]);
+
```
clementroze, line 24: Whenever we first load a page, this link should be the first thing users focus on when pressing tab (see [this](https://www.a11y-collective.com/blog/skip-to-main-content/)).

But we're resetting the focus on `<body>`, and adding `tabIndex={-1}` so that the body receives focus but doesn't actually clutter the tab order.

new-dti-website-redesign/src/app/layout.tsx
```diff
@@ -12,9 +14,26 @@ export default function RootLayout({
...
   return (
     <html lang="en">
-      <body>
+      <body className="relative" tabIndex={-1}>
+        <a
+          id="skip-to-main"
+          href="#main-content"
```
clementroze, line 30: We're not using `<Button href>` here because that would render a `<Link>`. However, `<Link>` isn't a real DOM element (ie: doesn't appear in the HTML tree when using Inspect). It's a wrapper that that renders `<a>` inside. So since the `href` is pointing towards an id (`#main-content`), then we have to use `<a>`

new-dti-website-redesign/src/app/layout.tsx
```diff
@@ -12,9 +14,26 @@ export default function RootLayout({
...
+          className={`${baseStyles} absolute -inset-4 opacity-0 z-50 p-2
```
clementroze: We're using `baseStyles` imported from `<Button>` to avoid having to repeat the same styles again (but we still need a few specific ones like the padding or the background colors, since those are set in the variants outside of `baseStyles`.

new-dti-website-redesign/src/components/Banner.tsx
```diff
@@ -5,7 +5,7 @@ type Props = {
 };
 
 const Banner = ({ label }: Props): ReactNode => (
-  <div className="flex items-center justify-center sticky top-20 z-20 min-[1200px]:pt-4 min-[1200px]:mt-[-67px]">
+  <div className="flex items-center justify-center sticky top-[72px] z-20 min-[1200px]:pt-4 min-[1200px]:mt-[-67px] w-fit mx-auto">
```
clementroze, line 8: Navbar has slightly changed height so we need to change the offset to avoid this narrow gap between the banner and the nav:


Screenshot 2025-05-22 at 1 39 43 AM

clementroze: `w-fit` is also to ensure that the wrapper of the banner doesn't have a huge width that overlaps other content behind it.

jujulcrane (commented):
Hi Clément this looks great!! Your attention to detail is always so impressive. The tabbing worked as you described in the test plan, although I did notice that the lettering of "Cornell Digital Tech & Innovation Project Team" overlapped with the Skip To Content button when the screen width was smaller. It's not super noticeable but I think it should be a pretty easy fix. 
Screenshot 2025-05-24 at 4 37 18 PM
Regarding the banner, there were no gaps however I'm wondering if you intended for the banner to no longer have rounded edges even before it takes up the full width of the screen. Here are screenshots of what I mean: 
Screenshot 2025-05-24 at 4 39 06 PM
When the screen is a bit wider: 
Screenshot 2025-05-24 at 4 39 15 PM

Other than those two tiny things - awesome job!!

clementroze:
Hey @jujulcrane!

thank you for confirming the test plan!

for the overlap you mentioned, I don’t think thats necessarily an issue — I had intentionally made it `position: absolute` and positioned it at the top left. This makes it so that it’s the first thing people see immediately when pressing `tab`. [Here](https://jaketrent.com/post/accessibility-skip-to-main-content/) are some examples of “skip to main content” links for reference.

also good catch on the banner border radius! Am about to hop on my flight for the Motherland 🥖 but will address when I arrive! Ty! :)

clementroze:
okayy @jujulcrane i fixed the bugs:

### Banner border radius:
https://github.com/user-attachments/assets/05e875bf-79ca-4591-a301-ffa85c58d288


### "Skip to main content" link overlap + placement on "Apply" page:

|Before|After|
|-|-|
|Screenshot 2025-05-25 at 4 45 11 PM|Screenshot 2025-05-25 at 4 45 31 PM|

jujulcrane (approved):
yay looks great!

PR #973, opened 2025-05-24
[New Website] Update `<Accordion>` component, add to "Apply" and Design system library pages

clementroze:
# Changes

- Changes to `<Accordion>` component:
  - Made a custom component instead of using `<details>` and `<summary>`
  - Animated opening/closing of accordion
  - Minor stylistic changes including color, padding, border
  - Added prop for icon

- "Apply" page:
  - Added FAQ section, comprised of 3 tabs and their respective accordions for Q&A
  - Added Q&A content inside a `.json`
  - Used `react-markdown` package so that the page could parse newlines + links from the `.json`

- Design system library:
  - Added `/design-system/accordion` page with examples

# Video

(if the videos don't play try downloading them)

### New component changes:
https://github.com/user-attachments/assets/87b33a6c-697d-4763-aaae-2625ac226391

### Responsiveness
https://github.com/user-attachments/assets/ed7ae4e1-cb70-403b-8fcf-240c2b3b27e4

# Test plan
- Head to /apply
  - Basic functionality/visuals:
    - Make sure FAQ section is there
    - Make sure switching tabs shows different content
    - Make sure accordions work as expected
    - Make sure the "How do we evaluate applications?" question's content has a newline
    - Make sure the "What could I do to learn more about DTI?" question's content has a clickable link
    - Make sure the accordions in "Technical Prep" tab have icons
  - Accessibility:
    - Using `tab` on your keyboard, make sure arrow keys navigate you between the tabs
    - Using `tab`, make sure `enter` or `space` correctly triggers the accordion open/close
    - Make sure focus state outlines appear
  - Responsiveness:
    - Make window width smaller and make sure it flows correctly
- Head to `/design-system/accordion`
  - Make sure two examples (with and w/o icon) are visible


new-dti-website-redesign/package.json
```diff
@@ -14,7 +14,8 @@
     "next": "15.2.3",
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
-    "react-fast-marquee": "^1.6.5"
+    "react-fast-marquee": "^1.6.5",
+    "react-markdown": "^10.1.0"
```
clementroze, line 18: This is a package that parses markdown and renders React elements

See [this](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#what-is-this)

new-dti-website-redesign/src/app/apply/page.tsx
```diff
@@ -4,14 +4,50 @@ import Banner from '../../components/Banner';
...
+              p: ({ children }) => <p>{children}</p>,
+              a: ({ href = '#', children }) => (
+                <Link
+                  href={href}
+                  target="_blank"
+                  className="text-foreground-1 underline underline-offset-3"
+                >
+                  {children}
+                </Link>
+              )
```
clementroze, line 40: Different paragraphs should be rendered when source data is separated by `\n\n`

Links should render with `[text](url)`

new-dti-website-redesign/src/app/apply/page.tsx
```diff
@@ -54,8 +90,36 @@ export default function Apply() {
...
+              },
+              {
+                label: 'Behavioral Prep',
+                content: (
+                  <ul className="flex flex-col gap-4 p-8 pl-12 border-t-1 border-border-1 list-disc">
+                    {faqs.BehavioralPrep.map((item, idx) => (
+                      <li key={idx}>{item}</li>
+                    ))}
+                  </ul>
+                )
```
clementroze, line 114: Since the content in the "Behavioral Prep" tab are just bullet points, we're rendering them directly from `faqs.json` without passing through `renderFaqContent`

new-dti-website-redesign/src/app/test-components/page.tsx
```diff
@@ -12,7 +12,7 @@ import IconWrapper from '../../components/IconWrapper';
 import FeatureCard from '../../components/FeatureCard';
 import Chip from '../../components/Chip';
 import Tabs from '../../components/Tabs';
-import FAQAccordion from '../../components/FAQAccordion';
+import Accordion from '../../components/Accordion';
```
clementroze, line 15: FYI we're soon going to deprecate this page in favor of the Design System Library.

new-dti-website-redesign/src/components/Accordion.tsx
```diff
@@ -0,0 +1,81 @@
+'use client';
```
clementroze, line 1: FYI, some parts of this file follows the original [`FaqAccordion.tsx`](https://github.com/cornell-dti/idol/blob/main/new-dti-website-redesign/src/components/FAQAccordion.tsx) file made by Adrienne, but I'm renaming it to `<Accordion>` for more general usecases.

new-dti-website-redesign/src/components/Accordion.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }, [isOpen]);
+
+  // generate a unique ID for the accordion
+  const id = useId();
+  const contentId = `accordion-content-${id}`;
+
+  return (
+    <div className="border-t-1 border-border-1">
+      <h3 className="h6">
+        <button
```
clementroze, line 30: One of the biggest changes is that we've moved away from the native HTML `<details>` and `<summary>` elements. While these elems provide build-in accessibility and toggle behavior, we're not able to extensively style or animate them which is a bummer.

Instead, this is a custom React component that uses explicit open/close state with `useState`. It provides smooth height transitions for expanding/collapsing content. And most importantly, it implements full accessibility support with ARIA attributes. See more about these with my individual comments.

new-dti-website-redesign/src/components/Accordion.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  useEffect(() => {
+    if (contentRef.current) {
+      setHeight(contentRef.current.scrollHeight);
+    }
+  }, [isOpen]);
+
+  // generate a unique ID for the accordion
+  const id = useId();
+  const contentId = `accordion-content-${id}`;
```
clementroze, line 25: We're using React's `useId` to generate unique IDs for each accordion instance. This ensures that when we have several instances on our page, we avoid ID duplicates which could mess up accessibility.

new-dti-website-redesign/src/components/Accordion.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <h3 className="h6">
+        <button
+          id={`${contentId}-header`} // unique ID for the header, referenced by aria-labelledby in <div>
+          aria-controls={contentId} // points to ID of controlled content region (the sibling <div>)
+          aria-expanded={isOpen}
+          onClick={() => setIsOpen(!isOpen)}
+          className="w-full p-4 sm:p-8 flex justify-between items-center cursor-pointer bg-background-1 hover:bg-background-2 transition-colors duration-[120ms] focusState text-left"
+          style={{
+            background: isOpen ? `var(--background-2)` : ''
+          }}
```
clementroze, line 38: This is in the `style` attribute because `background` is being set based on the `isOpen` state.

new-dti-website-redesign/src/components/Accordion.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </svg>
+        </button>
+      </h3>
+
+      <div
+        aria-labelledby={`${contentId}-header`} // uses <button>'s ID to label this region
+        id={contentId} // unique ID for content panel, referenced by aria-controls in <button>
+        ref={contentRef}
+        role="region"
+        aria-hidden={!isOpen}
```
clementroze, line 71: Since we're only animating the height of the FAQ contents to visually show the expanded/collapsed state, the content is still actually rendered. It's just invisible. However, screen readers will still pick up on it.

That's why when the accordion is supposed to be collapsed, we ˙ave `aria-hidden` so the screen reader doesn't pick up on it.

new-dti-website-redesign/src/components/Accordion.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const id = useId();
+  const contentId = `accordion-content-${id}`;
+
+  return (
+    <div className="border-t-1 border-border-1">
+      <h3 className="h6">
+        <button
+          id={`${contentId}-header`} // unique ID for the header, referenced by aria-labelledby in <div>
+          aria-controls={contentId} // points to ID of controlled content region (the sibling <div>)
+          aria-expanded={isOpen}
```
clementroze, line 33: This is a bit tricky, but essentially, we're linking the accordion header (the `<button>`) and its content panel (the following `<div>`) via ARIA attributes for accessibility reasons.

The `<button>` has an ID and uses `aria-control` to reference the content panel's ID

The content panel itself uses `aria-labelledby` to point back to the `<button>`'s ID.

This thus creates an explicit relationship between the control (the button trigger) and the region (the hidden/shown content) it toggles.

new-dti-website-redesign/src/components/FAQAccordion.tsx
clementroze, line 1: (It's not actually gone lol, just changed the name)

new-dti-website-redesign/src/components/Tabs.tsx
```diff
@@ -12,9 +12,10 @@ type TabsProps = {
   tabs: Tab[];
   className?: string;
   center?: boolean;
+  tabsContainerPadding?: boolean;
 };
 
-export default function Tabs({ tabs, className = '', center }: TabsProps) {
+export default function Tabs({ tabs, className = '', center, tabsContainerPadding }: TabsProps) {
```
clementroze, line 18: Just an extra prop to help with some styling.

new-dti-website-redesign/src/components/Accordion.tsx
clementroze: I modeled this component based off of [this](https://www.aditus.io/patterns/accordion/#react) btw

jujulcrane (commented):
This looks great! I'm wondering if the tabbing is working as you intended? I'm not super familiar with how it's supposed to work but I assumed that it would visit all of the buttons, including behavioral prep and technical prep which it currently skips over. Let me know if I'm mistaken! 

https://github.com/user-attachments/assets/a1ebd66d-99fa-461c-9252-7bbb1c4263b3

Also not sure if this is in the scope of this pr but the indentations for the roles in role descriptions are a bit inconsistent which looks a bit off to me: 

https://github.com/user-attachments/assets/6bf48472-03a2-4515-8be8-f9bb050c80aa

clementroze:
> This looks great! I'm wondering if the tabbing is working as you intended? I'm not super familiar with how it's supposed to work but I assumed that it would visit all of the buttons, including behavioral prep and technical prep which it currently skips over. Let me know if I'm mistaken!

Hi Juju, ah yes thanks for bringing this up, this is a totally valid point! So actually according to [documentation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/), `tab` is used to jump from interactive element to interactive element on the whole page. However, once the tabs are focused upon, it's the `left` and `right` arrows which should be used to move to the next/previous list. 

Feel free to check it out on localhost just to verify it works! :)



> Also not sure if this is in the scope of this pr but the indentations for the roles in role descriptions are a bit inconsistent which looks a bit off to me:


Oooh yes i agree, that behavior is kinda inconsistent. Just pushed a fix!

jujulcrane (approved):
Looks great, thanks for fixing the indentation!

clementroze:
> Looks great, thanks for fixing the indentation!

Of course, and thank you for bringing it up 😄

PR #980, opened 2025-05-30
[New Website] Design System Library Clean up

clementroze:
# Changes

- Deleted `/test-components` and `/test-page` pages
- Added page sections (Hero, Feature, CTA, Footer, Title, Separator) in Design System Library)
- Cleaned up folder organization by moving common components used in the Library in a `utils` folder
- Added Chris' TeamCard and TimelineCard components to the Library

# Test plan
- Make sure the yellow DevNavbar only links to the Design System Library
- Make sure there's a new category in the `/design-system` called "Page sections" with all these new blocks

# Note

Much of the content (text, card thumbnails, placement of elements) is still temporary. When we're closer to launching the whole website redesign, I'll make sure all the content is up to date :)


clementroze (commented):
OKayyy so it looks like there's lots of changes on the PR, but don't worry most of is just moving things around (eg: adding new folders/directories, or migrating components into the libary)

Lmk if you have any questions!

cchrischen (approved):
I think we can move the link to the design system into the footer at this point. Thoughts?

clementroze:
> I think we can move the link to the design system into the footer at this point. Thoughts?
> 
> Could you also make the aside `overscroll-auto`?


Good point, i removed the banner since it's in the footer here:

Screenshot 2025-05-31 at 8 10 33 PM


And changed the sidebar to `overflow-auto`

PR #981, opened 2025-05-30
[New Website] Update `<Tabs>` and `<FancyTabs>`

clementroze:
### Summary 

This pull request is the first step towards implementing feature Foo

- [x] implemented X
- [ ] fixed Y

- [ ] Updated backend api specification

### Breaking Changes 


new-dti-website-redesign/public/apply/roles/design.svg
clementroze, line 1: I renamed these SVGs from `designer`/`developer` to `design`/`development` for consistency

jujulcrane (approved):
The icons look so good!!! One little thing that maybe doesn't matter or is inevitable is that the background selection of the pills is wider than the key selection outer ring (although if someone is looking at the website with such a small viewport they probably won't be using keys...) 

https://github.com/user-attachments/assets/a25f4dbc-caf0-464f-a128-e4aba6f8b01d

You can totally ignore this but I just thought it wouldn't hurt to bring it up incase you weren't aware.

clementroze:
> The icons look so good!!! One little thing that maybe doesn't matter or is inevitable is that the background selection of the pills is wider than the key selection outer ring (although if someone is looking at the website with such a small viewport they probably won't be using keys...)
> 
> You can totally ignore this but I just thought it wouldn't hurt to bring it up incase you weren't aware.

Thanks @jujulcrane!! 

Ah great catch, wasn't aware of this bug! Yeah it's true that at this viewport, it's probably safe to assume users are probably on a touchscreen mobile device and won't be using keyboards. But I fixed it anyways, it was an ez fix 😎 . Plus quality + care are important!


https://github.com/user-attachments/assets/fd578e3c-5efa-4d86-bc2f-9414c1785ce1

PR #985, opened 2025-06-12
404 page

clementroze:
wip


cchrischen (approved):
Looks good! Any reason it's in a `.js` rather than `.tsx`?

clementroze:
> Looks good! Any reason it's in a `.js` rather than `.tsx`?

Oh whoops, i had taken it from [here](https://nextjs.org/docs/app/api-reference/file-conventions/not-found) but seems like `.tsx` works too. thank you!

PR #997, opened 2025-08-01
[New Website] SectionSep component fix

ladriennel:
### Summary 

- Minor fixes to SectionSep component
  - apply the default sectionSep class unless disableSectionSepStyle is passed as a prop (used to remove margins and width constraints when needed)


clementroze (approved):
Thank you Adrienne!

PR #1011, opened 2025-08-13
[New Website] Screenreader audit

clementroze:
# Changes

- Notion [ticket](https://www.notion.so/Perform-screenreader-audit-24c0ad723ce180839543e14f2cfcb6d1?source=copy_link)

We should make an effort to ensure that the website is as accessible by anyone, regardless of their way of using the website.  This includes individuals who navigate the website using only their keyboard (and by using Screen Reader technology)

I audited the website using only a screen reader to check for things like image alt text, correct button labels, semantic navigation, intuitive labels, etc.

Made a few fixes to make things as clear as possible to someone who might be navigating the website blind for example, or who might not be able to use their mouse.

# Test plan
- Test out website using screen reader technology


clementroze:
Ok thanks for reviewing! Back on campus I'll show you how the screen reader navigation works :)

PR #1022, opened 2025-08-21
specify pm apps only open to upperclassmen

cchrischen:
### Summary 
PMs are only recruiting upperclassmen for the upperclassmen cycle.


clementroze (approved):
Thanks Chris!

clementroze (approved):
Thanks Chris!

PR #1035, opened 2025-09-02
add accepted applicant summary to app dashboard

ladriennel:
### Summary 
- Added a summary row to bottom of dashboard table listing how many applicants are accepted for each round

### Notion

[Edit application dashboard](https://www.notion.so/FA25-Roadmap-2410ad723ce18141927bea8c5860f4ab?p=2410ad723ce181ef8c39f9602396c0c8&pm=s)

### Test Plan 
- Go to Interview Status Dashboard
- Add applicants with different statuses
- Check that the summary row at the bottom of table updates accordingly
- Filter by different rounds/applicants and check updated counts

Screenshot 2025-09-02 at 5 28 25 PM


shah-esha (approved):
Overall, everything looks good and is working smoothly! One thing I noticed for future improvement: the accepted count currently updates when we filter by status (accepted, rejected, etc). This means that if we apply a filter other than “accepted,” the summary shows 0 accepted, even if there are accepted applicants. This could be confusing, so a good next step would be to separate the filters: one for status and another for role.

PR #1042, opened 2025-09-07
add interview scheduler note

cchrischen:
### Summary 
Add a note to remind applicants they cannot cancel interview slots after they sign up.


amelieRosso (commented):
looks good!

PR #1052, opened 2025-09-17
Update TEC requirements

ladriennel:
### Summary 

Updated TEC requirements to be monthly periods with no carryover from previous periods. Members must submit 1 per period and leads 2 per period. 

- Changed TEC deadlines
- Fixed TEC form and TEC dashboard to reflect new requirements
- Adjusted emails so that leads only send emails to remind members who need to complete tec for the current period, no semester reminders

### Linear
[tec updates](https://linear.app/cornelldti/issue/IDOL-21/tec-updates)

### Test Plan
- In TEC dashboard normal view, members only green if they are up to date on tec for all periods so far
- In TEC dashboard period view, members only green if they are up to date on current period's tec requirement
- Notify button sends emails to all members who have missing credits for current period (in logs in non-prod)

### DIFF
--- a/backend/src/API/mailAPI.ts
+++ b/backend/src/API/mailAPI.ts
@@ -8,7 +8,12 @@ import { BadRequestError, PermissionError } from '../utils/errors';
 import { env } from '../firebase';
 import TeamEventAttendanceDao from '../dao/TeamEventAttendanceDao';
 import TeamEventsDao from '../dao/TeamEventsDao';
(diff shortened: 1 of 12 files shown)


oscarwang20 (changes requested):
Overall looks good, we might want to look more into how to make this more generalizable to future semesters seeing as how these requirements keep on changing, but that's a discussion for another day. Left some comments, lmk if you have any questions.

oscarwang20 (approved):

PR #1062, opened 2025-09-29
[Website] Added fallback images to Team page

clementroze:
Added fallback images so it doesn't just show the alt text on the images

There are 4 fallback image variants. It cycles between the 4
Screenshot 2025-09-29 at 3 56 10 PM

## BEFORE:
image

## AFTER:
image

image

image

test plan: open /team sure each newbie's image doesn't show the alt text but instead shows a different variant of the bear


cchrischen (approved):
cute!

PR #1068, opened 2025-10-05
Updated coffee chat suggestions script

ladriennel:
### Summary 

- Updated semester from Spring 2025 to Fall 2025
- Removed dev advisor checks since that is no longer a coffee chat category
- Ran script to update new coffee chat suggestions

### Linear
[coffee chat suggestions](https://linear.app/cornelldti/issue/IDOL-38/coffee-chat-suggestions)

### Test plan
- Go to coffee chat bingo board
- Check that each category has populated suggestions, no duplicate names (newbies should be easy to verify for accuracy of suggestions)

### Note 
I wanted to make a pr to merge the script since we would need to use it every semester (originally from Chris's branch).

Right now, we also need to manually check for members who make multiple submissions to the google form or submit their name in lowercase (i.e. validate all the data). Do we want to handle those cases in the script?, or since it's not too much work it's fine?


backend/scripts/populate-coffee-chat-suggestions.ts
```diff
@@ -0,0 +1,101 @@
...
+    (mem) => mem.semesterJoined === 'Fall 2025'
+  );
+
+  return suggestions;
+};
+
+const main = async () => {
+  const membersByCategory = await getMembersByCategory();
+
+  const members = await memberPromise;
```
shah-esha, line 77: This looks really great Adrienne! I did notice that you fetch `members` both here and in the `getMembersByCategory` function (line 45). I would suggest passing `members` as a parameter instead to avoid redundant fetching. Otherwise, great job!

PR #1079, opened 2025-10-21
Allow leads to take the member slot in interview scheduler

qiandrewj:
ily thanks


cchrischen:
Can you highlight a slot green if you are either in the lead or members slot? Currently does not highlight green for when you are signed up as a member slot as a lead.
Screenshot 2025-10-21 at 4 36 26 PM

PR #1138, opened 2026-02-28
Interview Scheduler new ics email invites

ladriennel:
### Summary 

The previous gapi/Google OAuth approach required the https://www.googleapis.com/auth/calendar.events scope to  solve an issue we came across where applicants were receiving errors from trying to add their time slot to gcal. The issue was that this scope is classified as sensitive by Google and when added, it requires an app verification before it can be used in production which is more overhead than needed for this feature.

This new approach uses .ics attachments sent via email that don't require OAuth and can work with any calendar app. When sign-ups and cancellations occur, an email is sent with the ics attachment that includes the time slots location, date/time, and a uid. This uid is the same across the invitation + cancellation, so the same event is modified each time. Previously calendar invites were only sent to applicants, now DTI members also receive them.

- Removed `calendar.ts` and any previous, related gapi integration
- Applicants and DTI members who sign up to interview receive confirmation emails with an ics calendar invite after signing up in an `InterviewScheduler` instance.   
- Applicants receive cancellation emails if time slot is deleted or admin removes them, DTI members receive this email for same reason as well as if they decide to cancel their own sign up
- Leads currently receive no notifications 
- Applicant confirmation card now includes text saying that they received an email rather than a button with option to add to gcal

### Test Plan 
- Create a test Interview Scheduler instance with an applicant of your personal (non Cornell) email (so that you can mimic an applicant). Add time slots (need to temporarily change role to lead).
- Through personal email, access IDOL local host and sign up for a time slot. Check confirmation card that notes a success and email has been sent. In non-prod, the email won't actually be sent, but an stdout will appear in the backend terminal with the contents
- Check that the email contents and ics content are correct and match sign up details. 
- Through Cornell email, sign up for a time slot as an IDOL member. Check email contents match. 
- Check that removing an applicant from time slot sends cancellation to applicant, deleting entire time slot sends cancellation to applicant and member, and cancelling sign up as member sends email.

### Notion Link 
[notion](https://www.notion.so/Idol-SP26-2ea0ad723ce18080ae5ffbae682175ad?p=2f70ad723ce180399dc0c7519db03893&pm=s)

### Notes
- To test the actual email and ics calendar functionality, I ran a temporary script with the exact ics details + email content from stdout testing to confirm it works.

Screenshot 2026-02-28 at 5 03 51 PM


backend/src/API/interviewSchedulerAPI.ts
```diff
@@ -4,7 +4,12 @@ import InterviewSlotDao from '../dao/InterviewSlotDao';
 import { BadRequestError, NotFoundError, PermissionError } from '../utils/errors';
```
jane-leon: Hi Adrienne. Everything looks really well done, and when tested, I did receive the email and got it added to my GCal. There was only really one comment really on what to change. However, I do wonder if you have tested out deleting an available time slot from like leads side. I know this isn't common so it's not that big of a worry.

jane-leon (approved):

PR #1142, opened 2026-03-17
Reimbursement Typescript + Firestore/Daos

jane-leon:
### Summary 

This pull request adds TypeScript type definitions for the Budget Reimbursement Tool feature.
  - Added reimbursement types to common-types/index.d.ts (User, Team, Request, Category, etc.)

Update : Pull Request also adds Firestore collections, DAOs, and tests, completing database layer

### Notion/Figma Link 

More details on how the design doc for the reimbursement tool can be found here (it has been updated, look at tab REDONE): https://docs.google.com/document/d/1JcQlSsk6EcJHifOd76LpUQO-VddWYEyaEoVTcVbUoN4/edit?usp=sharing 

### Test Plan 
Types are defined and compiles without errors.

Update: For DAOs
  - All TypeScript types compile without errors
  - New DAO tests pass (run npm test from root)
  - Tests cover all CRUD operations and query methods
  - DAOs integrate with Firestore collections

### Breaking Changes 

 Database schema change (adds new Firestore collections: users, teams, reimbursement_requests, categories)


backend/src/dao/ReimbursementRequestDao.ts
```diff
@@ -0,0 +1,160 @@
...
+  async createRequest(request: ReimbursementRequest): Promise<ReimbursementRequest> {
+    return this.createDocument(request.requestId, request);
+  }
+
+  /**
+   * Updates a reimbursement request
+   * @param request - The updated request data
+   * @returns The updated `ReimbursementRequest`
+   */
+  async updateRequest(request: ReimbursementRequest): Promise<ReimbursementRequest> {
```
ladriennel, line 92: I think we should check `isImmutable` here and throw if true.

common-types/index.d.ts
```diff
@@ -605,3 +605,74 @@ interface Alumni {
...
+type ReimbursementRequestStatus =
+  | 'pending'
+  | 'needs_changes'
+  | 'approved'
+  | 'in_progress_with_cornell'
+  | 'settled';
+
+/** Reimbursement user */
+interface ReimbursementUser {
+  userId: string;
```
ladriennel: I think it might be worth dropping `ReimbursementUser` and matching requester + admin behavior based on role + and netid as `requesterID`. This would remove the constraint of having to manually edit the user every semester as people change role/subteam a lot: biz leads can have admin perms, any other lead and ops can have requester perms (because to my understanding only biz leads manage reimbursements). Phone number and address can live in the request type probably. lmk how you feel!

common-types/index.d.ts
```diff
@@ -605,3 +605,74 @@ interface Alumni {
...
+  requesterId: string;
+  teamId: string;
+  amount: number;
+  reason: string;
+  attendees: string[];
+  dateOfPurchase: number;
+  dateSubmitted: number;
+  status: ReimbursementRequestStatus;
+  receiptUrl: string;
+  adminNote: string | null;
```
ladriennel: What is the purpose of `adminNote`, `notes`, and `messages`? I think for the necessary workflow, just keeping `messages` should be enough right?

common-types/index.d.ts
```diff
@@ -605,3 +605,74 @@ interface Alumni {
...
+}
+
+/** Team budget information */
+interface ReimbursementTeam {
+  teamId: string;
+  teamName: string;
+  budget: number;
+  totalSpent: number;
+  assignedAdmins: string[];
+  createdAt: number;
```
ladriennel: Is there a specific purpose for the `createdAt` field for any of the tables/reimbursement tracking views? If not, I think that keeping track of it isn't necessary.

ladriennel (changes requested):
Hi Jane ! great work designing the reimbursement tracker and making tests as well! 

I think that besides the comments I left below, we could also add a method for resetting the a teams budget/state every semester.

backend/src/utils/reimbursementPermissions.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export function hasReimbursementAdminPermissions(role: Role): boolean {
+  return role === 'business-lead';
+}
+
+/**
+ * Determines if a user has requester permissions for reimbursements.
+ * @param role - The user's Role
+ * @returns true if the user has requester permissions
+ */
```
ladriennel, line 22: pm's can also request, so they should be added here. nit: it might also be worth moving these checks to `permissionsManager.ts` to be consistent with previous perm patterns but not completely necessary

backend/src/dao/ReimbursementTeamDao.ts
```diff
@@ -0,0 +1,116 @@
...
+   * Gets a reimbursement team by its teamId
+   * @param teamId - The team ID
+   * @returns The `ReimbursementTeam` with a matching teamId (null if it doesn't exist)
+   */
+  async getTeam(teamId: string): Promise<ReimbursementTeam | null> {
+    return this.getDocument(teamId);
+  }
+
+  /**
+   * Gets all teams assigned to a specific admin
```
ladriennel, line 31: Just curious, what is the use case for this?

ladriennel (approved):

PR #1143, opened 2026-03-19
Fixed alumni Card responsivity for tablet/mobile

valerie-wong:
In this PR, I adjusted the styling of the Alumni Cards to be more responsive to different screen sizes and handle long text better.

BEFORE:
Right now you can see that depending on how long some of the words are, the spacing of the fields is uneven.
    
    
    
- As is, it’s not a huge issue, but when the screen changes sizes then it starts getting weird:
    
    
    
My PR changes the layout of the cards different by screen siz

- Large/laptop: keep as image on left then 2x3 layout of fields
- Medium/tablet: keep as image on left then 3x2 layout of fields
- Small/mobile: image on top and then 3x2 layout of fields below

To test this, I would try messing with the screen widths and seeing how they funtion. I would run through searching through alums and filters to see if tehre are also any other edge cases that break. One thing to note, is that on tablet, the filter bar on the left takes up a lot of space. I believe this is out of scope for this specific task as it lies within the Map layout as opposed to just the Alumni Card Component.


ladriennel (changes requested):
Hi val, the ui changes look good! I only have 1 comment about a slight formatting thing with the grids.

frontend/src/components/Alumni/Alumni.module.css
```diff
@@ -284,3 +280,46 @@
...
+  }
+
+  .infoColumn {
+    flex: unset;
+  }
+
+  .socialIcons {
+    align-self: flex-end;
+  }
+}
```
jane-leon, line 382: Hii Val, everything looks great, the responsivity looks so much better. I think my only comment would be to   Increase spacing between cards, like changing gap: 8px to gap: 12px as currently they're quit close to each other, almost touching.

frontend/src/components/Alumni/Alumni.module.css
```diff
@@ -284,3 +278,112 @@
...
+    height: 80px !important;
+  }
+
+  .cardContent {
+    width: 100%;
+  }
+
+  .infoGrid {
+    display: grid;
+    grid-template-columns: 1 1;
```
ladriennel: should be `1fr 1fr` right?

frontend/src/components/Alumni/Alumni.tsx
```diff
@@ -30,13 +26,20 @@ const Alumni: React.FC = () => {
     DEFAULT_MIN_YEAR,
     DEFAULT_MAX_YEAR
   ]);
+  const [filtersOpen, setFiltersOpen] = useState(false);
+
```
ladriennel, line 38: I notice that on the 768px breakpoint, when you click to open the filter modal, it correctly renders at the bottom of screen but as you scroll it doesn't stay pinned to the bottom of the screen.

ladriennel (commented):
Great fixes! Just left a few comments. I also noticed that this branch is quite behind main, even from the start, so I would suggest in future you pull from main and then start working, to prevent potential conflicts!

ladriennel (approved):
Looks great!! 

Only one thing that still needs fixing I believe is line 369 in `Alumni.module.css` to change `1 1` to `1fr 1fr`

PR #1144, opened 2026-03-22
Implemented Alumni DB admin dashboard

mshi-27:
### Summary 

This pull request implements a new feature in the admin dashboard, Edit Alumni. Provides a nice UI to add alumni to the Firestore database, supporting location (compatible with the alumni map), image upload, and more. 

- New Edit Alumni page at /admin/alumni-admin-view, linked from the main admin dashboard
- Searchable table of alumni; Add Alumni; per-row Edit / Delete;
- Profile image upload to Storage path alumImages/{uuid} (Cornell NetID required before upload); resolves alumImages/… paths via ImagesAPI for preview.
- Searchable dropdown backed by existing city-coordinate records; “Add …” calls geocoding to create a CityCoordinates row when needed; on save, syncs the alumni id on city-coordinate documents (remove from old pins, add to new) so the alumni map stays consistent.
- Endpoint POST /city-coordinates/geocode supports adding new locations

image
image

https://www.figma.com/design/ApnK2bbEIQqpRAyBAv7LDR/IDOL-Ashley-Working-File?node-id=758-6310&p=f&t=nrH1eQ0BEnbw3dke-0

### Test Plan 
- List/Search: Confirm table loads; search matches name, email, company, location, job fields, grad year, DTI roles, subteams.
- Add: Create an alumni with NetID, required fields, optional image upload; save; row appears.
- Edit: Change fields including location; save; verify Firestore alumni doc and that map-related city-coordinate docs add/remove the alumni id when location changes
- New location: Use “Add …” in the location dropdown; confirm geocoding succeeds or shows “Location not found…” when Nominatim returns nothing.
- Delete: Confirm modal; delete; alumni removed

### Notes 
- not sure how to update openapi.yaml with /alumni and /city-coordinates
-


frontend/src/components/Admin/EditAlumni/EditAlumni.module.css
```diff
@@ -0,0 +1,296 @@
...
+
+.modalContactRow {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  color: rgba(0, 0, 0, 0.65);
+  font-size: 14px;
+}
+
+.rowActionsCell {
```
ladriennel: nit: the `rowActionsCell` class is duplicated, so it should be merged into one

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,253 @@
...
+import { Emitters } from '../../../utils';
+
+export default function EditAlumni(): JSX.Element {
+  const [modalOpen, setModalOpen] = useState(false);
+  const [modalMode, setModalMode] = useState<'add' | 'edit'>('add');
+  const [selectedAlumni, setSelectedAlumni] = useState<Alumni | undefined>(undefined);
+  const [alumni, setAlumni] = useState<readonly Alumni[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [searchQuery, setSearchQuery] = useState('');
+  const [saveLoading, setSaveLoading] = useState(false);
```
ladriennel: nit: `saveLoading` , `nameInputOpen`, `emailInputOpen`,`linkedinInputOpen` can probably be moved as states into the modal since they are only really used by the modal not this parent component. I think it'll be cleaner, but not necessary to make the code work.

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,253 @@
...
+      openDelete(alum);
+    }
+  }
+
+  function openDelete(alum: Alumni): void {
+    setAlumniToDelete(alum);
+    setDeleteModalOpen(true);
+  }
+
+  async function deleteAlumni(alum: Alumni): Promise<void> {
```
ladriennel, line 108: Right now when Alumni are deleted, their document + image are correctly removed, but their id stays stored in their corresponding `CityCoordinates`. We should update this to use `removeAlumniFromLocation`

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,253 @@
...
+            <Table.Row>
+              <Table.Cell colSpan="11" textAlign="center">
+                No alumni found.
+              </Table.Cell>
+            </Table.Row>
+          )}
+          {!isLoading &&
+            filteredAlumni.length > 0 &&
+            filteredAlumni.map((alum) => (
+              <Table.Row key={`${alum.firstName} ${alum.lastName}`}>
```
ladriennel: Even though its likely that alumni first and last names are unique, it's safer to use `alum.uuid` as key.

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,253 @@
...
+                <Table.Cell>{`${alum.firstName} ${alum.lastName}` || '—'}</Table.Cell>
```
ladriennel: Template literals don't return false, so the `—` won't render here, but once we require names + email for adding/editing alumni, this fallback won't be necessary anymore

frontend/src/components/Alumni/AlumniCard.tsx
```diff
@@ -20,6 +21,14 @@ type AlumniCardProps = {
 
 const AlumniCard: React.FC<AlumniCardProps> = ({ alum }) => {
   const [imgSrc, setImgSrc] = useState(alum.imageUrl || getFallbackImage(alum.uuid));
+  const ALUMNI_IMAGE_PATH_PREFIX = 'alumImages/';
```
ladriennel: nit: I suggest moving the constant out of the component to module level

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+
+export type AlumniFormState = {
+  uuid: string;
+  imageUrl: string;
+  firstName: string;
+  lastName: string;
+  name: string;
+  email: string;
+  linkedin: string;
+
```
ladriennel: nit: I notice a lot of random empty lines you can remove

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+}
+
+function fromFormState(form: AlumniFormState): Alumni {
+  return {
+    uuid: form.uuid,
+    imageUrl: form.imageUrl,
+    firstName: form.firstName.trim(),
+    lastName: form.lastName.trim(),
+    email: form.email.trim(),
+    linkedin: form.linkedin.trim() ?? null,
```
ladriennel: `.trim()` doesn't return null, you can probably use`||` instead

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+
+    gradYear: form.gradYear ?? null,
+    location: form.location.trim(),
+    company: form.company.trim(),
+
+    jobRole: form.companyRole.trim(),
```
ladriennel: `jobRole` should default to 'Other' (this is a little more nit, but it's based on convention we already started with)

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+      }
+    }
+  }, [open, mode, initialAlumni]);
+
+  // Revoke object URLs on unmount to avoid memory leaks
+  React.useEffect(
+    () => () => {
+      if (imagePreviewUrl) URL.revokeObjectURL(imagePreviewUrl);
+    },
+    [imagePreviewUrl, allLocations.length, loadingLocations]
```
ladriennel: You should remove `allLocations.length` and `loadingLocations` as deps here

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+                    setForm((f) => ({
+                      ...f,
+                      firstName,
+                      lastName
+                    }));
+                  }}
+                />
+              )}
+              <button
+                onClick={() => setNameInputOpen(!nameInputOpen)}
```
ladriennel: Name, email, and linkedin have identical edit patterns, so it may be worth extracting to reduce duplicated code.

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+                  placeholder="Search or add a location"
+                  options={(() => {
+                    const query = locationQuery.trim().toLowerCase();
+                    const uniqueNames = Array.from(
+                      new Set(allLocations.map((c) => c.locationName))
+                    );
+                    const filtered =
+                      query.length > 0
+                        ? uniqueNames.filter((name) => name.toLowerCase().includes(query))
+                        : [];
```
ladriennel: When the location query is empty, e.g. haven't started typing, I think we can consider showing some number of locations by default, but this is just a design choice

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+
+      <Modal.Actions>
+        <Button basic className={styles.cancelButton} onClick={onClose}>
+          Cancel
+        </Button>
+        <Button
+          primary
+          className={styles.saveButton}
+          loading={saveLoading}
+          disabled={saveLoading || !form.uuid.trim() || !isValidSubteamCommaSpaceList(form.subteam)}
```
ladriennel: We shouldn't allow alumni to be created without a name and email, but right now it only enforces netid. We should also enforce at most 2 `dtiRoles` and `subteams` chosen.

frontend/src/components/Modals/DeleteAlumniConfirmationModal.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { Button, Modal } from 'semantic-ui-react';
+import styles from '../Admin/EditAlumni/EditAlumni.module.css';
+import Alumni from '../Alumni/Alumni';
```
ladriennel: This import of `Alumni` is unnecessary since the type is globally available. 
nit: I think for file organization it would be better to name this file `AlumniDeleteConfirmationModal.tsx`

ladriennel (changes requested):
Hi Max! Lots of great work so far, thanks for all the hard work!! I've written quite a lot (sorry lol), so lmk if you have any questions. Some things are design choices and nits, so you can just consider those. 

Updating `openapi.yaml` is not blocking anything in this pr, so if you want to leave as a follow up later you can, but what you would do is add entries for all the new API routes following the existing patterns in that file: path, method, summary, request body schema (where applicable), and response schema.

I also suggest cleaning up some of the test entries in the Firestore dev db once you're done, so it's not cluttered. But this can be done a little later!

ladriennel:
> Hi Max! Lots of great work so far, thanks for all the hard work!! I've written quite a lot (sorry lol), so lmk if you have any questions. Some things are design choices and nits, so you can just consider those.
> 
> Updating `openapi.yaml` is not blocking anything in this pr, so if you want to leave as a follow up later you can, but what you would do is add entries for all the new API routes following the existing patterns in that file: path, method, summary, request body schema (where applicable), and response schema.
> 
> I also suggest cleaning up some of the test entries in the Firestore dev db once you're done, so it's not cluttered. But this can be done a little later!

I also notice 3 issues that should be fixed but can be implemented later so as to not overload this pr:
1) This issue I believe would be a quicker fix than the following 2, but we can still push it to a different pr, but when creating a new alumni, it's allowing you to create an instance with duplicate ids (overriding existing data). To fix, I think we can throw an error in backend Dao and catch/re-throw it in the frontend `saveAlumni` function in `EditAlumni.tsx` to prevent saving.
2) I noticed that when you are editing locations and you create a new one, it automatically calls `geocodeAndStore` which immediately creates a `CityCoordinates` document even if we don't save that location for the alumni. It's non-breaking since it doesn't add the alumni to it, but it leaves orphaned documents in the db. I think we should consider fixing this, but it would probably require refactoring or adding a new endpoint, so it can be touched on later.
3) Similarly, images are still uploaded to storage on cancel, so the solution here would be to upload on save instead.

backend/src/utils/geocodingService.ts
```diff
@@ -119,35 +136,41 @@ export class GeocodingService {
...
-        longitude: result.longitude,
-        alumniIds: []
-      });
+    if (existingAtCoords) {
+      return existingAtCoords;
     }
 
-    return result;
+    // Create a new CityCoordinates entry, using the canonical display_name as locationName
+    return cityCoordinatesDao.createCityCoordinates({
```
valerie-wong, line 234: i noticed that there is also an if (!existingAtCoords) {
  await cityCoordinatesDao.createCityCoordinates({
    locationName: locationString, ...

if no coordinate exist i believe that this might make duplicated entries

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,722 @@
...
+                  if (newLocation) {
+                    const coords = await CityCoordinatesAPI.geocodeAndStore(newLocation);
+                    await CityCoordinatesAPI.addAlumniToLocation(
+                      coords.latitude,
+                      coords.longitude,
+                      alumniToSave.uuid,
+                      coords.locationName
+                    );
+                  }
+                } catch {
```
valerie-wong: Is this intentionally supposed to be empty?

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
valerie-wong: +1 We have two alums both named Alyssa Zhang ...

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,253 @@
...
+            <Table.HeaderCell>Grad Year</Table.HeaderCell>
+            <Table.HeaderCell>Subteam</Table.HeaderCell>
+            <Table.HeaderCell />
+          </Table.Row>
+        </Table.Header>
+
+        <Table.Body>
+          {isLoading && (
+            <Table.Row>
+              <Table.Cell colSpan="11" textAlign="center">
```
valerie-wong: I thought that this table has 10 columns (9 data + 1 actions), but the loading/empty rows use colSpan="11". Not sure if this should be 10 instead?

valerie-wong (commented):
Great work Max! Took a look and had a few things in addition to what Adrienne left comments for!

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
mshi-27: good catch, thanks!

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,800 @@
...
+                <div className={styles.modalName}>
+                  {form.firstName || form.lastName
+                    ? `${form.firstName ?? ''} ${form.lastName ?? ''}`.trim()
+                    : 'New Alumni'}
+                </div>
+              }
+              edit={
+                <Form.Field
+                  control={Input}
+                  value={`${form.firstName ?? ''} ${form.lastName ?? ''}`}
```
ladriennel: Right now when field is initially empty, it's rendering a space in front of what you start typing if you don't clear it (cause it's hard to notice), so it's messing up the split of first name and last name I think.

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,800 @@
...
+            !form.uuid.trim() ||
+            !form.email.trim() ||
+            !form.firstName.trim() ||
+            !isValidSubteamField(form.subteam)
+          }
+          onClick={async () => {
+            let formToSave = form;
+            const pendingUuid = form.uuid.trim();
+            const oldLocation = initialAlumni?.location ?? null;
+            if (pendingImageBlob) {
```
ladriennel: Deferring the image saving is good here, but I think the try/catch structure is a little off. You can move the image upload block (730–744) inside the setSaveLoading try/catch so image upload errors are caught by the existing catch

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,800 @@
...
+              ref={fileInputRef}
+              type="file"
+              accept="image/*"
+              onChange={handleImageUpload}
+              style={{ display: 'none' }}
+            />
+            {uploadError && <span className={styles.uploadError}>{uploadError}</span>}
+          </div>
+          <div className={styles.modalTopText}>
+            <InlineEditableField
```
ladriennel, line 437: nit: we should mark email + name fields with an asterick to indicate they are required

frontend/src/components/Modals/AlumniAddEditModal.tsx
```diff
@@ -0,0 +1,800 @@
...
+  onToggleOpen,
+  display,
+  edit
+}: InlineEditableFieldProps): JSX.Element {
+  return (
+    <div className={rowClassName}>
+      {leading}
+      {isOpen ? edit : display}
+      <button type="button" onClick={onToggleOpen} className={styles.iconButton}>
+        {isOpen ? <span style={{ fontWeight: 600 }}>Save</span> : <Icon name="pencil" />}
```
ladriennel: I think labeling this button "Save" is slightly misleading since saving all the alumni data doesn't actually require you to save the name, email, or linkedin fields individually. I think relabeling it to "Done" or something else (if it sounds better) would be more clear?

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,242 @@
...
+  function openDelete(alum: Alumni): void {
+    setAlumniToDelete(alum);
+    setDeleteModalOpen(true);
+  }
+
+  async function deleteAlumni(alum: Alumni): Promise<void> {
+    await AlumniAPI.deleteAlumni(alum.uuid);
+    const location = alum.location?.trim();
+    if (location) {
+      await CityCoordinatesAPI.removeAlumniByLocationName(location, alum.uuid);
```
ladriennel: Instead of using `removeAlumniByLocationName`, we can use the same pattern used in the save handler in `AlumniAddEditModal.tsx` on line 754. I think this would be more helpful because we have preexisting logic and re-geocoding an existing location could potentially have float differences

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  function handleRowAction(
+    alum: Alumni,
+    _: React.SyntheticEvent<HTMLElement>,
+    data: DropdownProps
+  ): void {
+    const value = data.value as string | undefined;
+    if (value === 'edit') {
+      // TODO: open edit modal
```
ladriennel: nit: can remove completed TODOs

frontend/src/components/Admin/EditAlumni/EditAlumni.tsx
```diff
@@ -0,0 +1,245 @@
...
+      const allCoords = await CityCoordinatesAPI.getAllCityCoordinates();
+      await Promise.all(
+        allCoords
+          .filter((coord) => coord.alumniIds.includes(alum.uuid))
+          .map((coord) =>
+            CityCoordinatesAPI.removeAlumniFromLocation(coord.latitude, coord.longitude, alum.uuid)
+          )
+      );
+    }
+    await ImagesAPI.deleteImage(alum.imageUrl);
```
ladriennel: Alumni image deletion is already handled in the backend API through `deleteAlumni,` so this line isn't necessary

frontend/src/API/CityCoordinatesAPI.ts
```diff
@@ -10,4 +16,76 @@ export default class CityCoordinatesAPI {
...
+    const response = await APIWrapper.delete(
+      `${backendURL}/city-coordinates/${latitude}/${longitude}/alumni/${alumniId}`
+    );
+    return response.data.cityCoordinates;
+  }
+
+  /**
+   * Resolves a human-readable location name to lat/long, then removes the alumni from that city-coordinates document.
+   */
+  static async removeAlumniByLocationName(
```
ladriennel: Can remove this method now since it's not being used anymore.

ladriennel (approved):
Great work Max, and thanks for working on this huge PR!! I have 2 comments left that are minor, so once those are fixed, everything looks good.

And the build error is from a pre-existing issue, so if it keeps blocking CI, try adding `jest.setTimeout(15000)` to the top of `ShoutoutsDao.test.ts` to unblock for now (we can resolve the actual issue separately). lmk if you still get issues though!

PR #1149, opened 2026-04-12
Coffee Chat suggestions/category refactor

ladriennel:
### Summary 
Currently, coffee chat suggestions are stored quite inefficiently where every member has a reference to all other members satisfying each category. Every semester, you had to manually update a `COFFEE_CHAT_BINGO_BOARD `constant in code, run the script, and open a PR. With this refactor, the collection is now modeled after the categories: there will be 16 Firestore documents for each bingo board category with an index denoting its position on the board and storing a list of all associated members. 
- Admins can upload csv of suggestions through the UI in the admin review coffee chats page. Note: this csv shape is still very constrained to what the script expects.
- When members request their suggestions by going to the coffee chat bingo page, the API reads the 16 documents and filters them out on their own categories (meaning that the actual data doesn't exclude "self" from any categories as before).

[Notion](https://www.notion.so/Idol-SP26-2ea0ad723ce18080ae5ffbae682175ad?p=3110ad723ce180369970df350c4900e2&pm=s)

### Test Plan 
- Verify firebase (non-prod for now) has all 16 categories with right name and right corresponding members
- Load admin coffee chats page, make sure all categories render
- Try uploading csv with button, should completely refresh categories + members
- Load coffee chat form page for members, check categories render, members correct, self not included
- Ensure old coffee chat submission logic unaffected

### Notes 
- Need to populate prod database with new collection and remove old collection when pr ready


backend/src/API/coffeeChatAPI.ts
```diff
@@ -160,18 +161,78 @@ export const getCoffeeChatBingoBoard = (): Promise<string[][]> =>
...
-  const suggestions = await CoffeeChatDao.getCoffeeChatSuggestions(email);
-  if (!suggestions) {
-    throw new BadRequestError(
-      `Coffee chat suggestions does not exist for member with email ${email}`
+  const [member, suggestions] = await Promise.all([
+    getMember(email),
+    CoffeeChatDao.getCoffeeChatSuggestions()
+  ]);
+
+  if (!member) return suggestions;
```
jane-leon: The old version threw BadRequestError in this case. I was wondering why we removed that check as if getMember(email) returns undefined we skip the self-filter entirely and return suggestions that still contain the requester themselves which is what this function is supposed to prevent.

backend/scripts/populate-coffee-chat-categories.ts
```diff
@@ -80,38 +77,48 @@ const getMembersByCategory = async (members: IdolMember[]) => {
 
   suggestions['a newbie'] = filteredSuggestions(
     members,
-    (mem) => mem.semesterJoined === 'Fall 2025'
+    (mem) => mem.semesterJoined === 'Spring 2026'
+  );
+
+  const alumniSnapshot = await db.collection('alumni').where('gradYear', '==', 2025).get();
```
jane-leon, line 83: 'Spring 2026' and gradYear === 2025 will silently produce empty lists next semester. I was wondering since the whole point of this refactor is to avoid the "manually update a constant and open a PR" cycle, 
 we could may consider storing "current newbie semester" and "current alumni gradYear" in a settings doc that way we would only need to change via Firebase, might be too much though.

backend/scripts/populate-coffee-chat-categories.ts
```diff
@@ -80,38 +77,48 @@ const getMembersByCategory = async (members: IdolMember[]) => {
...
+    (name, index): CoffeeChatCategory => ({ name, members: suggestions[name], index })
+  );
 };
 
 const main = async () => {
   const members = await memberPromise;
-  const membersByCategory = await getMembersByCategory(members);
+  const categories = await getMembersByCategory(members);
 
   /* commented out for now - no need to filter self from categories
```
jane-leon: i think this block can go cuz dead code, like the commented out portion

backend/src/firebase.ts
```diff
@@ -85,6 +85,16 @@ export const coffeeChatSuggestionsCollection: admin.firestore.CollectionReferenc
     }
   });
 
+export const coffeeChatCategoriesCollection: admin.firestore.CollectionReference<CoffeeChatCategory> =
```
jane-leon, line 78: I don't think nothing references coffeeChatSuggestionsCollection anymore (I grepped). If not, then we can delete this export together with the prod cleanup you mentioned in with the "remove old collection when PR ready".

jane-leon (approved):
lgtm adrienne, just left some nits and some qs, great job!!

backend/src/API/coffeeChatAPI.ts
ladriennel: You're right it shouldn't return suggestions in this case, I'll fix that!

backend/scripts/populate-coffee-chat-categories.ts
ladriennel: i think this might be ok not to add right now because the categories change a lot, so this might not persist well through each semester. I also wonder if this logic is worth keeping at all since we can likely rely on member's self-submissions for the most part, and with this refactor, admins can directly fix the suggestions if need be. (but we can remove it later since it doesn't do anything harmful for now especially since alumni and newbies aren't a category this sem)

PR #1152, opened 2026-04-22
Updated tec period 

ladriennel:
### Summary 
Updated last tec period from 5/3 to 5/4

### DIFF
--- a/backend/src/consts.ts
+++ b/backend/src/consts.ts
@@ -14,7 +14,7 @@ export const REQUIRED_LEAD_TEC_CREDITS = 2;
 export const TEC_DEADLINES = [
   new Date('2026-02-22T23:59:59'),
   new Date('2026-03-19T23:59:59'),
-  new Date('2026-05-03T23:59:59')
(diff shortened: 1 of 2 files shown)


backend/src/consts.ts
```diff
@@ -14,7 +14,7 @@ export const REQUIRED_LEAD_TEC_CREDITS = 2;
 export const TEC_DEADLINES = [
   new Date('2026-02-22T23:59:59'),
   new Date('2026-03-19T23:59:59'),
-  new Date('2026-05-03T23:59:59')
+  new Date('2026-05-04T23:59:59')
```
zkdlinyne, line 17: looks good! I'll add these changes into my current branch as well for the default fall-back const I created

zkdlinyne (approved):
Looks good Adrienne!

backend/src/consts.ts
zkdlinyne: some checks failing, maybe yarn format ?

PR #1153, opened 2026-04-24
Add TEC Period Editing Capabilities to IDOL

zkdlinyne:
### Summary 

This PR moves TEC period deadlines and lead/member credit requirements out of hardcoded consts.ts files and into a Firestore document, so they can be edited in directly on IDOL without changing the code. On the frontend, a small panel (available only to admins) was added to edit, delete, and save (update) TEC Periods.

**Backend**
- Added a TECConfig type to common-types/index.d.ts to store array of strings periodEndDates, and integers requiredMemberTecCredits and requiredLeadTecCredits
- Added a tec-config Firestore collection (see Database Schema Change below for what that means for Firebase) to backend/src/firebase.ts. Config is a document with id "current"
- Created backend/src/dao/TecConfigDao.ts to return a clean TecConfig object (includes fallback DEFAULT_TEC_CONFIG if the doc is invalid)
- Added two endpoints in api.ts (backend): GET /tec-config and PUT /tec-config (which calls TecConfigDao.updateTecConfig
- API/mailAPI.ts now has the period reminder fetching the config from the firebase instead of just reading the constants
- Removed constants TEC_DEADLINES, REQUIRED_MEMBER_TEC_CREDITS, and REQUIRED_LEAD_TEC_CREDITS from backend/src/consts.ts, since they are no longer of use.

**Frontend**
- Added the file frontend/src/API/TecConfigAPI.ts with notable functions getTecConfig() and updateTecConfig(config)
- Frontend utils.ts functions getTECPeriod and getPeriod no longer take TEC_DEADLINES as a parameter (no longer a constant) 
- Current components now fetch the config on mount
- Added admin editing capabilities (styling is not great, let me know how to improve, the "Save" button is brutally blue)
- Removed TEC_DEADLINES, REQUIRED_MEMBER_TEC_CREDITS, and REQUIRED_LEAD_TEC_CREDITS from frontend/src/consts.ts, since they are no longer of use.

### Test Plan 

1. Navigate to Admin
2. Go to Edit Team Events
3. Fiddle around with the calendar & period end dates, change the required Lead & Member TEC Credits
4. Click Save -> Done -> the same editing component should still be visible
5. To check if this updated on Firestore, navigate to cornelldti-idol-dev in Firebase and check collection "tec-config" -> document "current" -> see whether the field values have changed
Screenshot 2026-04-23 at 11 07 59 PM
Screenshot 2026-04-23 at 11 10 15 PM

### Notes 

Have never worked with yaml files before, so those are AI generated. 
Used some AI for the styling of the TecConfigEditor component, didn't really know how to import the calendar either, I tried to stay moderately consistent with the styling, but let me know specifically what to change!
_Currently missing updated backend tests & updated documentation_

### Database Schema Change 

Currently, in the cornelldti-idol-dev database, I've added a new collection called "tec config" which stores the previously hard-coded array of period end dates. The collection stores one singular document, called "current," with 3 fields: an array of period end dates, and the numbers of required member tec credits per period and lead tec credits per period (1 and 2 respectively).

### DIFF
--- a/backend/src/dao/TecConfigDao.ts
+++ b/backend/src/dao/TecConfigDao.ts
@@ -1,0 +1,46 @@
+import { tecConfigCollection } from '../firebase';
+import { DEFAULT_TEC_CONFIG } from '../consts';
+import { BadRequestError } from '../utils/errors';
+
+
+const TEC_CONFIG_DOC_ID = 'current';
+
+export default class TecConfigDao {
+  /** Returns the saved TEC config from Firestore document, or DEFAULT_TEC_CONFIG if invalid for some reason. */
+  static async getTecConfig(): Promise<TECConfig> {
+    const snap = await tecConfigCollection.doc(TEC_CONFIG_DOC_ID).get();
+    if (!snap.exists) {
+      return DEFAULT_TEC_CONFIG;
+    }
+    const data = snap.data();
+    if (!isValidTecConfig(data)) {
+      return DEFAULT_TEC_CONFIG;
+    }
+    return data;
+  }
+
+  static async updateTecConfig(config: TECConfig): Promise<TECConfig> {
+    if (!isValidTecConfig(config)) {
+      throw new BadRequestError("Invalid TEC config");
+    }
+
+    const normalizedConfig: TECConfig = {
+      periodEndDates: [...config.periodEndDates].sort(), // assures chronological order
+      requiredMemberTecCredits: config.requiredMemberTecCredits,
+      requiredLeadTecCredits: config.requiredLeadTecCredits
+    };
+
+    await tecConfigCollection.doc(TEC_CONFIG_DOC_ID).set(normalizedConfig);
+    return normalizedConfig;
+  }
+}
+
+function isValidTecConfig(data: TECConfig | undefined): data is TECConfig {
+  if (!data) return false;
+  return (
+    Array.isArray(data.periodEndDates) &&
+    data.periodEndDates.every((d) => typeof d === 'string') &&
+    typeof data.requiredMemberTecCredits === 'number' &&
+    typeof data.requiredLeadTecCredits === 'number'
+  );
+}
--- a/backend/src/dao/TecConfigDao.ts
+++ b/backend/src/dao/TecConfigDao.ts
@@ -2,7 +2,6 @@ import { tecConfigCollection } from '../firebase';
 import { DEFAULT_TEC_CONFIG } from '../consts';
 import { BadRequestError } from '../utils/errors';
 
-
 const TEC_CONFIG_DOC_ID = 'current';
 
 export default class TecConfigDao {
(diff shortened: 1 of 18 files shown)


backend/src/dao/TecConfigDao.ts
```diff
@@ -0,0 +1,45 @@
...
+    await tecConfigCollection.doc(TEC_CONFIG_DOC_ID).set(normalizedConfig);
+    return normalizedConfig;
+  }
+}
+
+function isValidTecConfig(data: TECConfig | undefined): data is TECConfig {
+  if (!data) return false;
+  return (
+    Array.isArray(data.periodEndDates) &&
+    data.periodEndDates.every((d) => typeof d === 'string') &&
```
ladriennel: As per your question from earlier, I think it would be a good idea to validate if it's a parseable date to be defensive as well as maybe checking if `data.periodEndDates` isn't empty

frontend/src/components/Admin/TeamEvent/TeamEvents.tsx
```diff
@@ -93,14 +101,17 @@ const TeamEvents: React.FC = () => {
...
+  }, [isLoading, tecConfig]);
 
   return (
     <div>
       <div className={[styles.formWrapper, styles.wrapper].join(' ')}>
         <h1>Create a Team Event</h1>
         <TeamEventForm formType={'create'}></TeamEventForm>
       </div>
+      <div className={[styles.formWrapper, styles.wrapper].join(' ')}>
+        <TecConfigEditor />
```
ladriennel: nit: the `TecConfigEditor` and its parent component/`TeamEvents.tsx` page both fetch and have their own copies of `tecConfig`, so when you update the periods and save, you have to refresh for changes to show on the period groups.

frontend/src/utils.ts
```diff
@@ -244,10 +242,10 @@ export const getTimeString = (unixTime: number): string => {
  * @returns The index of the current TEC period in the `TEC_DEADLINES` array.
```
ladriennel: I think you said you would update most documentation later, but I just wanted to note to update it here as well (and if you're changing old functionality it might be easier to change the doc when u do, so updating the change later doesn't get lost).

backend/src/dao/TecConfigDao.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export default class TecConfigDao {
+  /** Returns the saved TEC config from Firestore document, or DEFAULT_TEC_CONFIG if invalid for some reason. */
+  static async getTecConfig(): Promise<TECConfig> {
+    const snap = await tecConfigCollection.doc(TEC_CONFIG_DOC_ID).get();
+    if (!snap.exists) {
+      return DEFAULT_TEC_CONFIG;
+    }
+    const data = snap.data();
+    if (!isValidTecConfig(data)) {
```
ladriennel, line 27: I think that for when the data isn't valid, returning a default constant as a fallback slightly defeats the purpose of migrating the data to the database. If there is some issue with the actual data, there is no error logging or error surfaced to user. It might be better to throw here instead.

ladriennel (changes requested):
Hi Eva! Great work with the pr and new `tecConfig` setup! I think that the ui looks fine for now, and personally I think blue save button looks fine lol. yaml also looks good, and it's really only for documentation.

ladriennel (approved):
Hi Eva! Everything looks good, thanks for all the great work this sem!!

PR #1165, opened 2026-08-12
Updated recruitment timeline, fixed link bug

ladriennel:
### Summary 
- Updated Club Fest date and added new event
- Added loop to productLinks, so card for Loop members would render


new-dti-website-redesign/src/utils/memberUtils.ts
```diff
@@ -103,5 +103,8 @@ export const productLinks: { [key: string]: { name: string; link?: string } } =
   },
   redi: {
     name: 'redi'
+  },
+  loop: {
+    name: 'Loop'
```
zkdlinyne, line 108: Looks all good on the loop side!

zkdlinyne (approved):
Hey Adrienne, looks all good! I tested clicking on Annie's profile, and I don't get an error message anymore.

PR #1167, opened 2026-08-18
More Updates Recruitment Timeline

ladriennel:
### Summary 

Made more updates to recruitment events. Fixed minor bug to make sure upperclassmen timeline populated as default.


mshi-27 (approved):
Hi Adrienne, everything looks great! Ran the website locally and double checked that all dates and locations are correct, and the default cycle is set to Fall upperclassmen.

PR #1168, opened 2026-08-24
More Timeline Fixes

ladriennel:
### Summary 
- Added new Cursor event to recruitment timeline
- Updated location for DCC x DTI event

### Notes 
- In future to avoid making multiple PRs to update this timeline, we can add a feature to IDOL that allows you to directly update the events. This will require some system to create a pull request upon changes (will need to decide if we want a bot that runs on a schedule to check for changes or we manually hit some confirmation that triggers a pr?)


jane-leon (approved):
Lgtm! Also good note on adding that feature to IDOL, i think a manual button that need to be clicked would be better than on a bot on schedule just since the bot might run on friday, but lets say we need this event to show up on the site asap and its monday.

PR #1172, opened 2026-08-27
Updated products + fixed bug on website

ladriennel:
### Summary 

- Added CU Raise and Loop to product page on team website
- Updated Product Strategy course details 
- Fixed bug causing course instructor card image to not update correctly when switching between members


new-dti-website-redesign/src/components/TeamCard.tsx
zkdlinyne, line 100: I think the reason why there's a check warning coming up is because aria-label doesn't count as satisfying one of the lint rules "must have content" even though I see what the intent was here. Maybe adding an empty child will help?

zkdlinyne (changes requested):
Hey Adrienne, I tested locally and everything looks good. There's a small check warning in TeamCard.tsx under the components folder, I added a comment. Also something small, the Intro the Product Strategy notion isn't updated (button on website links to the correct notion page but information isn't correct) which isn't an issue with the PR but just wanted to flag it. 
Other than that looks great!

new-dti-website-redesign/src/components/TeamCard.tsx
ladriennel: Hi! I think it's a preexisting lint warning, but I'll fix it, thanks for catching!

zkdlinyne: Hi!! I ran locally, should be all good to go!

zkdlinyne (approved):
Hey Adrienne, looks all good!

PR #1174, opened 2026-09-05
Internal/External Toggle available on Admin + Member UI Changes

zkdlinyne:
### Summary 

Adds an optional Internal/External classification for team events, gated by a semester setting that leads/admins can turn on or off. Member changes include that they do not get a toggle, they only see the whether an event is Internal/External in the dropdown menu during submission when that setting is on.

**Config**

- Added considerEventKind to TECConfig (saved on the existing tec-config/current Firestore document).
- Admin Edit Team Events page has a “Consider Internal/External” toggle button next to Create a Team Event. 
- Default is false, when off, the create form looks like it does today.

**Creating events**

- When the Internal/External setting is on, creating an event requires an Internal or External entry.
- That value is stored as optional (kind ? : 'internal' | 'external' on TeamEventInfo.)

**Members**

- On the Submit Team Event Credits section of Team Event Credits, if the Internal/External setting is on and the event has a kind, the dropdown shows a color-coded bubble to the right of the name: light blue Internal, light green External, just like on the Fall 2 TECs google sheet.

- When considerEventKind is ON:

Remaining copy is split into internal/external buckets: “You need X more internal and Y more external this period,” using the saved kind minimum requirements (members 1 internal / 2 external; leads 1 internal / 4 external). Toggle OFF still uses the old single remaining line.

Check Team Event Credits uses those same two buckets: header, approved counts per period (N internal, M external), remaining, and Internal/External on event cards.

Approved credits only count toward a bucket if the event has a kind. Unlabeled events don’t fill either minimum.

Kind requirements are filled in if they’re missing on the config document, so remaining doesn’t fall back to calculateCredits(..., requiredCredits = 1) and show 1 internal / 1 external when nothing is submitted.

Admin dashboard / TEC period editor also count and edit by kind when the toggle is on; reminder emails are unchanged.

**Tests**

- TecConfigDao tests cover the new boolean.

### Test Plan 

Admin: (members can't test yet because admin/leads must create events first)
1. Navigate to Edit Team Events
2. Create a Team Event with the Internal/External setting ON
3. Check to see if you're allowed to create an event without selected Internal or External with the setting ON.
4. After testing step 3, create the event, then navigate to Member side (Team Event Credits -> Submit a Team Event Credit) 
5. Open the drop down to select an event to submit a photo for, and see your newly-created event with an Internal or External bubble label in its row.

### Notes 

Internal/external is not shown yet on admin event cards, event details, the TEC dashboard, the Edit Event modal, member approved/pending cards, or emails or anything. 

This PR ended up being a little over 600 lines. I kept it as one because the second half can't really ship without the first.

The first portion of the PR is the semester toggle plus storing the "kind" type on events (create/edit, member dropdown bubbles). The second portion is what happens once that toggle is on: Internal and External are two separate requirements, not one total (i.e. you must complete X internal and Y internal vs. you must complete X total TECs).

If those ended up in a separate PR, members would see Internal/External labels while “you need X more” and Check Team Event Credits still used the old single total number. The admin dashboard would also still treat people as on-track from a combined total, which is completely incorrect. That’s why I didn’t split the PR.

### Breaking Changes 

- Database schema change: TECConfig now requires considerEventKind: boolean.

Still left to do in separate PR: UI changes for members (internal/external labels on event cards), Emails (add internal/external to text in reminder emails)


frontend/src/components/Admin/TeamEvent/TeamEvents.tsx
ladriennel, line 142: One thing I noticed was when toggling "Consider Internal/External" and then making changes and saving in the TEC editor below without refreshing the page (like changing the number of credits), undos the toggle change. The tec-config collection also reflects the stale data. This is probably a state sync issue between the toggle and the other TEC edit section modifying the same document but in separate forms.

ladriennel (commented):
Hi Eva, thanks for putting this out so quick! I left 1 comment for now and the rest of the current functionality looks pretty good to me. I know you're working on additional items, so I'll review more after!

One thing to note though, we should make the `kind` field on each event editable (in same place we edit all the other fields), and this feature can be available when the toggle is on. If this pr isn't too heavy with the rest of the logic, it could be added with the changes here, otherwise it's not the most urgent.

zkdlinyne:
This PR ended up being a little over 600 lines. I kept it as one because the second half can't really ship without the first.

The first portion of the PR is the semester toggle plus storing the "kind" type on events (create/edit, member dropdown bubbles). The second portion is what happens once that toggle is on: Internal and External are two separate requirements, not one total (i.e. you must complete X internal and Y internal vs. you must complete X total TECs).

If those ended up in a separate PR, members would see Internal/External labels while “you need X more” and Check Team Event Credits still used the old single total number. The admin dashboard would also still treat people as on-track from a combined total, which is completely incorrect. That’s why I didn’t split the PR.

**Member-facing additions:** 

When considerEventKind is ON:

- Remaining copy is split into internal/external buckets: “You need X more internal and Y more external this period,” using the saved kind minimum requirements (members 1 internal / 2 external; leads 1 internal / 4 external). Toggle OFF still uses the old single remaining line.

- Check Team Event Credits uses those same two buckets: header, approved counts per period (N internal, M external), remaining, and Internal/External on event cards.

- Approved credits only count toward a bucket if the event has a kind. Unlabeled events don’t fill either minimum.

- Kind requirements are filled in if they’re missing on the config document, so remaining doesn’t fall back to calculateCredits(..., requiredCredits = 1) and show 1 internal / 1 external when nothing is submitted.

- Admin dashboard / TEC period editor also count and edit by kind when the toggle is on; reminder emails are unchanged.

_Still left to do in separate PR_: UI changes for members (internal/external labels on event cards), Emails (add internal/external to text in reminder emails)

frontend/src/components/Forms/TeamEventCreditsForm/TeamEventCreditsForm.tsx
```diff
@@ -196,15 +229,24 @@ const TeamEventCreditForm: React.FC = () => {
             Select a Team Event: <span className={styles.red_color}>*</span>
           </label>
           <div className={styles.bold}>
-            {remainingCredits > 0 && (
+            {tecConfig.considerEventKind && hasRemainingKindCredits && (
+              <span className={styles.red_color}>
+                You need {remainingInternalCredits} more internal and {remainingExternalCredits}{' '}
```
ladriennel: Nit: Could you change the text for the TEC reminder on the internal/external toggle on to match the original a little more? Something like `You must submit at least {remainingInternalCredits} more internal and {remainingExternalCredits} more external TEC in the current period.`

ladriennel (approved):
Hi Eva! Thanks for making these changes so efficiently! Everything looks good to me (just left 1 nit). The main thing I would suggest is to update the PR summary with your most recent implementation (you can probably merge some of the content in your comment into this). 

Once you merge, we should ensure that the toggle, credit requirements, and event kind tag are configured properly in production. I can also check in with leads to let them know we added this.

PR #1177, opened 2026-09-14
Add Reimbursement Frontend API + Connect

jane-leon:
### Summary <!-- Required -->
This PR connects the Requestor Dashboard to the real reimbursement backend and replaces all mock data with live API calls.                                                      
  - Added a new ReimbursementAPI module to talk to the backend                                                        
  - Replaced mock data on the dashboard with real API calls                                                         
  - Added loading spinner and auto-refresh after submitting                                                           
  - Dashboard shows a Budget Overview per team the user belongs to, labeled with the team name                        
  - Case-insensitive team ↔ subteam matching (e.g. subteam "IDOL" now matches team "idol")                            
  - Leads support: if a user has "leads" in their subteams, the dashboard/submit modal surfaces the 5 lead-group teams (biz-leads, pm-leads, ops-leads, design-leads, dev-leads) so leads can pick which group the reimbursement is for. Each has its own budget.                                                                                            
  - Added a seed script yarn workspace backend seed-reimbursement-teams that creates the 5 lead-group teams                                                                                  
  - Rebuilt the submit form: prefills name/NetID, uses a URL input for the receipt link, and validates required fields before letting you submit                                                                                          
  - Submit button now actually creates the request, shows success/error toasts, and refreshes the dashboard         
  - Fixed a date-of-purchase off-by-one bug (YYYY-MM-DD was parsed as UTC midnight and rendered a day earlier in Eastern time; now parsed as local midnight)                                                                         
  - Removed the old mock data file                                                                                    
  - Hid the up/down arrow spinners on the Amount input for a cleaner look                                             
  - Added spacing above the Reason and Receipt link fields so they aren't cramped              

### Notion/Figma Link <!-- Optional -->
(https://app.notion.com/p/Reimbursement-Frontend-API-and-wire-Requestor-dashboard-3c80ad723ce180738bb8c20322d2bd3d?source=copy_link)

### Test Plan <!-- Required -->

Tested locally :                                  
  - Submitted a request with valid inputs → appears in Firestore `reimbursement-requests` and refreshes into the   
  "My Requests" table without manual reload
  - Validation gates: Next stays disabled for blank required fields, `amount <= 0`, and invalid receipt URLs (missing    
  protocol / not a URL)                                     
  - Back button preserves form state where closing and reopening the modal resets it                                         
  - Attendees comma-separated string is parsed into `string[]` in Firestore                                                                            
  - Dashboard shows "not assigned to any reimbursement team" and disables the submit button when no `reimbursement-teams` doc's `teamName` matches the user's `subteams`

<img width="1274" height="630" alt="image" src="https://github.com/user-attachments/assets/c284d136-a040-4406-817c-254d3f294525" />
<img width="1181" height="560" alt="image" src="https://github.com/user-attachments/assets/fa68d45c-dd7c-43b3-9fa9-b99a50b447ca" />
<img width="1148" height="552" alt="image" src="https://github.com/user-attachments/assets/f63277ff-9497-472d-8170-2a513b995b74" />
<img width="1348" height="528" alt="image" src="https://github.com/user-attachments/assets/2cddba39-8f6e-424c-afcb-a2aa024abe78" />
<img width="1420" height="503" alt="image" src="https://github.com/user-attachments/assets/c87c1174-4057-42ed-86c5-5cf4adf202fe" />


### Notes <!-- Optional -->

  - **Team = user mapping**: ReimbursementTeam doesn't currently track requestor membership directly. The frontend derives it by intersecting user.subteams (lowercased) with team.teamName (lowercased). For "leads" subteam members, it also surfaces any team whose name ends with -leads. This assumes reimbursement-team names line up with subteam strings.
  - **Receipt storage**: we're storing a URL string (Drive link) rather than uploading files to Firebase Storage.               
  - Only users with lead roles (`ops-lead`, `dev-lead`, `design-lead`, `business-lead`, `product-lead`, `tpm`, `pm`) can submit, this is enforced by the existing backend `hasReimbursementRequesterPermissions` check.      

  Out of scope for this PR:                                                                                                                                                                                                         
  - Request detail view, edit flow, "View all" page                                                                   
  - Admin approval dashboard (which is what actually updates totalSpent, pending requests don't decrement the budget yet)                                                                                                                
  - Deciding who gets admin view for each lead-group budget (e.g. does only biz-lead see biz-leads budget, or all leads?)                                                                                                             
  - Non-lead teams (IDOL, Redi, Loop, Carriage, CUReviews, CoursePlan, CU Apts) still need to be seeded in Firestore before members of those teams can submit


frontend/src/components/Forms/ReimbursementDashboard/SubmitRequestModal.tsx
```diff
@@ -1,105 +1,168 @@
...
+    try {
+      await ReimbursementAPI.createRequest({
+        teamId,
+        amount: Number(amount),
+        reason: reason.trim(),
+        attendees: attendees
+          .split(/[,\n]/)
+          .map((a) => a.trim())
+          .filter(Boolean),
+        dateOfPurchase: new Date(dateOfPurchase).getTime(),
```
zkdlinyne: While testing, I noticed that even though while filling out the form the Date of Purchase is listed as the current day (i.e. 9/18/2026), the Date of Purchase value on the Reimbursement Dashboard after submitting a request is behind by a day (i.e. 9/17/2026). I think this is because dateOfPurchase from input type=”date” via new Date(“YYYY-MM-DD”) is UTC midnight, so some evening reimbursement request submissions may end up being off by a day. 
I think constructing a local date may fix this

frontend/src/components/Forms/ReimbursementDashboard/ReimbursementDashboard.tsx
```diff
@@ -1,16 +1,39 @@
...
+    setTeams(allTeams);
+    setLoading(false);
+  }, []);
+
+  useEffect(() => {
+    loadData();
+  }, [loadData]);
+
+  const myTeams = useMemo(
+    () => teams.filter((t) => user.subteams.includes(t.teamName)),
```
zkdlinyne: nit but .includes is case sensitive, so "idol" won't be equal to "IDOL."

zkdlinyne (changes requested):
Hi Jane!! Thanks so much for implementing this PR, just left a couple comments :)

Structurally, the frontend is all good!!! and I tested and everything shows up in firestore as well. I did come across something interesting while testing and trying to assign myself the role of a lead:

A developer on IDOL (with subteams: ["idol"]) matches the team and can open the modal, then gets a permission error from hasReimbursementRequesterPermissions.

A lead on subteams: ["leads"] passes the backend role check, then sees “not assigned to any reimbursement team” and a disabled Submit button. I know you talked about refactoring in your pr summary but just wasn't sure if this is what you meant :)

I think this is due to the dashboard only enabling Submit if user.subteams matches team.teamName, so not really an issue with your frontend changes in this pr.

Also nit but /forms/reimbursement (in frontend/src/pages/forms/index.tsx) has no adminOnly flag, it will probably show up for all members

frontend/src/API/ReimbursementAPI.ts
```diff
@@ -0,0 +1,65 @@
...
+            headerMsg: "Couldn't get your reimbursement requests",
+            contentMsg: `Error was: ${val.error}`
+          });
+          return [];
+        }
+        return val.requests as ReimbursementRequest[];
+      });
+  }
+
+  public static getAllTeams(): Promise<ReimbursementTeam[]> {
```
ladriennel: Just noticed that right now `getAllTeams` is hitting the backend endpoint for `getAllReimbursementTeams` which is gated by the `isLeadOrAdmin` check which wouldn't permit access by a normal pm/tpm role. It's been working for us regardless of what role we set ourselves to because we have admin access likely.

ladriennel: I also noticed that if a member doesn't have the correct perms, it only blocks them at request submission not page access. This is not something we necessarily have to change in this pr, but I wanted to ask if you agree that it might be better to hide the page itself from those members.

ladriennel (commented):
I just left one comment (kinda), but overall everything looks really good. Thanks Jane!

zkdlinyne (approved):
Looks all good Jane!

